PR #12, opened 2022-01-11
[client/rest] task: convert to encoded address

AnthonyLaw:
**Note: Do not merge until CI is ready.**

### Current behavior 
- All the addresses in the REST response are in decoded format. 
- Example `98E0D138EAF2AC342C015FF0B631EC3622E8AFFA04BFCC56`

### What was the issue?
- `Decoded` address format is not user-friendly. We have feedback from the community and discussion, change the address to encoded format.
- Example `TDQNCOHK6KWDILABL7YLMMPMGYRORL72AS74YVQ`

### What's the fix?
- Convert addresses from decoded to encoded (base32) format.
- updated unit test

### DIFF
--- a/client/rest/src/db/dbUtils.js
+++ b/client/rest/src/db/dbUtils.js
@@ -82,12 +82,11 @@ const dbUtils = {
 	},
 
 	/**
-	 * Formats binary to a base32 address or hex address
-	 * @param {MongoDb.Binary} binary Address|NamespaceId from MongoDb.
-	 * @param {boolean} formatAddressUsingBase32 if base32 format should be used when formatting an address. Hex otherwise.
-	 * @returns {string} the address in base32 format or hex format depending on formatAddressUsingBase32
-	 */
-	bufferToUnresolvedAddress: (binary, formatAddressUsingBase32) => {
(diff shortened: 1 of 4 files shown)


Jaguar0625:
We already know REST needs to be refactored (along with the explorer and wallets). It’s a bit of a mess right now. We also know we need to sit down and carefully plan out what features and queries we want from REST. 

Let’s leave this PR until after the SN program, and then we can tackle REST after the node configuration tool. It doesn’t make sense to push out an update to REST right now and have everyone update. We already know it will need to be overhauled.

It doesn’t make sense to push out an update to REST right now and have everyone update. We already know it will need to be overhauled.

Jaguar0625:
We already know REST needs to be refactored (along with the explorer and wallets). It’s a bit of a mess right now. We also know we need to sit down and carefully plan out what features and queries we want from REST.

Let’s leave this PR until after the SN program, and then we can tackle REST after the node configuration tool. It doesn’t make sense to push out an update to REST right now and have everyone update. We already know it will need to be overhauled.

It doesn’t make sense to push out an update to REST right now and have everyone update. We already know it will need to be overhauled.

PR #13, opened 2022-01-12
[monorepo] task: add catapult bug report template

0x6861746366574:
Standardization ahoy.


0x6861746366574:
@gimre-xymcity We should also add instructions for how to pull logs from "symbol-node-configurator" for users - to confirm, are all the log files located in `/catapult-client/_build`?

gimre-xymcity:
@0x6861746366574 are you referring to _node_ logs or BUILD logs?

.github/ISSUE_TEMPLATE/client/catapult/bug.md
```diff
@@ -0,0 +1,41 @@
...
+
+> Please provide detailed steps to reproduce this bug. 
+> [1.] Go to '...'
+> [2.] Click on '...'
+> [3.] Scroll down to '....'
+> [4.] See error. 
+
+#### Logs
+
+> Please attach your logs as text - not as screenshots. You can upload them to [Pastebin](https://pastebin.com/) or as a [Gist](https://gist.github.com/).
```
gimre-xymcity, line 34: if we want server logs we need to consider few things:
 1. some kind of (optional) log anonymization - I'm not sure if we want that, as things like (node) pubkeys that are present in logs might be valuable for investigation.
  2. there probably will be _multiple_ log files so some might require zip with whole logs dir (and this ofc won't go to pastebin or gist, I guess google drive or anything like that would be most covenient - as long as it's not some shady 3rd party site, think megaupload etc.)
  3. re 2. we'd probably have to archive those _somewhere_ - just for future reference.

PR #17, opened 2022-01-12
lint(python): convert indentation from spaces to tabs

Jaguar0625:


sdk/python/symbolchain/core/nem/external/ed25519.py
```diff
@@ -180,9 +180,9 @@ def scalarmult_B(e):
 def encodeint(y):
```
gimre-xymcity, line 180: ugh, should we exlude external from lints? 😬

Jaguar0625: there weren't too many changes. i exclude js one because that would require quite a bit of changes.

gimre-xymcity: y, but it did reformat of conditions

gimre-xymcity (approved):

PR #18, opened 2022-01-12
task(rest): drop tweetnacl dependency and unused crypto wrappers

Jaguar0625:


client/rest/src/catapult-sdk/crypto/sha3Hasher.js
```diff
@@ -1,78 +0,0 @@
-/*
```
gimre-xymcity, line 1: just to make sure "createHasher part is actually keccak not sha3, right?
(doesn't matter for review, just asking to make sure)

Jaguar0625: no, it was sha3. initially, i think, it was keccak or sha3 based on "configuration".

gimre-xymcity: it _looks_ like keccak (see custom finalize)

but tbh, more worrying part is something similar in code that does exactly that via sha3().update()...
`sha3_256.update(catapult.utils.convert.hexToUint8(encodedPath + branchLinks.join('')))`

this is _probably_ wrong, isn't it?

Jaguar0625: no call to output function { arrayBuffer, hex } ... is finalizing hash (i.e. adding sha3 padding)

gimre-xymcity (approved):

PR #23, opened 2022-01-13
Includes reorder

gimre-xymcity:
Reorder after discussion in #infra-github with @Jaguar0625

### DIFF
--- a/client/catapult/extensions/mongo/src/MongoBulkWriter.h
+++ b/client/catapult/extensions/mongo/src/MongoBulkWriter.h
@@ -31,10 +31,10 @@
 #include "catapult/exceptions.h"
 #include "catapult/types.h"
 #include <boost/asio/io_context.hpp>
-#include <bsoncxx/json.hpp>
-#include <mongocxx/client.hpp>
 #include <mongocxx/config/version.hpp>
 #include <mongocxx/exception/bulk_write_exception.hpp>
+#include <bsoncxx/json.hpp>
+#include <mongocxx/client.hpp>
 #include <mongocxx/pool.hpp>
 #include <unordered_set>
 
--- a/client/catapult/extensions/finalization/FinalizationExtension.cpp
+++ b/client/catapult/extensions/finalization/FinalizationExtension.cpp
@@ -19,13 +19,13 @@
 *** along with Catapult. If not, see <http://www.gnu.org/licenses/>.
 **/
 
-#include "finalization/src/io/FileProofStorage.h"
 #include "src/FinalizationBootstrapperService.h"
 #include "src/FinalizationConfiguration.h"
 #include "src/FinalizationMessageProcessingService.h"
 #include "src/FinalizationOrchestratorService.h"
 #include "src/FinalizationService.h"
 #include "src/FinalizationSyncSourceService.h"
+#include "finalization/src/io/FileProofStorage.h"
 #include "catapult/config/ConfigurationFileLoader.h"
 #include "catapult/extensions/ProcessBootstrapper.h"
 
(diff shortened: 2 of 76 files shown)


client/catapult/extensions/mongo/src/MongoBulkWriter.h
```diff
@@ -31,10 +31,10 @@
 #include "catapult/exceptions.h"
 #include "catapult/types.h"
 #include <boost/asio/io_context.hpp>
-#include <bsoncxx/json.hpp>
-#include <mongocxx/client.hpp>
 #include <mongocxx/config/version.hpp>
 #include <mongocxx/exception/bulk_write_exception.hpp>
+#include <bsoncxx/json.hpp>
+#include <mongocxx/client.hpp>
```
Jaguar0625, line 37: mongo, bson, mongo looks odd

gimre-xymcity: that is because it prioritiezes:
(mongo|bson)/subdir/

and then
(mongo|bson)/

and within them sorts alphabetically...
could split them into separate groups

Jaguar0625 (approved):

client/catapult/extensions/mongo/src/MongoBulkWriter.h
Jaguar0625: probably would split, but your choice

linters/cpp/Rules.py
```diff
@@ -147,18 +148,45 @@ def namespace_check(ns_unified, full_path):
...
+
 		if 'validators' in path_elements:
 			return '"Validators.h"'
 
 		if 'observers' in path_elements:
 			return '"Observers.h"'
 
-		return '"{}.h"'.format(path_elements[-1][:-4])
+		return sorted_includes[0].include
+		# return '"{}.h"'.format(path_elements[-1][:-4])
```
Jaguar0625, line 168: delete ?

linters/cpp/Rules.py
```diff
@@ -147,18 +148,45 @@ def namespace_check(ns_unified, full_path):
...
+
+	@staticmethod
+	def _checked_o_or_v(path_elements):
```
Jaguar0625, line 171: o or v means ?

Jaguar0625 (commented):
should we run clang_format as part of lint?

linters/cpp/exclusions.py
```diff
@@ -129,25 +129,65 @@
...
 	'tests/catapult/utils/TraitsTests.cpp': 'catapult/utils/traits/Traits.h',
-	'tests/catapult/utils/StlTraitsTests.cpp': 'catapult/utils/traits/StlTraits.h'
+	'tests/catapult/utils/StlTraitsTests.cpp': 'catapult/utils/traits/StlTraits.h',
+
+	'tests/int/node/basic/LocalNodeBasicTests.cpp': 'tests/int/node/test/LocalNodeBasicTests.h',
+	'tests/int/stress/NemesisBlockLoaderIntegrityTests.cpp': 'catapult/extensions/NemesisBlockLoader.h',
+	'tests/int/stress/UtUpdaterIntegrityTests.cpp': 'catapult/chain/UtUpdater.h'
 }
 
 PLUGINS_FIRSTINCLUDES = {
```
Jaguar0625, line 139: if we are using clang_format, should we mostly remove these and let it order?

gimre-xymcity: ofc, would need to rip out all headers ordering/checking, I updated the rules, just to make sure what clang did with _our rules_ makes some sense

linters/cpp/Rules.py
gimre-xymcity: was for observers and validators, as I was adding this one by one, but it's perfectly ok to remove it now 🙃

Jaguar0625:
[2022-11-16T17:06:03.364Z] ERROR: /home/ubuntu/jenkins/workspace/l_generated_symbol_linters_PR-23/linters/cpp/Rules.py Imports are incorrectly sorted and/or formatted.

PR #24, opened 2022-01-14
fix(sdk/python): propagate running transaction vectors from travis.yml

gimre-xymcity:
(I'm not sure why we had that script in first place ¯\_(ツ)_/¯)


sdk/python/tests.sh
```diff
@@ -1,3 +1,6 @@
 #!/bin/bash
 
+PYTHONPATH=. SCHEMAS_PATH="../../catbuffer/schemas/test-vectors" pytest tests/generator
```
Jaguar0625, line 3: shouldn't this go in vectors.sh, not tests.sh ?

gimre-xymcity: this is _much_ faster than running test-vectors... so imo can go here
earlier both catbuffer test vectors and (big) test vectors were part of a common stage `test`, but catbuffer tests did not have branch limitation:

```yaml
    - stage: test
      name: unit tests
# ...
    - name: catbuffer test vectors
# ...
    - name: test vectors
      if: branch = env(DEV_BRANCH) OR branch = env(RELEASE_BRANCH)
```

Jaguar0625: these are vectors though, don't really want to run through them when running unit tests 😅

gimre-xymcity: tomato, tomato ^^
`unittest discover` - `384 tests in 1.943s`
catbuffer "vectors" - `188 passed in 0.89s`

gimre-xymcity:
merged into dev in 4b24ef3194859655d6b51d44dcc7152424aee2bb

PR #25, opened 2022-01-14
task(sdk/python): flatten by removing 'core' subdirectory

Jaguar0625:


gimre-xymcity:
(rebased on top of dev)

gimre-xymcity (approved):

PR #27, opened 2022-01-16
[tests/vectors] test: add test vectors and update vectors.sh scripts

Jaguar0625:


sdk/python/vectors.sh
```diff
@@ -1,4 +1,6 @@
 #!/bin/bash
 
-python3 -m tests.vectors.all --blockchain nem --vectors ../test-vectors/nem
-python3 -m tests.vectors.all --blockchain symbol --vectors ../test-vectors/symbol
+set -ex
+
+python3 -m tests.vectors.all --blockchain nem --vectors "$(git rev-parse --show-toplevel)/tests/vectors/nem"
+python3 -m tests.vectors.all --blockchain symbol --vectors "$(git rev-parse --show-toplevel)/tests/vectors/symbol"
```
gimre-xymcity, line 6: why/how was it working  earlier? was it downloading vectors anywhere?

gimre-xymcity (approved):

sdk/python/vectors.sh
Jaguar0625: it was not working in monorepo. previously, travis was downloading vectors.

PR #28, opened 2022-01-16
[catbuffer/parser] feature: restore catparser single pass mode

Jaguar0625:
- remove intermediate yaml output from source control
- update python generator to use ast models instead of yaml input


gimre-xymcity (changes requested):
why removing yml schemas though?

sdk/python/generator/Generator.py
```diff
@@ -113,73 +94,59 @@ def bind_size_fields(struct_type):
...
-			# (currently this only happens for metadata value)
-			if 'size_' == size_field_name:
-				size_field_name = 'size'
-
 			size_field = struct_type.get_field_by_name(size_field_name)
 			size_field.set_bound_field(field)
 
 		if field.type_instance.sizeof_value:
-			struct_field = struct_type.get_field_by_name(field.yaml_descriptor['value'])
+			struct_field = struct_type.get_field_by_name(field.ast_model.name[:-5])
```
Jaguar0625, line 101: @gimre-xymcity - maybe you have some suggestion how to get around this hack? 😅

sdk/python/generator/Generator.py
```diff
@@ -71,39 +58,33 @@ def to_type_formatter_instance(type_instance):
...
+	elif isinstance_builtin(ast_model, FixedSizeInteger):
+		object_type = IntObject
 
-	return object_type(yaml_descriptor)
+	return object_type(ast_model)
 
 
 def create_printer(type_instance, name):
-	if type_instance.is_builtin:
+	if type_instance.is_builtin or isinstance(type_instance.ast_model, Alias):
```
gimre-xymcity, line 76: why is Alias needed here (`is_builtin` should already cover it 🤔)

sdk/python/generator/Generator.py
```diff
@@ -71,39 +58,33 @@ def to_type_formatter_instance(type_instance):
...
-	factory_type = type_map[struct_type.yaml_descriptor['factory_type']]
+	factory_type = type_map[struct_type.ast_model.factory_type]
 
-	discriminators = factory_type.yaml_descriptor.get('discriminator', [])
+	discriminators = [] if not hasattr(factory_type, 'discriminator') else factory_type.discriminator
 	for i, discriminator in enumerate(discriminators):
-		discriminators[i] = fix_name(discriminator)
-
-	for init in factory_type.yaml_descriptor['initializers']:
-		fix_item(init, 'target_property_name')
```
gimre-xymcity, line 106: hmm, so where/how are you doing this? (I mean this is needed at a later stage when comparing against "already fixed" field name iirc)

sdk/python/generator/Generator.py
```diff
@@ -113,73 +94,59 @@ def bind_size_fields(struct_type):
...
 		if field.type_instance.is_array and isinstance(field.get_type().size, str):
 			size_field_name = field.get_type().size
-			# 'unfix' name when searching for associated 'size' field:
-			#  * fields are added using 'original' name,
-			#  * but 'size' property of given array type is already fixed
-			#
-			# (currently this only happens for metadata value)
-			if 'size_' == size_field_name:
-				size_field_name = 'size'
-
```
gimre-xymcity, line 123: where did this go? :D

sdk/python/generator/Generator.py
gimre-xymcity: I assume ast_model must have ".value" field (?), as it is present in yaml, e.g. (nem):
(although I'm somewhat wondering if "value" was intended name for this...)

```
 - comments: mosaic id size
    disposition: sizeof
    name: mosaic_id_size
    signedness: unsigned
    size: 4
    type: byte
    value: mosaic_id
```

sdk/python/generator/Generator.py
gimre-xymcity: ok, I assume everything is moved to printers stage, which seems like a sane thing to do

sdk/python/generator/AbstractImplMap.py
```diff
@@ -30,4 +30,4 @@ def add(self, parent_type: StructObject, child_type: StructObject):
 		self.mapping[parent_type.typename]['children'].append(child_type)
 
 	def get(self, name):
-		return self.mapping[name]
+		return self.mapping.get(name, None)
```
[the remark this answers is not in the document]
Jaguar0625: i was trying to generate just code for block.cats, which includes transaction.cats (but no concrete transactions). 
this works around that. 

i left changes because i thought it would be nice if generator would not crash in that case, but could revert as we don't _really_ expect that anywhere.

sdk/python/generator/FactoryFormatter.py
```diff
@@ -62,14 +63,17 @@ def get_deserialize_descriptor(self):
 		body += f'{self.printer.name} = {self.printer.load()}\n'
 
 		body += 'mapping = {\n'
-		names = [f'{concrete.typename}' for concrete in self.factory_descriptor['children']]
-		body += indent(
-			',\n'.join(map(self.create_discriminator, names))
-		)
+
+		if self.factory_descriptor:
```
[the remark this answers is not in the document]
Jaguar0625: same as above

sdk/python/generator/Generator.py
Jaguar0625: > ok, I assume everything is moved to printers stage, which seems like a sane thing to do

yep. that made it easier for me to follow.

sdk/python/generator/Generator.py
Jaguar0625: benefit of moving to printers stage, is that we don't need these fixups anymore.

sdk/python/generator/Generator.py
Jaguar0625: i looked again and was adding wrong field. 😅 - it works now.

gimre-xymcity (approved):
🎉 🚀 🌑

PR #46, opened 2022-01-21
[jenkins] feature: run CI pipeline in docker

Jaguar0625:
provides better isolation

### DIFF
--- a/client/rest/Jenkinsfile
+++ b/client/rest/Jenkinsfile
@@ -1,7 +1,9 @@
 defaultCiPipeline {
 	platform = ['ubuntu']
-	gitHubId = 'Symbol-Github-app'
+	ciBuildDockerfile = 'javascript.Dockerfile'
+
 	packageId = 'client-rest'
+
 	publisher = 'docker'
-	docker_image_name = 'symbolplatform/symbol-rest'
+	// dockerImageName = 'symbolplatform/symbol-rest'
 }
--- a/jenkins/docker/javascript.Dockerfile
+++ b/jenkins/docker/javascript.Dockerfile
@@ -1,0 +1,18 @@
+FROM ubuntu:20.04
+
+# install tzdata first to prevent 'geographic area' prompt
+RUN apt-get update >/dev/null \
+	&& apt-get install -y tzdata \
+	&& apt-get install -y git curl
+
+# nodejs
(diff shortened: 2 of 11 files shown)


jenkins/docker/javascript.Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+RUN curl -fsSL https://deb.nodesource.com/setup_16.x | bash - \
+	&& apt-get install -y nodejs
+
+# rest dependencies
+RUN apt-get install -y make gcc g++
+
+# add ubuntu user (used by jenkins)
+RUN useradd -ms /bin/bash ubuntu
+
+USER ubuntu
```
gimre-xymcity, line 18: we'll discuss later, but neither of those two should be needed
(i.e. USER always forces given uid, and we probably want to use `--user` from outside like we did)

gimre-xymcity (changes requested):
rest of the change looks ok

client/rest/Jenkinsfile
```diff
@@ -1,7 +1,9 @@
 defaultCiPipeline {
 	platform = ['ubuntu']
-	gitHubId = 'Symbol-Github-app'
+	ciBuildDockerfile = 'javascript.Dockerfile'
+
 	packageId = 'client-rest'
+
 	publisher = 'docker'
-	docker_image_name = 'symbolplatform/symbol-rest'
+	// dockerImageName = 'symbolplatform/symbol-rest'
```
Wayonb, line 8: What's the plan for publishing images?

Wayonb (approved):

client/rest/Jenkinsfile
Jaguar0625: idk, i just noticed that wasn't being used (and it was snake_case), so commented it out for now

jenkins/shared-library/vars/defaultCiPipeline.groovy
```diff
@@ -63,6 +67,7 @@ void call(Closure body) {
 						steps {
 							println("Parameters: ${params}")
 							sh 'printenv'
+							sh 'ls -alh /home/ubuntu'
```
gimre-xymcity, line 70: drop ^^

gimre-xymcity (approved):

PR #48, opened 2022-01-21
[linters] task: tabify missed files, rename yml to yaml

Jaguar0625:


sdk/python/symbolchain/symbol/ExtendedTypeParsingRules.py
```diff
@@ -6,8 +6,7 @@
 def extend_type_parsing_rules(symbol_type_converter, base_type_parsing_rules=None):
 	builder = TypeParserBuilder('sc', symbol_type_converter, base_type_parsing_rules)
 
-	# note, some of those below could be automatically generated,
-	# but right now symbol.yml contains also state-related objects
+	# note, some of those below could be automatically generated
```
Jaguar0625, line 9: @gimre-xymcity - not sure what this comment should say as there is no symbol.y(a)ml

linters/scripts/find_space_indents.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+set -ex
+
+! rg --files-with-matches '^  ' "$(git rev-parse --show-toplevel)" | grep -vE ".json|.md|.LESSER|.eslintrc|.yaml|.txt"
```
Jaguar0625, line 5: @Wayonb - not sure if we should integrate this into CI as we currently don't have place to run something across all files in repo and/or that ALWAYS runs 🤔

sdk/javascript/src/nem/external/tweetnacl-nacl-fast-keccak.js
```diff
@@ -12,9 +12,9 @@ const { keccak_512 } = require('js-sha3');
 // See for details: http://tweetnacl.cr.yp.to/
```
gimre-xymcity: as mentioned, I would not touch this file at all...

sdk/python/symbolchain/symbol/ExtendedTypeParsingRules.py
gimre-xymcity: it's fine, we could generate (some/ most) of the rules based on AST iterate through all things, 
 - for pods, flags and enums we could probably always fire rule
 - for structs - only those that are in hints of transactions

sdk/javascript/src/nem/external/tweetnacl-nacl-fast-keccak.js
Jaguar0625: i think it's low risk (most diffs can ignore whitespace) and we did same to external one in py 😅

linters/scripts/find_space_indents.sh
Wayonb: Make sense.  Is the rg tool part of the new docker images?  If so we could add another step or as part of lint.

sdk/javascript/src/nem/external/tweetnacl-nacl-fast-keccak.js
gimre-xymcity: 😬

gimre-xymcity (approved):

PR #49, opened 2022-01-21
[sdk/python] task: drop intermediary Field and virtual objects

Jaguar0625:
attach metadata to AST objects via AstExtensions instead


sdk/python/generator/Generator.py
```diff
@@ -10,122 +10,89 @@
...
-			size_field = struct_type.get_field_by_name(size_field_name)
-			size_field.set_bound_field(field)
-
-		if field.ast_model.is_size_reference:
-			struct_field = struct_type.get_field_by_name(field.ast_model.value)
-			field.set_bound_field(struct_field)
+	for field_model in struct_model.fields:
+		if field_model.display_type.is_array and isinstance(field_model.size, str):
+			size_field_name = field_model.size
+			size_field_model = next(field2_model for field2_model in struct_model.fields if size_field_name == field2_model.name)
```
gimre-xymcity, line 52: would extract the inline (same below)

(or maybe better: move it to AstExtensions:?)

sdk/python/generator/Generator.py
```diff
@@ -10,122 +10,89 @@
...
 	# build map of types
-	type_map = {}
-	abstract_impl_map = AbstractImplMap()
-
-	types = []
-	for ast_model in ast_models:
-		instance = to_virtual_type_instance(ast_model)
-		types.append(instance)
-		type_map[ast_model.name] = instance
+	type_map = {ast_model.name: ast_model for ast_model in ast_models}
```
gimre-xymcity, line 89: not sure about ast_model naming in this method, but don't mind much, so up to you

gimre-xymcity (approved):

sdk/python/generator/Generator.py
Jaguar0625: i don't mind it since it's called by the generate entry point that is using that naming

PR #53, opened 2022-01-24
feat: Enable Codeql for Python and JavaScript

Wayonb:
Enable Codeql for Python and JavaScript for testing.  If looks good then will enable for cpp.

### DIFF
--- a/.github/workflows/codeql-analysis.yaml
+++ b/.github/workflows/codeql-analysis.yaml
@@ -1,0 +1,61 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,70 @@
+# For most projects, this workflow file will not need changing; you simply need
```
Jaguar0625, line 1: if possible, use yaml extension instead of yml

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,70 @@
...
+    # ✏️ If the Autobuild fails above, remove it and uncomment the following three lines
+    #    and modify them (or add more) to build your code if your project
+    #    uses a compiled language
+
+    #- run: |
+    #   make bootstrap
+    #   make release
+
+    - name: Perform CodeQL Analysis
+      uses: github/codeql-action/analyze@v1
```
Jaguar0625, line 70: run yaml linter on file

.github/workflows/codeql-analysis.yml
Jaguar0625: i see now i was wrong and github probably only picks up yml extension 😬

gimre-xymcity (approved):

PR #56, opened 2022-01-25
[monorepo] task: add dependabot.yml

Jaguar0625:


Wayonb (approved):

linters/scripts/find_space_indents.sh
```diff
@@ -2,4 +2,12 @@
 
 set -ex
 
-! rg --files-with-matches '^  ' "$(git rev-parse --show-toplevel)" | grep -vE ".json|.md|.LESSER|.eslintrc|.yaml|.txt"
+! rg \
+		--files-with-matches \
+		--hidden \
```
gimre-xymcity, line 7: ok I think I get what you were saying earlier, I think solution to this is to put dirs in `.ignore` file (or `.gitignore`)
(otherwise ripgrep will be traversing whole .git and that's waste of time ^^)

syntax is same as .gitignore, so I think we can do something like this:
`!.github`

gimre-xymcity (approved):

PR #68, opened 2022-01-26
[monorepo]: add subprojects to dependabot and update libs

Jaguar0625:


sdk/python/requirements.txt
```diff
@@ -1,7 +1,7 @@
-cryptography==3.4.6
+cryptography==36.0.1
```
gimre-xymcity, line 1: is this _right_ ? o0 is this same package actually?

gimre-xymcity: hhmmmm, wth o0

image

gimre-xymcity (approved):

sdk/python/requirements.txt
Jaguar0625: (probably a typo going from 3.4 => 3.5 ^^)

PR #74, opened 2022-01-28
[monorepo] task: use poetry for publishing python packages

Jaguar0625:


catbuffer/parser/pyproject.toml
```diff
@@ -0,0 +1,20 @@
+[tool.poetry]
+name = 'catparser'
+version = '3.1.0'
```
Jaguar0625, line 3: using existing version, but we'll need to bump before publishing

catbuffer/parser/pyproject.toml
```diff
@@ -0,0 +1,20 @@
...
+packages = [{ include = 'catparser' }]
+
+repository = 'https://github.com/symbol/symbol/tree/main/catbuffer/parser'
+
+keywords = ['symbol', 'catbuffer', 'catparser', 'parser', 'catbuffer-parser']
+
+classifiers = ['Programming Language :: Python :: 3.7']
+
+[tool.poetry.dependencies]
+python = "^3.7"
```
Jaguar0625, line 20: i don't see how to add obsoletes (but presumably not needed here)

jenkins/shared-library/resources/artifacts/scripts/pypi-publish.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
Jaguar0625, line 1: small enough, that i think bash (instead of py) is ok

jenkins/shared-library/resources/artifacts/scripts/pypi-publish.sh
```diff
@@ -0,0 +1,20 @@
...
+
+set -ex
+
+# prepare toml (add requirements)
+while IFS= read -r line; do poetry add "$line"; done < requirements.txt
+cat pyproject.toml
```
Jaguar0625, line 7: unsure why this isn't showing output on jenkins

jenkins/shared-library/resources/artifacts/scripts/pypi-publish.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+set -ex
+
+# prepare toml (add requirements)
+while IFS= read -r line; do poetry add "$line"; done < requirements.txt
```
Jaguar0625, line 6: this is adding all the requirements to poetry from requirements.txt (this allows us to have one list)

aside: `cat requirements.txt | poetry add` but shellcheck doesn't like that

jenkins/shared-library/vars/defaultCiPipeline.groovy
```diff
@@ -185,16 +186,16 @@ void call(Closure body) {
 				stages {
 					stage('publish RC') {
 						when {
-							anyOf {
-								branch env.DEV_BRANCH
-								allOf {
+							allOf {
```
Jaguar0625, line 189: since packages with same version can't be published multiple times, i think we only want to publish on manual builds, else we'll get lots of build problems if we don't auto increment  (and i don't see good reason for that).

unsure if we should restrict test/alpha publishes from dev branch

jenkins/shared-library/vars/publish.groovy
```diff
@@ -54,15 +52,23 @@ void npmPublisher(Map config, String phase) {
...
+	String scriptFullFilepath = fileHelper.copyToTempFile(getResource('artifacts/scripts/pypi-publish.sh'), 'scripts/pypi-publish.sh')
+	withCredentials([
+		usernamePassword(
+			credentialsId: credentialsId,
 			usernameVariable: 'TWINE_USERNAME',
-			passwordVariable: 'TWINE_PASSWORD')]) {
-				runScript('python3 -m twine upload --repository pypi dist/*')
-		}
+			passwordVariable: 'POETRY_PYPI_TOKEN_PYPI')
+	]) {
```
Jaguar0625, line 70: as discussed, should change this to use single credential (like npm)

sdk/python/pyproject.toml
```diff
@@ -0,0 +1,20 @@
+[tool.poetry]
+name = 'symbol-sdk-python'
```
Jaguar0625, line 2: dropped core, but problematic in that we need obsoletes somehow 🤔

sdk/python/pyproject.toml
```diff
@@ -0,0 +1,20 @@
...
+maintainers = ['Symbol Contributors <contributors@symbol.dev>']
+license = 'MIT'
+
+readme = 'README.md'
+
+packages = [{ include = 'symbolchain' }]
+
+repository = 'https://github.com/symbol/symbol/tree/main/sdk/python'
+
+keywords = ['symbol', 'sdk', 'sdk-core', 'Symbol SDK', 'Symbol SDK core']
```
Jaguar0625, line 15: not sure if we need "core" keywords

jenkins/shared-library/vars/publish.groovy
Wayonb: Updated the pypi creds to be secret text.  You should update them to be like npm

sdk/python/pyproject.toml
gimre-xymcity: as I've mentioned, I think obsoletes is only needed _once_ when releasing initial package that replaces different package (and there is NAME change involved), besides that obsoletes should not be needed)

sdk/python/pyproject.toml
gimre-xymcity: probably _not_ (I'd throw them away)

only reason we have them is that we've been discussed 2-level sdks, where one is low - txes, signing etc
other is high and can interact with rest servers

gimre-xymcity (approved):

Wayonb (approved):

PR #84, opened 2022-02-01
[sdk/javascript] add common TransactionFactory code

Jaguar0625:
port TransactionDescriptorProcessor and RuleBasedTransactionFactory

### DIFF
--- a/sdk/javascript/src/RuleBasedTransactionFactory.js
+++ b/sdk/javascript/src/RuleBasedTransactionFactory.js
@@ -1,0 +1,196 @@
+const { BaseValue } = require('./BaseValue');
+const { ByteArray } = require('./ByteArray');
+const { TransactionDescriptorProcessor } = require('./TransactionDescriptorProcessor');
+
+const buildEnumStringToValueMap = EnumClass => new Map(Object.getOwnPropertyNames(EnumClass)
+	.filter(name => name.toUpperCase() === name)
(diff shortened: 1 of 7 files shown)


sdk/javascript/src/RuleBasedTransactionFactory.js
```diff
@@ -0,0 +1,195 @@
...
+			return enumValue;
+		});
+	}
+
+	/**
+	 * Creates struct parser (to allow nested parsing).
+	 * @param {string} name Class name.
+	 */
+	addStructParser(name) {
+		const StructClass = this._getModuleClass(name);
```
gimre-xymcity, line 135: Not sure if lint allows that, but wondering if we should use names like Struct_Class or StructClass in py version as well (instead of current struct_class)

sdk/javascript/src/RuleBasedTransactionFactory.js
```diff
@@ -0,0 +1,195 @@
...
+		});
+	}
+
+	/**
+	 * Creates array type parser, based on some existing element type parser.
+	 * @param {string} name Class name.
+	 */
+	addArrayParser(name) {
+		const elementRule = this.rules.get(name);
+		const elementName = 0 === name.indexOf('struct:') ? name.substring('struct:'.length) : name;
```
gimre-xymcity, line 155: guess`name.replace(/^struct:/, '')` ? ;)

PR #97, opened 2022-02-09
[monorepo] feat: checkout submodule

Wayonb:
Linters and Jenkins folder is shared across monorepos.  The Jenkins job needs to check out submodule to get access to the Dockerfile to run the pipeline.

### DIFF
--- a/jenkins/shared-library/vars/createMultibranchJob.groovy
+++ b/jenkins/shared-library/vars/createMultibranchJob.groovy
@@ -31,6 +31,7 @@ void call(Map jobConfiguration) {
 									gitHubBranchDiscovery {
 										strategyId(EXCLUDE_PULL_REQUESTS_STRATEGY_ID)
 									}
+
 									gitHubPullRequestDiscovery {
 										strategyId(USE_CURRENT_SOURCE_STRATEGY_ID)
 									}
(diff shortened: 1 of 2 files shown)


jenkins/shared-library/vars/defaultCiPipeline.groovy
```diff
@@ -298,3 +299,11 @@ String resolvePath(String rootPath, String path) {
 Boolean shouldPublishImage(String shouldPublish) {
 	return shouldPublish == null ? false : shouldPublish.toBoolean()
 }
+
+void runInitializeScriptIfPresent() {
+	String initFile = 'init.sh'
+	if (fileExists(initFile)) {
+		println 'Running initialized script'
```
Jaguar0625, line 306: initialize

jenkins/shared-library/vars/createMultibranchJob.groovy
```diff
@@ -43,6 +44,39 @@ void call(Map jobConfiguration) {
...
 									}
+
+									cleanBeforeCheckoutTrait {
+										extension {
+											// Deletes untracked submodules and any other subdirectories which contain .git directories.
+											deleteUntrackedNestedRepositories(true)
+										}
+									}
+
+									submoduleOptionTrait {
```
Jaguar0625, line 55: why is this needed ?
is this cloning the submodules ?

Wayonb: Yea, Jenkins needs access to the Dockerfiles in ``jenkins/docker`` to build the image before the stages in pipeline. I needed to clone the submodules.

Jaguar0625: is there a way to only clone a subset of the repo using sparsecheckout, like init.sh is doing?

Wayonb: I was looking last night but havent found a way yet.  Will continue looking today.  Hopefully with fresh eyes might have better luck :)

Wayonb: I havent found a good way to checkout the submodule using sparsecheckout.  The git plug doesnt support it for submodules(https://issues.jenkins.io/browse/JENKINS-42628)
Workaround would be
1. Use git hooks(post-checkout) which runs the init.sh
2. In the pipeline, create a new stage that will run init.sh script before the Dockerfile.  This will move the Dockerfile into the stages
3. Could probably enable ``Shallow clone`` to save some time and space.
@Jaguar0625 @gimre-xymcity any thoughts?

Jaguar0625: i think this is ok for now, but you can try shallow clone and see if it improves things. if so, you can make another PR with those changes. 👍

Jaguar0625 (approved):

PR #98, opened 2022-02-09
[sdk/python] allow descriptor-inlined cosignatures

gimre-xymcity:
## What's the issue?
Right now cosignatures in nem can't be used, when transaction  is created via descriptors

Added processing of `Cosignature` and `SizePrefixedCosignature` in nem's TransactionFactory

Tested using upcoming PR, since this work the same as any other structs, separate unittests are not needed.

### DIFF
--- a/sdk/python/symbolchain/nem/TransactionFactory.py
+++ b/sdk/python/symbolchain/nem/TransactionFactory.py
@@ -71,9 +71,9 @@ class TransactionFactory:
 		factory.autodetect()
 
 		struct_names = [
-			'Message', 'NamespaceId', 'MosaicId', 'Mosaic', 'SizePrefixedMosaic', 'MosaicLevy',
+			'Cosignature', 'Message', 'NamespaceId', 'MosaicId', 'Mosaic', 'SizePrefixedMosaic', 'MosaicLevy',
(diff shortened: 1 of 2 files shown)


sdk/python/symbolchain/nem/TransactionFactory.py
```diff
@@ -86,7 +86,8 @@ def _build_rules(base_type_parsing_rules):
 		for name, typename in sdk_type_mapping.items():
 			factory.add_pod_parser(name, typename)
 
-		for name in ['struct:SizePrefixedMosaic', 'struct:SizePrefixedMosaicProperty', 'struct:SizePrefixedMultisigAccountModification']:
-			factory.add_array_parser(name)
+		arrays = ['SizePrefixedMosaic', 'SizePrefixedMosaicProperty', 'SizePrefixedMultisigAccountModification', 'SizePrefixedCosignature']
```
Jaguar0625, line 89: consider renaming arrays => array_names

Jaguar0625 (approved):

sdk/python/symbolchain/nem/TransactionFactory.py
gimre-xymcity: ok, that's actually what I had earlier but it looked like it's crossing width boundary, but seems pycodestyle does not care 😁

PR #116, opened 2022-02-14
 [client/rest] build: fix security vulnerabilities by monkey patching dev deps

Jaguar0625:
tests use (seemingly) unmaintained hippie package, which has insecure dependencies.

## What's the issue?
dependabot warnings.

use `npx npm-force-resolutions` to update its dependencies to secure versions.
    
yes


Jaguar0625:
this didn't work 😅

PR #125, opened 2022-02-17
[sdk/python]: add 'create from descriptor' test vectors

Jaguar0625:
test vectors for catbuffers don't create objects.

## What's the issue?
low code coverage.

added 'create from descriptor' vectors.

yes


sdk/python/testvectors/__main__.py
```diff
@@ -0,0 +1,263 @@
+import argparse
```
gimre-xymcity: I'm guessing wrong base?

gimre-xymcity (approved):

sdk/python/testvectors/__main__.py
Jaguar0625: y

Jaguar0625:
merged into feature branch: 70567f7bbe2618fe98f62d009161077895c3238b

PR #126, opened 2022-02-17
[sdk] fix block handling

gimre-xymcity:
## What's the issue?
* `@alignment` was missing on `block.transactions`
* besides that, wrong code was generated because only `is_byte_constrained` was handled correctly

 * added missing attribute
 * modified both js and py generator appropriately

 * Manually verified that generated code seems fine and that only block handling has changed.

### DIFF
--- a/sdk/javascript/generator/printers.py
+++ b/sdk/javascript/generator/printers.py
@@ -60,34 +60,53 @@ class TypedArrayPrinter(Printer):
 	def get_default_value():
 		return '[]'
 
+	@property
+	def is_variable_size(self):
+		# note: self.descriptor.field_type.alignment condition at the end is needed to filter out receipts
+		descriptor = self.descriptor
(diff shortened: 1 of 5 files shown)


sdk/javascript/generator/printers.py
```diff
@@ -60,34 +60,53 @@ def get_type(self):
...
+		descriptor = self.descriptor
+		return (descriptor.field_type.is_byte_constrained or descriptor.extensions.is_contents_abstract) and descriptor.field_type.alignment
+
 	def get_size(self):
-		if self.descriptor.field_type.is_byte_constrained:
-			# note: use actual `.size` field
-			alignment = self.descriptor.field_type.alignment
-			return f'this.{self.name}.map(e => arrayHelpers.alignUp(e.size, {alignment})).reduce((a, b) => a + b, 0)'
+		# note: use actual `.size` field
+		align_element = f'arrayHelpers.alignUp(e.size, {self.descriptor.field_type.alignment})'
```
Jaguar0625, line 71: would move into if like below

Jaguar0625 (approved):

PR #130, opened 2022-02-18
[catbuffer/parser]: extend alignment attribute to specify padding of last element

Jaguar0625:
there is no way to specify padding of last element in an aligned array

## What's the issue?
within a block, the last transaction is NOT padded
within an aggregate, the last transaction is padded

add optional `[not] pad_last` qualifier to alignment attribute

yes

### DIFF
--- a/catbuffer/parser/catparser/AstPostProcessor.py
+++ b/catbuffer/parser/catparser/AstPostProcessor.py
@@ -24,7 +24,7 @@ class AstPostProcessor:
 					if not hasattr(field.field_type, attribute.name):
 						raise AstException(f'field {field.name} ({field.field_type}) does not have property {attribute.name}')
 
-					setattr(field.field_type, attribute.name, attribute.value)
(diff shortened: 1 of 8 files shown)


catbuffer/parser/catparser/AstPostProcessor.py
```diff
@@ -24,7 +24,7 @@ def apply_attributes(self):
 					if not hasattr(field.field_type, attribute.name):
 						raise AstException(f'field {field.name} ({field.field_type}) does not have property {attribute.name}')
 
-					setattr(field.field_type, attribute.name, attribute.value)
+					setattr(field.field_type, attribute.name, attribute.values)
```
Jaguar0625, line 27: since qualifiers are attached to alignment property, passing down all values. this can look a bit awkward in tests, `foo.sort_key = ['abc']`.

gimre-xymcity (approved):

PR #134, opened 2022-02-18
[sdk/python] block test vectors (receipts will be next time)

gimre-xymcity:
 * So far we did not have test vectors for whole blocks

## What's the issue?
 * Adding this to improve coverage.

 * #126 + #132 
 * this changes adds block recipes + handling inside generator,
  
 *  generated test vectors pass, but that might be luck, so would be good to compare/cross-check with actual server-generated block.


sdk/python/symbolchain/symbol/TransactionFactory.py
```diff
@@ -61,7 +68,20 @@ def _build_rules(type_rule_overrides):
...
+		for name in struct_names:
+			factory.add_struct_parser(name)
+
+		byte_arrays = {
+			'ProofGamma': sc.ProofGamma,
+			'ProofVerificationHash': sc.ProofVerificationHash,
+			'ProofScalar': sc.ProofScalar
+		}
+		for name, typename in byte_arrays.items():
+			factory.add_pod_parser(name, typename)
```
gimre-xymcity, line 84: not sure if those should be merged with sdk_type_mapping below

sdk/python/testvectors/__main__.py
```diff
@@ -192,52 +211,79 @@ def create_entry(self, schema_name, test_name, factory, descriptor):
...
+
+			recipes = getattr(module_descriptor[1], 'block_recipes')
+			entries.extend(self.create_blocks(module_descriptor[0], recipes))
 
 			print(f'[+] module {self.network_name}.{module_descriptor[0]}: ok')
 
 		return entries
 
 
+def dump_to_file(entries, filepath):
```
gimre-xymcity, line 275: 😁

PR #156, opened 2022-03-11
[jenkins] feat: catapult code coverage job

Wayonb:
There is no code coverage job for catapult client

## What's the issue?
Currently code coverage for the catapult client is generated manually.

created a daily code coverage job for the catapult client.
It is base on gcc 10 until update all the test images

Yes

### DIFF
--- a/jenkins/catapult/configuration.py
+++ b/jenkins/catapult/configuration.py
@@ -31,7 +31,7 @@ def load_compiler_configuration(filepath):
 
 		sanitizers = configuration_yaml['sanitizers'].split(',') if 'sanitizers' in configuration_yaml else []
 		architecture = configuration_yaml['architecture']
-		enable_code_coverage = configuration_yaml['enable_code_coverage'] if 'enable_code_coverage' in configuration_yaml else False
+		enable_code_coverage = configuration_yaml.get('enable_code_coverage', False)
 
 		configuration_keys = ['compiler', 'stl', 'sanitizers', 'architecture', 'enable_code_coverage']
 		return namedtuple('CompilerConfiguration', configuration_keys)(compiler, stl, sanitizers, architecture, enable_code_coverage)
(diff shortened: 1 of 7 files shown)


jenkins/catapult/jenkins/catapult-client-build-catapult-project-daily.groovy
```diff
@@ -73,6 +73,13 @@ pipeline {
 						}
 					}
 				}
+				stage('code coverage') {
+					steps {
+						script {
+							dispatch_build_job('gcc-10-code-coverage', 'tests-diagnostics')
```
Jaguar0625, line 79: why tests-diagnostics? would probably not use that for cc build

Wayonb: no specific reason.  will change to ``tests-metal``

jenkins/catapult/configuration.py
```diff
@@ -31,9 +31,10 @@ def load_compiler_configuration(filepath):
 
 		sanitizers = configuration_yaml['sanitizers'].split(',') if 'sanitizers' in configuration_yaml else []
 		architecture = configuration_yaml['architecture']
+		enable_code_coverage = configuration_yaml['enable_code_coverage'] if 'enable_code_coverage' in configuration_yaml else False
```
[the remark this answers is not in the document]
Wayonb: use the get() with a default value of False.  bool() seem to return True for any string which is not empty.  
``enable_code_coverage: nope`` would enable code coverage

Jaguar0625 (approved):

jenkins/catapult/runDockerBuildInnerBuild.py
```diff
@@ -158,7 +159,7 @@ def main():
 
 	builder = BuildManager(args, process_manager, environment_manager)
 	env = LinuxEnvironment(builder.use_conan, process_manager, environment_manager)
-	env.prepare()
+	env.prepare(builder.enable_code_coverage, args.source_path)
```
gimre-xymcity, line 162: why is dest dir `source_path/_build`? 😬 
(isn't it possible to make build inside /tmp, like earlier?)

gimre-xymcity (approved):

PR #179, opened 2022-04-06
[monorepo] feat: Auto-merge of dependabot prs

Wayonb:
Dependabot PRs are manually merge into GitHub

## What's the issue?
Merge of PR is manual

Add a GitHub action which will auto merge all PRs from dependabot after the CI checks pass.

Note: commit-message prefix field is limited to 15 characters, monorepo was used in such case.

Tested the auto merge of Dependabot PR

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -8,6 +8,8 @@ updates:
       day: sunday
     target-branch: dev
     labels: [catbuffer-parser]
+    commit-message:
+      prefix: '[monorepo] fix'
 
   - package-ecosystem: pip
     directory: /client/catapult/scripts
@@ -16,6 +18,8 @@ updates:
       day: sunday
     target-branch: dev
     labels: [catapult]
+    commit-message:
+      prefix: '[monorepo] fix'
 
   - package-ecosystem: npm
     directory: /client/rest
@@ -25,6 +29,8 @@ updates:
     target-branch: dev
     labels: [rest]
     versioning-strategy: increase
+    commit-message:
+      prefix: '[monorepo] fix'
 
   - package-ecosystem: pip
     directory: /jenkins/catapult
@@ -33,6 +39,8 @@ updates:
       day: sunday
     target-branch: dev
     labels: [jenkins]
+    commit-message:
+      prefix: '[jenkins] fix'
 
   - package-ecosystem: pip
     directory: /linters/cpp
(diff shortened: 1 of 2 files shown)


.github/dependabot.yaml
```diff
@@ -8,6 +8,8 @@ updates:
       day: sunday
     target-branch: dev
     labels: [catbuffer-parser]
+    commit-message:
+      prefix: '[monorepo] fix'
```
Jaguar0625, line 12: 1. why monorepo
2. will this pass commit message linter ?

.github/dependabot.yaml
```diff
@@ -66,6 +80,8 @@ updates:
       day: sunday
     target-branch: dev
     labels: [sdk-javascript]
+    commit-message:
+      prefix: '[sdk] fix'
```
Jaguar0625, line 84: sdk/javascript ?

.github/workflows/auto-merge.yaml
```diff
@@ -0,0 +1,32 @@
+name: Dependabot auto-approve-merge
+
+# yamllint disable-line rule:truthy
```
Jaguar0625, line 3: can you quote value instead?

.github/dependabot.yaml
Wayonb: 1. I noted above that the commit-message prefix field is limited to 15 characters.  ``[catbuffer/parser] fix`` is too long.  Do we want to create different tag?
2. It should create a commit title of ``[monorepo] fix: xxxxxxx``` where xxxxxxx is the original commit title.

.github/dependabot.yaml
Wayonb: Same limitation on prefix field

.github/workflows/auto-merge.yaml
Wayonb: hmm....not sure if you can quote the ``on`` but I will try

.github/dependabot.yaml
Jaguar0625: i would drop fix ?

.github/workflows/auto-merge.yaml
Wayonb: cool.  Added quotes worked

.github/dependabot.yaml
Wayonb: Let me see which ones will fit.

Jaguar0625 (approved):

PR #212, opened 2022-05-10
[monorepo] fix: trigger all jobs on configuration change

Wayonb:
When the build configuration file is updated all job are not run.

## What's the issue?
When the build configuration file is updated need to verify that all job still runs

Trigger all Jenkins jobs for the repo when the build configuration is updated.

Yes

### DIFF
--- a/jenkins/shared-library/vars/monorepoSeedPipeline.groovy
+++ b/jenkins/shared-library/vars/monorepoSeedPipeline.groovy
@@ -67,7 +67,7 @@ void call(Closure body) {
 }
 
 String resolveBuildConfigurationFile()  {
-	filepath = helper.resolveBuildConfigurationFile()
+	String filepath = helper.resolveBuildConfigurationFile()
 	logger.logInfo("build configuration file: ${filepath}")
 	return filepath
 }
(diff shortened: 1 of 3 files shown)


jenkins/shared-library/vars/monorepoSeedPipeline.groovy
```diff
@@ -69,8 +66,8 @@ void call(Closure body) {
 	}
 }
 
-String resolveBuildConfigurationFile(String configurationFile)  {
-	String filepath = Paths.get('.github').resolve(configurationFile)
+String resolveBuildConfigurationFile()  {
+	filepath = helper.resolveBuildConfigurationFile()
```
Jaguar0625, line 70: does this pass groovy linter? i thought you would need explicit type?

Jaguar0625 (approved):

jenkins/shared-library/vars/monorepoSeedPipeline.groovy
Wayonb: yea, it pass linter.
I will update it with the type.

PR #226, opened 2022-05-20
[monorepo] fix: python local apps not in PATH

Wayonb:
Python apps that gets installed in local, are not in the default PATH.

## What's the issue?
Failed to run coverage python app

Add local bin to the PATH so local apps can be run.

Yes


0x6861746366574:
image

🗞️

Jaguar0625 (approved):
you can squash into previous commit in dev

PR #234, opened 2022-05-23
[monorepo] task: add jacoco support

Wayonb:
There is no support for Java code coverage.

## What's the issue?
Java code coverage is not getting verified.

Jacoco option is added to verify the code coverage meets what is configured in the pom.xml

### DIFF
--- a/jenkins/shared-library/vars/codeCoverage.groovy
+++ b/jenkins/shared-library/vars/codeCoverage.groovy
@@ -11,17 +11,26 @@ void uploadCodeCoverage(String flag) {
 	}
 }
 
+void logCodeCoverageMinimum(Integer minimumCodeCoverage) {
+	logger.logInfo("Minimum code coverage is ${minimumCodeCoverage}")
+}
+
 void verifyCodeCoverageResult(String tool, Integer minimumCodeCoverage) {
 	Map codeCoverageCommand = [
-		'coverage': { target ->
+		'coverage': { Integer target ->
+			logCodeCoverageMinimum(target)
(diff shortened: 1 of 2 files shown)


jenkins/shared-library/vars/codeCoverage.groovy
```diff
@@ -11,17 +11,26 @@ void uploadCodeCoverage(String flag) {
...
 			runScript("coverage report --fail-under=${target}")
 		},
-		'nyc': { target ->
+		'nyc': { Integer target ->
+			logCodeCoverageMinimum(target)
 			runScript('npx nyc@latest report --lines')
 			runScript("npx nyc@latest check-coverage --lines ${target}")
+		},
+		'jacoco': { Integer target ->
+			logger.logInfo('Minimum code coverage is set pom.xml')
```
Jaguar0625, line 31: set *in* pom? should you read from pom and display result here?

Wayonb: yea, I was thinking that also.
Or to keep the code coverage consistent across all tools, take the minimumCodeCoverage and overwrite the value in the pom.  Any preference?

Jaguar0625: overwriting is fine too, but in that case would probably put 0 in pom with comment

Wayonb: Will read the value and display it here.  Thanks.

Jaguar0625 (approved):

Issue #235, opened 2022-05-26
[sdk/python] Functionality to check if a (string) address is valid


cryptoBeliever:
## Current Behaviour

There is no functionality to check if a string address is valid. Currently, the only way to check is to construct an `Address` object and catch if any error is raised.

```
try:
     --- construct address ---
    return true
except Exception as e:
    return false   
```

In symbolblockchain/Network.py there is a method `is_valid_address` but it expects the Address object to be passed as an argument. 

## Expected Behaviour

There should be a function to check if a string address is valid in the SDK.

Issue #214, opened 2022-05-12
[sdk/javascript] Functionality to check if a (string) address is valid


yilmazbahadir:
## Current Behaviour

There is no functionality to check if a string address is valid. Currently, the only way to check is to construct an `Address` object and catch if any error is thrown and this is less than _ideal_.

```
try {
	const address = new NemFacade.Address(searchVal);
	// const address = new SymbolFacade.Address(searchVal);

	return [address.toString(), '', ''];
} catch (e) {
	// handle failure
}
```

It's also worth mentioning that the following method could work if it expected a `string` instead of an `Address`  object (not sure if constructing a Network object is a good idea to check the validity of an address - since it's an instance method)
https://github.com/symbol/symbol/blob/ab98b8c37daa93d1a5620722d10457a27d0156bd/sdk/javascript/src/Network.js#L47 

## Expected Behaviour

There should be a function to check if a string address is valid in the SDK.

PR #236, opened 2022-05-27
[sdk/*] add function for checking address string validity to sdks

Jaguar0625:
[sdk/javascript] task: add isValidAddressString to network classes
    
     problem: there is no functionality to check if an address string is valid
              the existing isValidAddress requires a typed Address, which implies a properly formed string
    solution: add isValidAddressString to Network classes that accepts and validates string paramteter
    
       issue: #214

    [sdk/python] task: add is_valid_address_string to network classes
    
     problem: there is no functionality to check if an address string is valid
              the existing is_valid_address requires a typed Address, which implies a properly formed string
    solution: add is_valid_address_string to Network classes that accepts and validates string paramteter
    
       issue: #235

### DIFF
--- a/sdk/python/symbolchain/Network.py
+++ b/sdk/python/symbolchain/Network.py
@@ -1,14 +1,17 @@
 import hashlib
 from abc import abstractmethod
 
+BASE32_RFC4648_ALPHABET = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ234567'
+
 
 class Network:
 	"""Represents a network."""
 
-	def __init__(self, name, identifier):
+	def __init__(self, name, identifier, address_class):
 		"""Creates a new network with the specified name and identifier byte."""
 		self.name = name
 		self.identifier = identifier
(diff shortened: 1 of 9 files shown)


sdk/python/symbolchain/Network.py
```diff
@@ -28,6 +29,16 @@ def public_key_to_address(self, public_key):
 
 		return self.create_address(version, checksum)
 
+	def is_valid_address_string(self, address_string):
+		"""Checks if an address string is valid and belongs to this network."""
+		if self.address_class.ENCODED_SIZE != len(address_string):
+			return False
+
+		if any(ch not in 'ABCDEFGHIJKLMNOPQRSTUVWXYZ234567' for ch in address_string):
```
gimre-xymcity, line 37: would probably throw it into a 'const' and name it BASE32_RFC4648_ALPHABET (just for the reference, that it's not custom one)

gimre-xymcity (approved):

PR #237, opened 2022-05-27
[rest] build: switch js-sha3 to noble

Jaguar0625:
problem: js sdk has been upgraded to use noble for hashing but rest is still using js-sha3

solution: replace js-sha3 with noble


gimre-xymcity (approved):
👍

PR #239, opened 2022-05-28
[jenkins] fix: update submodule

Wayonb:
Every update to the Linters folder requires the submodule to be updated on every repo.

## What's the issue?
Jenkins job does not use the latest configuration on every repo.

1. Update submodule to the latest commit.  This will allow the Jenkins job to always use the latest configuration.
2. Remove the limit on the line length in the body

Yes, by running a job on product repo where submodule is not the latest.
The Jenkins job used the latest commit.


jenkins/shared-library/vars/defaultCiPipeline.groovy
```diff
@@ -67,6 +67,11 @@ void call(Closure body) {
 		}
 
 		stages {
+			stage('setup environment') {
+				steps {
+					runScript('git submodule update --remote')
+				}
+			}
```
Jaguar0625, line 74: why do we need this here in addition to helper.groovy?

linters/git/.gitlint
```diff
@@ -1,5 +1,5 @@
 [general]
-ignore=body-is-missing
+ignore=body-is-missing, body-max-line-length
```
Jaguar0625, line 2: i think we're using 140 line limit in all of our code files? are we sure we want to disable this check?

Wayonb: We can leave it.  I was only thinking of removing it since some of the dependabot pr was hitting it.

jenkins/shared-library/vars/defaultCiPipeline.groovy
Wayonb: The one in the helper is only for manual jobs.  This one will always run.
I was not sure if the ``git reset hard`` also reset the submodule.
From my testing it seems it does not but not sure.

Jaguar0625 (approved):

gimre-xymcity (approved):

Issue #101, opened 2022-02-10
[catbuffer/parser] pass down information about aligned/unaligned structs


gimre-xymcity:
All fields in symbol txes structs are aligned on word boundary. That however is not true for 'state' structs.
Generators could use the information to optimize access.

gimre-xymcity:
catbuffer part was released with client/catapult/v1.0.3.2
js change is scheduled with sdk 3.0.4

PR #281, opened 2022-06-28
[sdk/javascript] optimized JS field deserialization

gimre-xymcity:
> Note: This is the default pull request template.

## What is the current behavior? / What's the issue?

JS generated code is using universal struct reading, that is not leveraging symbol structs alignment

struct generator is calling `field.extensions.printer.load`.
In case of simple types, we could just handle it inside the printer, howevere in case of builtin fields (i.e. Amount) BuiltinPrinter does not have info about field itself, therefore wouldn't be able to access parent - aligned - struct 

Only sensible way to do it was to pass down `is_aligned` in calls to load() 

## What was NOT changed:

right now missing piece is handling of structs, so i.e. (from HashlockTransaction.deserialize)
```js
		const fee = Amount.deserializeAligned(view.buffer);
		view.shiftRight(fee.size);
		const deadline = Timestamp.deserializeAligned(view.buffer);
		view.shiftRight(deadline.size);
		const mosaic = UnresolvedMosaic.deserialize(view.buffer);
		view.shiftRight(mosaic.size);
```

note that `UnresolvedMosaic.deserialize` is called. otherwise it would require generating deserializeAligned on all structs, I'm not quite sure if that is _wanted_ change, if so, should probably be done separately

 * verified generated code
 * catbuffer test vectors are passing
 * additional unit tests added


gimre-xymcity:
This change should fix #101

sdk/javascript/generator/printers.py
```diff
@@ -29,9 +29,14 @@ def get_default_value(self):
 	def get_size(self):
 		return self.descriptor.size
 
-	def load(self, buffer_name='byteArray'):
+	def load(self, buffer_name='byteArray', is_aligned=False):
 		data_size = self.get_size()
-		return f'converter.bytesToInt({buffer_name}, {data_size}, {js_bool(not self.descriptor.is_unsigned)})'
+		arguments = f'{buffer_name}, {data_size}, {js_bool(not self.descriptor.is_unsigned)}'
+		# is_aligned - handles both generation of deserializeAligned for pod and enum types as long as generation of fields within struct
```
Jaguar0625, line 35: missing ending?

sdk/javascript/generator/printers.py
```diff
@@ -217,13 +224,17 @@ def get_default_value(self):
...
-		if DisplayType.STRUCT == self.descriptor.display_type and self.descriptor.is_abstract:
+	def load(self, buffer_name='view.buffer', is_aligned=False):
+		display_type = self.descriptor.display_type
+		if DisplayType.STRUCT == display_type and self.descriptor.is_abstract:
 			# HACK: factories use this printers as well, ignore them
 			if 'parent' != self.name:
 				factory_name = self.get_type() + 'Factory'
 				return f'{factory_name}.deserialize({buffer_name})'
 
+		if is_aligned and display_type in [DisplayType.INTEGER, DisplayType.ENUM]:
```
Jaguar0625, line 235: tuple ? (i forget if there's a reason we use list instead of tuple in many places 😅 )

sdk/javascript/generator/TypeFormatter.py
```diff
@@ -65,10 +65,17 @@ def generate_ctor(self):
...
-		method_descriptor = self.provider.get_deserialize_descriptor()
-		method_descriptor.method_name = 'static deserialize'
+	def generate_deserializer(self, name, generated_name=None):
+		if not hasattr(self.provider, f'get_{name}_descriptor'):
+			return None
+
+		descriptor_getter = getattr(self.provider, f'get_{name}_descriptor')
+		method_descriptor = descriptor_getter()
+		if not method_descriptor:
+			return None
```
Jaguar0625, line 75: when can this return None when function exists (previous check passes)?

sdk/javascript/test/utils/converter_spec.js
```diff
@@ -274,6 +274,50 @@ describe('converter', () => {
...
+			assertBytesToIntOnMisaligned([0xFF, 0x11, 0x22, 0x33, 0x80], 4, false);
+			assertBytesToIntOnMisaligned([0xFF, 0x11, 0x22, 0x33, 0x44, 0x55, 0x66, 0x77, 0x80], 4, true);
+			assertBytesToIntOnMisaligned([0xFF, 0x11, 0x22, 0x33, 0x44, 0x55, 0x66, 0x77, 0x80], 4, false);
+		});
+	});
+
+	describe('bytesToIntUnaligned', () => {
+		assertBytesToInt(converter.bytesToIntUnaligned);
+
+		const assertBytesToIntUnalignedOnMisaligned = (input, size, isSigned, expectedValue) => {
```
Jaguar0625, line 304: Unaligned in name is probably redundant

sdk/javascript/generator/EnumTypeFormatter.py
```diff
@@ -28,14 +28,20 @@ def get_ctor_descriptor(self):
...
 		if self.enum_type.is_bitwise:
-			body += f'return new {self.typename}({self.int_printer.load()});'
+			body += f'return new {self.typename}({self.int_printer.load("byteArray", is_aligned)});'
 		else:
-			body += f'return this.fromValue({self.int_printer.load()});'
+			body += f'return this.fromValue({self.int_printer.load("byteArray", is_aligned)});'
 		return MethodDescriptor(body=body)
 
+	def get_deserialize_descriptor(self):
+		return self._get_deserialize_descriptor(False)
```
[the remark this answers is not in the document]
gimre-xymcity: weird, I tried it before and something was wrong 🤔

sdk/javascript/generator/PodTypeFormatter.py
```diff
@@ -43,6 +43,14 @@ def get_deserialize_descriptor(self):
 		body += f'return new {self.typename}({self.printer.load()});'
 		return MethodDescriptor(body=body)
 
+	def get_deserialize_aligned_descriptor(self):
+		if self._is_array:
+			return None
```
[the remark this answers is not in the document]
gimre-xymcity: a) this only means that deserializeAligned won't be generated for types like Address, Hash256 etc... mostly cause it doesn't make much sense and it would have to be exactly same as deserialize

what deserialize() for ByteArray does is:
 * it creates "shallow" uint8array around (offset, buffer, size)
 * it then passes that shallow reference to ctor, which then copies it 
 
technically we could _try_ to optimize that, although I don't think it's _worth_ it (moreover, I assume there's some sensible copying happening under the hood)

sdk/javascript/generator/printers.py
gimre-xymcity: hm?

gimre-xymcity: no, I think I wanted to phrase it differently, but now IDK how ^^

sdk/javascript/generator/printers.py
gimre-xymcity: I usually forget tuple can be used in statements like this

sdk/javascript/generator/TypeFormatter.py
gimre-xymcity: if you have `get_deserialize_aligned_descriptor` that returns None-like thing
i.e. for pod types:
```
	def get_deserialize_aligned_descriptor(self):
		if self._is_array:
			return None
```

sdk/javascript/test/utils/converter_spec.js
gimre-xymcity: erm, in prev describe theres:
assertBytesToIntOnMisaligned

in this describe there's:
assertBytesToIntUnalignedOnMisaligned 

(it refers to method name)

sdk/javascript/generator/printers.py
```diff
@@ -29,9 +29,14 @@ def get_default_value(self):
...
-	def load(self, buffer_name='byteArray'):
+	def load(self, buffer_name='byteArray', is_aligned=False):
 		data_size = self.get_size()
-		return f'converter.bytesToInt({buffer_name}, {data_size}, {js_bool(not self.descriptor.is_unsigned)})'
+		arguments = f'{buffer_name}, {data_size}, {js_bool(not self.descriptor.is_unsigned)}'
+		# is_aligned - handles both generation of deserializeAligned for pod and enum types as long as generation of fields within struct
+		if is_aligned:
+			return f'converter.bytesToInt({arguments})'
+
+		return f'converter.bytesToIntUnaligned({arguments})'
```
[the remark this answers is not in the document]
Jaguar0625: i meant something like 

`function_name = 'bytesToInt' if is_aligned else 'bytesToIntUnaligned' `

but not necessary

sdk/javascript/test/utils/converter_spec.js
Jaguar0625: i know, but it is scoped within `bytesToIntUnaligned`. don't mind, just seemed a little verbose to me given naming of assertBytesToInt.

Jaguar0625 (approved):

Jaguar0625 (approved):

gimre-xymcity:
merged into staging branch as:
 * 103f79c9208e000dc11139b14dee07690bf1fc88  (catbuffer/schemas change)
 * 5b450f5b158ca5c20fe89df382e3460786b97341 (actual change)
 * 14be333fa395d170d6fe8cfe113f30a1f461f235 (linters name length)

PR #283, opened 2022-06-29
[sdk/python]: initialize optional fields to None to match deserializer

Jaguar0625:
problem: deserializer sets missing optional field to None but constructor sets it to default value
solution: initial optional fields with None in constructor too

### DIFF
--- a/sdk/javascript/generator/StructTypeFormatter.py
+++ b/sdk/javascript/generator/StructTypeFormatter.py
@@ -115,7 +115,18 @@ class StructFormatter(AbstractTypeFormatter):
 			if const_field:
 				body += f'{field_name} = {self.typename}.{const_field.name};\n'
 			else:
-				body += f'{field_name} = {field.extensions.printer.get_default_value()};\n'
+				value = field.extensions.printer.get_default_value()
+				if field.is_conditional:
+					conditional = field.value
+					condition_field_name = conditional.linked_field_name
+					condition_field = next(f for f in self.non_const_fields() if condition_field_name == f.name)
+					condition_model = condition_field.extensions.type_model
+
+					# only initialize default implicit union field in constructor
+					if f'{condition_model.name}.{conditional.value}' != condition_field.extensions.printer.get_default_value():
+						value = 'null'  # needs to be null or else field will not be destination when copying descriptor properties
+
+				body += f'{field_name} = {value};\n'
 
 		body += '\n'.join(
 			map(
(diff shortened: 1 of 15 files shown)


sdk/javascript/generator/StructTypeFormatter.py
```diff
@@ -318,7 +333,13 @@ def generate_serialize_field(self, field):
...
 			serialize_line = f'buffer.write({serialize_field});{field_comment}\n'
 
-		return indent_if_conditional(condition, serialize_line)
+		if not condition:
+			return serialize_line
+
+		serialize_field_default = field.extensions.printer.store(field.extensions.printer.get_default_value())
+		return indent_if_conditional(
+			condition,
+			f'buffer.write({check_unset(f"this.{lang_field_name(field.name)}")} ? {serialize_field_default} : {serialize_field});\n')
```
gimre-xymcity, line 342: this part is bit weird, you have early quit above, so if we're inside `if condition` it means that most likely `condition` is already checking variable (or well condition) that's about to be serialized...
So I _think_ `check_unset` here is kinda redundant, so I _guess_ you should be able to use same `serialize_line` as above?

sdk/python/examples/descriptors/nem_namespace.py
```diff
@@ -5,7 +5,6 @@ def descriptor_factory():
 			'type': 'namespace_registration_transaction',
 			'rental_fee_sink': 'TAMESPACEWH4MKFMBCVFERDPOOP4FK7MTDJEYP35',
 			'rental_fee': 50000 * 1000000,
-			'parent_name': '',
```
gimre-xymcity, line 8: hmm, looking at it again, why was this change needed?

(this is nem, so serialization should result in a difference between UNSET parent_name and EMPTY parent_name)

sdk/javascript/generator/StructTypeFormatter.py
Jaguar0625: you're right, can simplify

sdk/python/examples/descriptors/nem_namespace.py
Jaguar0625: wasn't *needed* 

> this is nem, so serialization should result in a difference between UNSET parent_name and EMPTY parent_name

sdk does not support that difference, so serialization is identical for '' and undefined

sdk/javascript/src/nem/models.js
```diff
@@ -1677,7 +1677,7 @@ class MosaicDefinition {
 		size += arrayHelpers.size(this.properties);
 		size += 4;
 		if (0 !== this.levySize)
-			size += (undefined === this.levy || null === this.levy) ? 0 : this.levy.size;
+			size += this.levy.size;
```
gimre-xymcity, line 1680: this (levySize + levy) _could_ have been potentially one of the problematic places, so if it works, we're probably missing some tests, BUT

if this should be handled differently in the future it's clearly not in the scope of this change :)

gimre-xymcity (approved):

PR #284, opened 2022-07-01
[cabuffer/parser]: add support for sizeref field attribute

Jaguar0625:
problem: message_envelope_size should be automatically initialized with the size of another field
solution: add sizeref field attribute


catbuffer/schemas/nem/mosaic/mosaic_definition.cats
```diff
@@ -76,6 +76,7 @@ struct MosaicDefinition
 	properties = array(SizePrefixedMosaicProperty, properties_count)
 
 	# size of the serialized levy
+	@sizeref(levy, 4)
```
gimre-xymcity, line 79: hmm, actually (asking cause I don't remember), is this +4 or +0?
(what will generator do if value of this field - when deserializing - is 0?)

Jaguar0625: +4

gimre-xymcity (approved):

Jaguar0625:
included in #294

PR #285, opened 2022-07-01
[sdk/python]: symbol and nem shared key derivation

gimre-xymcity:

no shared key derivation at all 😎 

## What's the issue?

In order to be able to add encryption/decryption (and message type handling) to SDKs, we need shared key derivation first

This change only adds new code handling key derivation, there are 3 parts:
 1. old nem key derivation (I'm calling it deprecated throughout the change)
 2. symbol HKDF-based derivation
 3. nem HKDF-based derivation  - I ASSUMED, we do NOT want to use keccak inside HKDF, but rather use same HKDF as in symbol.

 * existing Symbol test vectors are passing
 * I recovered earlier NEM test vectors for "deprecated" variant
 * given that 2) was tested and resulted in correct vectors, I generated test vectors for 3)
 
 note: there's still something wrong with nem derivation, will investigate it further after the weekend
 
## UPDATE:

 1. unfortunately in case of NEM, keccak must be used to yield proper shared key
 2. initially NEM derivation was using (nem)KeyPair "sk" which contains _raw_ key
 3. after moving it was using private_key property and therefore yielding incorrect results (cause private_key is 'unreversed')
 4. I fixed derivation, found older test-vectors commit `fix nis1 test vectors after changes in v0.7.0.1 (private key is NOT reversed anymore)` 50384b78297a4c722e4dd55aab7285350fc1c98e, and verified results 👍 
 5. I've reversed step 4 and:
   a) reversed priv keys inside `tests/vectors/nem/crypto/3.test-derive-deprecated.json` file
   b) removed reversal inside `tests/vectors/all.py` in `DeriveDeprecatedTester.process`
   c) an instead added SANE reversal inside NEM's `SharedKey.derive_shared_key_deprecated`
6. it seems that "otherPublicKey" prior to commit mentioned in 3, were something strange, so in general I have:
 * updated deprecated file to use vectors from that commit (at least they have deterministic "otherPublicKey")
 * reversed private keys as mentioned in 4a
7. finally I have updated nem HKDF file in a following way:
  * reversed private keys
  * updated "otherPublicKeys" - the ones from Symbol HKDF file are obtained through Symbol's priv-> pub not NEM priv->pub **so in effect HKDF file and deprecated file should use SAME priv and "otherPublic" keys**
  * ofc had to generate `scalarMulResult` and `sharedKey` values


sdk/python/symbolchain/nem/SharedKey.py
```diff
@@ -0,0 +1,17 @@
+import sha3
+
+from ..external import ed25519
+from ..SharedKey import SharedKey as BasicSharedKey
+
+
+class SharedKey(BasicSharedKey):
+	@staticmethod
+	def derive_shared_key(key_pair, other_public_key):
+		return BasicSharedKey._derive_shared_key(key_pair, other_public_key, b'nem-nis1')
```
Jaguar0625, line 10: probably would go with just nem

sdk/python/tests/nem/test_SharedKey.py
```diff
@@ -0,0 +1,24 @@
...
+
+# DETERMINISTIC_SALT = b'1234567890ABCDEF1234567890ABCDEF'
+
+
+# class SharedKeyDeprecatedTest(BasicSharedKeyTest, unittest.TestCase):
+# 	def get_test_descriptor(self):
+# 		return SharedKeyTestDescriptor(
+# 			key_pair_class=KeyPair,
+# 			derive_shared_key=lambda private, public: SharedKey.derive_shared_key_deprecated(private, public, DETERMINISTIC_SALT)
+# 		)
```
Jaguar0625, line 16: fix

sdk/python/tests/test/BasicSharedKeyTest.py
```diff
@@ -0,0 +1,26 @@
...
+
+from symbolchain.CryptoTypes import PrivateKey
+
+SharedKeyTestDescriptor = namedtuple('SharedKeyTestDescriptor', ['key_pair_class', 'derive_shared_key'])
+
+
+class BasicSharedKeyTest:
+	# pylint: disable=no-member
+
+	# region mutual shared
```
Jaguar0625, line 11: please port tests from SharedKeyTests.cpp `// region shared secret / key` section

sdk/python/tests/vectors/all.py
```diff
@@ -18,6 +18,7 @@ def __init__(self, facade_class, network_class):
 		self.verifier_class = self.facade_class.Verifier
 		self.network_class = network_class
 		self.address_class = self.facade_class.Address
+		self.shared_key_interface = self.facade_class.SharedKey
```
Jaguar0625, line 21: why interface instead of class in the name like all the others ?

sdk/python/tests/vectors/all.py
```diff
@@ -115,6 +116,33 @@ def process(self, test_vector, _):
...
+		super().__init__(3, 'test-derive-deprecated', 'derive-deprecated')
+		self.class_locator = class_locator
+
+	def process(self, test_vector, _):
+		# Arrange:
+		other_public_key = PublicKey(test_vector['otherPublicKey'])
+		key_pair = self.class_locator.key_pair_class(PrivateKey(test_vector['privateKey']))
+		salt = unhexlify(test_vector['salt'])
+		key = self.class_locator.shared_key_interface.derive_shared_key_deprecated(key_pair, other_public_key, salt)
+		return [(unhexlify(test_vector['sharedKey']), key)]
```
Jaguar0625, line 130: missing Act + Assert (x2)

sdk/python/tests/vectors/all.py
```diff
@@ -115,6 +116,33 @@ def process(self, test_vector, _):
...
+	def __init__(self, class_locator):
+		super().__init__(3, 'test-derive-hkdf', 'derive')
+		self.class_locator = class_locator
+
+	def process(self, test_vector, _):
+		# Arrange:
+		other_public_key = PublicKey(test_vector['otherPublicKey'])
+		key_pair = self.class_locator.key_pair_class(PrivateKey(test_vector['privateKey']))
+		key = self.class_locator.shared_key_interface.derive_shared_key(key_pair, other_public_key)
+		return [(unhexlify(test_vector['sharedKey']), key.bytes)]
```
Jaguar0625, line 143: compare SharedKey type directly

sdk/python/tests/vectors/all.py
```diff
@@ -215,10 +243,13 @@ def load_test_suites(blockchain):
 		VerifyTester(class_locator),
 		Bip32DerivationTester(class_locator),
 		Bip39DerivationTester(class_locator),
+		DeriveTester(class_locator),
```
Jaguar0625, line 246: would move up to match source code order

sdk/python/symbolchain/external/ed25519.py
```diff
@@ -235,6 +235,16 @@ def signature_hash_unsafe(m, sk, pk, hashobj=hashlib.sha512):
 	return encodepoint(R) + encodeint(S)
 
 
+def derive_shared_secret_unsafe(pk, sk, hashobj=hashlib.sha512):
+	if not iscanonical(pk):
```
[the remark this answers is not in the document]
gimre-xymcity: commented in a same way as other functions in this file 😬

sdk/python/symbolchain/facade/SymbolFacade.py
```diff
@@ -32,6 +33,7 @@ class SymbolFacade:
 	Address = Address  # pylint: disable=duplicate-code
 	KeyPair = KeyPair  # pylint: disable=duplicate-code
 	Verifier = Verifier  # pylint: disable=duplicate-code
+	SharedKey = SharedKey
```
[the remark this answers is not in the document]
gimre-xymcity: hmm, so seems we actually only need (X-5), where X is number of common lines...
so in case above we actually only need on Address and a KeyPair, so not sure if it's better to delete the one on Verifier or add additional one on SharedKey...

sdk/python/symbolchain/SharedKey.py
```diff
@@ -0,0 +1,22 @@
...
+from .CryptoTypes import SharedKey256
+from .external import ed25519
+
+
+class SharedKey:
+	@staticmethod
+	def _derive_shared_key(key_pair, other_public_key, info):
+		"""Derives shared encryption key from private key, public key and salt."""
+		shared_secret = ed25519.derive_shared_secret_unsafe(other_public_key.bytes, key_pair.private_key.bytes, hashlib.sha512)
+		salt = bytes(32)
```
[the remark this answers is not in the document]
gimre-xymcity: this is key derivation, it is ok to use zero salt (as long as we're using random IV for encryption).

moreover, [rfc5869](https://datatracker.ietf.org/doc/html/rfc5869) says this (bold mine):
> _salt_     optional salt value (a **non-secret** random value); if not provided, it is set to a string of HashLen zeros.

AFAIK, the reason to use _salt_ is not to introduce randomness - for that purpose "info" part of HKDF should be used, reason to use _salt_ is to add some more entropy into non-uniform IKM hkdf input. Technically shared_secret is an encoded point, therefore not really uniform, but for all purposes it's entropy should be 253bits, so it probably shouldn't be big concern. 
 
There are some more notes on the salt and HKDF itself in https://eprint.iacr.org/2010/264.pdf

sdk/python/symbolchain/nem/SharedKey.py
gimre-xymcity: wanted to make it 8b in length same as 'catapult'

sdk/python/tests/nem/test_SharedKey.py
gimre-xymcity: 🎉

sdk/python/tests/vectors/all.py
gimre-xymcity: it feels weird to call it class, when it's _not_ a class per se 😬  (I mean you can't use it to make any instances)

sdk/python/tests/vectors/all.py
gimre-xymcity: order seems to be bit _weird_, MosaicIdDerivationTester is before Bip tests, althought it is appended at the end of a list 😬

sdk/python/symbolchain/facade/SymbolFacade.py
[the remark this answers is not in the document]
Jaguar0625: can delete if that works 👍

sdk/python/symbolchain/nem/KeyPair.py
```diff
@@ -1,7 +1,8 @@
+
```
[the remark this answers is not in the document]
Jaguar0625: missed this one

sdk/python/symbolchain/nem/SharedKey.py
```diff
@@ -0,0 +1,23 @@
...
+	@staticmethod
+	def derive_shared_key(key_pair, other_public_key):
+		"""Derives shared encryption key from private key and public key."""
+
+		# note: key_pair.private_key is "unreversed", so need to reverse here again
+		return BasicSharedKey._derive_shared_key(other_public_key.bytes, key_pair.private_key.bytes[::-1], b'nem-nis1', sha3.keccak_512)
+
+	@staticmethod
+	def derive_shared_key_deprecated(key_pair, other_public_key, salt):
+		"""Derives shared encryption key from private key, public key and salt."""
```
Jaguar0625, line 18: might want to document how this is different and/or at least mention it is deprecated

sdk/python/symbolchain/SharedKey.py
```diff
@@ -0,0 +1,16 @@
...
+from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.primitives.kdf.hkdf import HKDF
+
+from .CryptoTypes import SharedKey256
+from .external import ed25519
+
+
+class SharedKey:
+	@staticmethod
+	def _derive_shared_key(other_public_key_bytes, private_key_bytes, info, hashing_algorithm=hashlib.sha512):
```
Jaguar0625, line 12: looking again, is there reason you're passing in bytes vs PublicKey/PrivateKey ?

sdk/python/tests/vectors/all.py
```diff
@@ -115,6 +116,41 @@ def process(self, test_vector, _):
...
+		self.class_locator = class_locator
+
+	def process(self, test_vector, _):
+		# Arrange:
+		other_public_key = PublicKey(test_vector['otherPublicKey'])
+		key_pair = self.class_locator.key_pair_class(PrivateKey(test_vector['privateKey']))
+		salt = unhexlify(test_vector['salt'])
+
+		# Act:
+		key = self.class_locator.shared_key_interface.derive_shared_key_deprecated(key_pair, other_public_key, salt)
```
Jaguar0625, line 131: key => shared_key ? (x2)

sdk/python/tests/vectors/all.py
Jaguar0625: i see what you mean because it only contains statics.

not sure,  but what if you had some construction like 
`SharedKey(key_pair).derive(public_key)`

Jaguar0625: although, i'd still call it class due to usage of class keyword ^^

sdk/python/tests/nem/test_SharedKey.py
```diff
@@ -0,0 +1,26 @@
...
+
+DETERMINISTIC_SALT = b'1234567890ABCDEF1234567890ABCDEF'
+
+
+class SharedKeyDeprecatedTest(BasicSharedKeyTest, unittest.TestCase):
+	@staticmethod
+	def get_test_descriptor():
+		return SharedKeyTestDescriptor(
+			key_pair_class=KeyPair,
+			derive_shared_key=lambda private, public: SharedKey.derive_shared_key_deprecated(private, public, DETERMINISTIC_SALT)
```
Jaguar0625, line 16: `private` `public` naming 😳

sdk/python/symbolchain/nem/KeyPair.py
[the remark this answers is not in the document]
gimre-xymcity: thought I fixed it

sdk/python/symbolchain/SharedKey.py
gimre-xymcity: yes, the reason is I need to pass reversed priv key in case of nem 😑

sdk/python/tests/nem/test_SharedKey.py
gimre-xymcity: (changed earlier?)

sdk/python/symbolchain/CryptoTypes.py
```diff
@@ -43,6 +43,16 @@ def __init__(self, public_key):
 		super().__init__(self.SIZE, public_key.bytes if isinstance(public_key, PublicKey) else public_key, PublicKey)
 
 
+class SharedKey256(ByteArray):
```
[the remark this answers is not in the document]
Jaguar0625: pending

sdk/python/tests/test/BasicSharedKeyTest.py
```diff
@@ -0,0 +1,91 @@
...
+		test_descriptor = self.get_test_descriptor()
+
+		key_pair_1 = test_descriptor.key_pair_class(PrivateKey.random())
+		key_pair_2 = test_descriptor.key_pair_class(PrivateKey.random())
+		invalid_public_key = bytearray(key_pair_2.public_key.bytes)
+		invalid_public_key[31] = 1
+
+		# Act + Assert:
+		with self.assertRaises(ValueError):
+			test_descriptor.derive_shared_key(key_pair_1, PublicKey(bytes(invalid_public_key)))
```
Jaguar0625, line 89: invalid_public_key_bytes ?

sdk/python/symbolchain/CryptoTypes.py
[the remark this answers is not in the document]
gimre-xymcity: ? (ah)

sdk/python/tests/test/BasicSharedKeyTest.py
```diff
@@ -0,0 +1,91 @@
...
+		# Act:
+		shared_key_1 = test_descriptor.derive_shared_key(key_pair_1, other_public_key_1)
+		shared_key_2 = test_descriptor.derive_shared_key(key_pair_2, PublicKey(bytes(other_public_key_2)))
+
+		comparator(shared_key_1, shared_key_2)
+
+	def test_shared_keys_generated_with_same_inputs_are_equal(self):
+		self._assert_derived_shared_result(lambda _x, _y: None, self.assertEqual)
+
+	def test_shared_keys_generated_for_different_private_keys_are_different(self):
```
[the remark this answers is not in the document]
gimre-xymcity: technically it's private key that is getting modified

Jaguar0625 (approved):

PR #296, opened 2022-07-05
[sdk]: add helper function for generating random mnemonics

Jaguar0625:
problem: sdk cannot generate seed phrase
    solution: add Bip32.random


gimre-xymcity (approved):
yay, cool change, will be handy for testing

PR #299, opened 2022-07-05
[sdk]: add helper function for calculating metadata payload

Jaguar0625:
problem: user manually has to do XOR
    solution: add metadataUpdateValue to simplify


sdk/javascript/src/symbol/metadata.js
```diff
@@ -0,0 +1,23 @@
...
+	for (i = 0; i < Math.min(oldValue.length, newValue.length); ++i)
+		result[i] = oldValue[i] ^ newValue[i];
+
+	for (; i < newValue.length; ++i)
+		result[i] = newValue[i];
+
+	return result;
+};
+
+module.exports = { metadataUpdateValue };
```
gimre-xymcity, line 23: perfectly fine, although I was wondering if we need some additional wrapper that would accept string (and treat it as utf8 and just do TextEncoder.encode() + call the above)

gimre-xymcity (approved):

sdk/javascript/src/symbol/metadata.js
gimre-xymcity: (ask @cryptoBeliever for opinion)

Jaguar0625: technically our metadata is bytes, so i'd leave as is for consistency with py

PR #306, opened 2022-07-08
[sdk/javascript]: js key derivation

gimre-xymcity:
This is change similar to #285 for JS.

All derive test vectors pass, additional tests for SharedKey derivation has been added, that mimic the ones from py version

### DIFF
--- a/sdk/javascript/src/SharedKey.js
+++ b/sdk/javascript/src/SharedKey.js
@@ -1,0 +1,178 @@
+const { SharedKey256 } = require('./CryptoTypes');
+const { hkdf } = require('@noble/hashes/hkdf');
+const { sha256 } = require('@noble/hashes/sha256');
+const tweetnacl = require('tweetnacl');
+
+// order matches order of exported methods
+const {
+	crypto_verify_32, gf, pack25519, unpack25519, pow2523, set25519
+} = tweetnacl.lowlevel;
+
+// region curve operations - unfortunatelly tweetnacl.lowlevel does not expose those functions, so needed to copy them here
+
+const neq25519 = (a, b) => {
+	const c = new Uint8Array(32);
+	const d = new Uint8Array(32);
+	pack25519(c, a);
+	pack25519(d, b);
+	return crypto_verify_32(c, 0, d, 0);
+};
+
+const par25519 = a => {
+	const d = new Uint8Array(32);
+	pack25519(d, a);
+	return d[0] & 1;
+};
+
+const inv25519 = (o, i) => {
+	const { M, S } = tweetnacl.lowlevel;
+
+	const c = gf();
+	for (let a = 0; 16 > a; a++)
+		c[a] = i[a];
+
+	for (let a = 253; 0 <= a; a--) {
+		S(c, c);
+		if (2 !== a && 4 !== a)
+			M(c, c, i);
+	}
+	for (let a = 0; 16 > a; a++)
+		o[a] = c[a];
+};
+
+const pack = (r, p) => {
+	const { M } = tweetnacl.lowlevel;
+	const tx = gf();
+	const ty = gf();
+	const zi = gf();
+
+	inv25519(zi, p[2]);
+	M(tx, p[0], zi);
+	M(ty, p[1], zi);
+	pack25519(r, ty);
+
+	r[31] ^= par25519(tx) << 7;
+};
+
+const unpackNeg = (r, p) => {
+	const {
+		D, M, A, S, Z
+	} = tweetnacl.lowlevel;
+
+	const gf0 = gf();
+	const gf1 = gf([1]);
+	const I = gf([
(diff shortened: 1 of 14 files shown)


sdk/javascript/test/nem/SharedKey_spec.js
```diff
@@ -0,0 +1,19 @@
+const { NemFacade } = require('../../src/facade/NemFacade');
+const { KeyPair } = require('../../src/nem/KeyPair');
+const { deriveSharedKeyDeprecated } = require('../../src/nem/SharedKey');
+const { runBasicSharedKeyTests } = require('../test/sharedKeyTests');
+
+describe('SharedKey (NEM)', () => {
+	runBasicSharedKeyTests({
+		KeyPair,
+		deriveSharedKey: NemFacade.deriveSharedKey
```
gimre-xymcity, line 9: deliberately calling through facade here and in symbol version below

sdk/javascript/test/test/sharedKeyTests.js
```diff
@@ -0,0 +1,79 @@
...
+
+		// Assert:
+		expect(sharedKey1).to.deep.equal(sharedKey2);
+	});
+
+	// endregion
+
+	// region invalid public key
+
+	it('public key not on the curve throws', () => {
```
gimre-xymcity, line 65: I think this test my occasionally fail : read point might end up on curve, and it _seems_ it might be failing more often then py variant, which in turn suggest that server-based isCanonical check might be bit too permissive...
will need to investigate

sdk/javascript/src/SharedKey.js
```diff
@@ -0,0 +1,178 @@
...
+const { hkdf } = require('@noble/hashes/hkdf');
+const { sha256 } = require('@noble/hashes/sha256');
+const tweetnacl = require('tweetnacl');
+
+// order matches order of exported methods
+const {
+	crypto_verify_32, gf, pack25519, unpack25519, pow2523, set25519
+} = tweetnacl.lowlevel;
+
+// region curve operations - unfortunatelly tweetnacl.lowlevel does not expose those functions, so needed to copy them here
```
Jaguar0625, line 11: ok to copy here, but why not export from tweetnacl version used by NEM ?

sdk/javascript/src/SharedKey.js
```diff
@@ -0,0 +1,178 @@
...
+	// check if result is neutral element
+	const gf0 = gf();
+	const areEqual = neq25519(result[1], result[2]);
+	const isZero = neq25519(gf0, result[0]);
+
+	// yes, this is supposed to be  bit OR
+	return 0 === (areEqual | isZero);
+};
+
+const deriveSharedSecretFactory = crypto_hash => (privateKeyBytes, otherPublicKey) => {
```
Jaguar0625, line 144: cryptoHash

sdk/javascript/src/SharedKey.js
```diff
@@ -0,0 +1,178 @@
...
+	const { scalarmult, Z } = tweetnacl.lowlevel;
+	const point = [gf(), gf(), gf(), gf()];
+
+	if (!isCanonicalKey(otherPublicKey) || 0 !== unpackNeg(point, otherPublicKey.bytes) || !isInMainSubgroup(point))
+		throw Error('invalid point');
+
+	// negate point == negate X coordinate and 't'
+	Z(point[0], gf(), point[0]);
+	Z(point[3], gf(), point[3]);
```
Jaguar0625, line 153: why do you need the negation before scalarmult?

Jaguar0625: ah, unpackneg 🆗

Jaguar0625 (approved):

sdk/javascript/src/SharedKey.js
Jaguar0625: i mean update that one to export more in lowlevel

gimre-xymcity: as mentioned on discord, imo it would look _weird_ if symbol version would use >two< tweetnacl imports.

sdk/javascript/src/nem/SharedKey.js
```diff
@@ -0,0 +1,46 @@
...
+
+	const sharedSecret = deriveSharedSecretImpl(reversedPrivateKeyBytes, otherPublicKey);
+	const sharedKeyBytes = new Uint8Array(SharedKey256.SIZE);
+	for (let i = 0; SharedKey256.SIZE > i; ++i)
+		sharedKeyBytes[i] = sharedSecret[i] ^ salt[i];
+
+	return new SharedKey256(keccak_256(sharedKeyBytes));
+};
+
+module.exports = { deriveSharedKey, deriveSharedKeyDeprecated };
```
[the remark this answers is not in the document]
gimre-xymcity: it _is_ same (?)

sdk/javascript/test/test/sharedKeyTests.js
gimre-xymcity: explained on slack, the issue was kind of PEBKAC, try catch around expect.to.throw was a bad idea, similar issue was in py, had to replace both with some manual checks

PR #309, opened 2022-07-12
[client\catatpult] fix: Windows tests failing

Wayonb:
Catapult client building on Windows is failing

## What's the issue?
There are couple of errors
1. warning RC4005: 'WIN_TARGET_NAME' : redefinition
2. fatal error LNK1169: one or more multiply defined symbols found

1. remove the WIN_TARGET_NAME to fix the ``warning RC4005: 'WIN_TARGET_NAME' : redefinition``
2. update cmake files to use no deps for lock hash and lock secret tests to fix ``fatal error LNK1169: one or more multiply defined symbols found``
3. update the ConvertToExceptionFunctionName() to fix the exception tests with msvc

Tested both on Windows and locally with gcc/clang

### DIFF
--- a/client/catapult/tests/catapult/utils/CatapultExceptionTests.cpp
+++ b/client/catapult/tests/catapult/utils/CatapultExceptionTests.cpp
@@ -79,12 +79,12 @@ namespace catapult {
 
 		// Boost function name got updated and now return in the form below.  This function coverts the function name to the
 		// expected value in the exception.
-		// Clang does not include the traits information in the function name.
+		// Only gcc includes the traits information in the exception function name.
 		// void catapult::CatapultExceptionTests_CanCopyConstructException() [with TTraits = catapult::{anonymous}::RuntimeErrorTraits]
 		std::string ConvertToExceptionFunctionName(const std::string& functionFullName) {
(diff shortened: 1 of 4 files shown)


client/catapult/tests/catapult/utils/CatapultExceptionTests.cpp
```diff
@@ -79,20 +79,20 @@ namespace catapult {
...
+			std::size_t functionNameEnd = functionFullName.find_first_of("<(");
 			if (std::string::npos == functionNameEnd)
 				return functionFullName;
 
 			auto functionNameStart = functionFullName.rfind("::", functionNameEnd);
 			functionNameStart = (std::string::npos != functionNameStart) ? functionNameStart + 2 : 0;
 			std::ostringstream oss(functionFullName.substr(functionNameStart, functionNameEnd - functionNameStart), std::ios_base::ate);
 
-#ifndef __clang__
+#if defined(__GNUC__) && !defined(__clang__)
```
Jaguar0625, line 95: is this failing on linux clang builds too ?

Jaguar0625 (approved):

PR #341, opened 2022-07-26
[sdk/javascript]: add VotingKeysGenerator

Jaguar0625:
problem: JS sdk doesn't have voting keys generator but PY sdk does
solution: add voting keys generator to JS sdk to achieve parity


sdk/javascript/test/symbol/resources/voting_keys_generator_test_vectors.yaml
```diff
@@ -0,0 +1,20 @@
+- name: test_vector_1
```
Jaguar0625, line 1: right now this file is duplicated between JS + PY sdks because it is used by unit tests. not sure what to do with it? we could move to testvectors and symlink perhaps?

@gimre-xymcity ?

sdk/javascript/src/symbol/VotingKeysGenerator.js
```diff
@@ -0,0 +1,65 @@
...
+
+		const view = new DataView(buffer);
+		view.setBigUint64(0, BigInt(startEpoch), true);
+		view.setBigUint64(8, BigInt(endEpoch), true);
+		view.setBigUint64(16, 0xFFFFFFFFFFFFFFFFn, true);
+		view.setBigUint64(24, 0xFFFFFFFFFFFFFFFFn, true);
+
+		setBuffer(view, 32, this.rootKeyPair.publicKey.bytes);
+		view.setBigUint64(64, BigInt(startEpoch), true);
+		view.setBigUint64(72, BigInt(endEpoch), true);
```
gimre-xymcity, line 43: hmm, maybe we should ad comment here, what this fields are
(or some simple in-code explanation/ascii-drawing of how the format looks like)

sdk/javascript/test/symbol/resources/voting_keys_generator_test_vectors.yaml
gimre-xymcity: hmm, looking at it, I think we should just move that part into testvectors, and run it as a part of all.js/all.py
(read: it doesn't really feel like a unit test, does it?)

not :100: sure, what do you think?

gimre-xymcity (approved):

PR #343, opened 2022-07-27
[monorepo]: move voting key generation tests to test_vectors

Jaguar0625:
problem: there are multiple copies of voting key generation test vectors
          and they are being used during unit tests
solution: promote vector file to tests/vectors
          process it as part of test_vectors stage


gimre-xymcity (approved):
👍

PR #354, opened 2022-08-03
[jenkins] fix: compiler and OS update for catapult client

Wayonb:
## What's the issue?
Catapult builds are using older version of compilers.

Latest gcc and clang compiler are used for catapult client builds.

Tested with Jenkins.

There are still some failures I am looking into with gcc 12 and sans tests


jenkins/catapult/compilers/ubuntu-clang/Dockerfile
```diff
@@ -0,0 +1,29 @@
+# Use the latest Ubuntu 22.04 image.
```
Jaguar0625, line 1: good job parameterizing 👍

jenkins/catapult/configurations/gcc-code-coverage.yaml
```diff
@@ -1,3 +1,3 @@
-compiler: ../compilers/gcc-10.yaml
+compiler: ../compilers/gcc-11.yaml
```
Jaguar0625, line 1: 1. should this be gcc latest (12) ?
2. should we have a clang-latest.yaml gcc-latest.yaml to minimize future changes when bumping versions ?

jenkins/catapult/configurations/gcc-westmere.yaml
```diff
@@ -0,0 +1,2 @@
+compiler: ../compilers/gcc-11.yaml
```
Jaguar0625, line 1: should this be gcc latest (12)?

jenkins/catapult/jenkins/catapult-client-build-compiler-image.groovy
```diff
@@ -0,0 +1,67 @@
...
+						echo """
+									env.GIT_BRANCH: ${env.GIT_BRANCH}
+								 MANUAL_GIT_BRANCH: ${MANUAL_GIT_BRANCH}
+
+							COMPILER_VERSION: ${COMPILER_VERSION}
+								  OPERATING_SYSTEM: ${OPERATING_SYSTEM}
+
+								   destImageName: ${destImageName}
+						"""
+					}
```
Jaguar0625, line 47: colon align printed lines?

jenkins/catapult/templates/UbuntuTestSanitizerBaseImage.Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM symbolplatform/symbol-server-compiler:ubuntu-clang-14
+ARG DEBIAN_FRONTEND=noninteractive
```
Jaguar0625, line 2: 1. ARG or ENV ?
2. can't you use UbuntuTestBaseImage.dockerfile?

jenkins/catapult/baseImageDockerfileGenerator.py
```diff
@@ -131,6 +131,9 @@ def rocks(self):
 		descriptor.options += DEPENDENCY_FLAGS['facebook_rocksdb']
 		descriptor.options += ['-DUSE_RTTI=1']
 
+		if self.compiler.c.startswith('gcc') and self.compiler.version == 12:
+			descriptor.options += ['-DFAIL_ON_WARNINGS=OFF']
```
Jaguar0625, line 135: 1. add comment about why we're (temporarily) disabling
2. can you just disable specific warnings (instead of all)

jenkins/catapult/baseImageDockerfileGenerator.py
```diff
@@ -355,8 +358,26 @@ def add_git_dependency(organization, project, versions_map, options, revision=1)
...
+		'make install',
+		'cd ..',
+		'rm -rf openssl'
+	], OPEN_SSL_OPTIONS=' '.join(options.openssl()), OPEN_SSL_CONFIGURE=' '.join(configure), VERSION=version, COMPILER=compiler)
+
+
 def generate_phase_deps(options):
 	print(f'FROM {options.layer_image_name("boost")}')
+
+	add_openssl(options, [])
```
Jaguar0625, line 379: why are you always installing openssl now ?

jenkins/catapult/runDockerBuildInnerPrepare.py
```diff
@@ -31,12 +31,6 @@ def main():
 	for name in bin_folder_names:
 		environment_manager.copy_tree_with_symlinks(DATA_VOLUME / 'binaries' / name, USER_HOME / name)
 
-	# LD_LIBRARY_PATH is not passed when llvm-symbolizer is started via asan, so move libs to system location
-	if is_dev_build:
-		system_bin_path = environment_manager.system_bin_path
-		environment_manager.move_glob_with_symlinks(USER_HOME / 'deps', 'libLLVM*', system_bin_path)
-		process_manager.dispatch_subprocess(['ls', '-laF', system_bin_path])
-
```
Jaguar0625, line 39: how was this fixed? i.e. why can it be removed?

jenkins/catapult/configurations/gcc-code-coverage.yaml
Wayonb: 1. Yes, can move it to 12.  All the daily tests should run with the latest compiler?  Weekly tests are the current - 1 compiler version?
2. Good idea, I think that would help

jenkins/catapult/templates/UbuntuTestSanitizerBaseImage.Dockerfile
Wayonb: 1. ARG is better to use since the variable is only required during the docker build.  ENV will be set when the container is running.
2. Yea, its same except I remove some tools.

jenkins/catapult/baseImageDockerfileGenerator.py
Wayonb: 1. ok
2. not sure it thats possible from cmake settings.

jenkins/catapult/baseImageDockerfileGenerator.py
Wayonb: Ubuntu got upgraded from 20.04 -> 22.04.  22.04 ships with openssl 3.x and catapult is still using openssl 1.1.1 so need to install in the base image

jenkins/catapult/runDockerBuildInnerPrepare.py
Wayonb: Sanitizer test base image has clang the compiler already installed.  No need to copy compiler files.
https://github.com/symbol/symbol/pull/354/files/3f5d57935b8bb5978627131d4f6b9022404a6f60#diff-579119f97023a7b6e06ea31aaac2573264eac7fd0b2cfe21051fdd638fbbe14a

jenkins/catapult/templates/UbuntuTestSanitizerBaseImage.Dockerfile
Jaguar0625: `ARG is better` you should check other usages of ARG vs ENV in dockerfiles as a post check-in task

jenkins/catapult/configurations/gcc-code-coverage.yaml
Jaguar0625: >  the daily tests should run with the latest compiler? Weekly tests are the current - 1 compiler version

i think that is fine, @gimre-xymcity

client/catapult/tests/catapult/utils/CatapultExceptionTests.cpp
```diff
@@ -97,7 +98,15 @@ namespace catapult {
 			if (std::string::npos != foundTraits) {
 				auto traitsEnd = functionFullName.find("]", foundTraits);
 				auto functionTraits = functionFullName.substr(foundTraits + 2, traitsEnd - foundTraits - 2);
-				oss << "<" << functionTraits << ">";
+				oss << "<";
+
+	// A bug in gcc 12 where the trait information is missing the namespace.
+	// Gcc 12 returns this - [with TTraits = {anonymous}::RuntimeErrorTraits]
+	#if (__GNUC__ == 12)
```
Jaguar0625: if this is GCC 12 bug, shouldn't it be `#if 12 != __GNUC_`

jenkins/catapult/jenkins/catapult-client-build-base-image-all.groovy
```diff
@@ -43,32 +43,32 @@ pipeline {
...
 						}
 					}
 				}
 
-				stage('clang 13') {
+				stage('clang prior') {
 					steps {
 						script {
-							dispatch_build_base_image_job('clang-13', 'ubuntu', false)
+							dispatch_build_base_image_job('clang-prior', 'ubuntu', true)
```
Jaguar0625, line 64: did you intentionally change false to true?

jenkins/catapult/jenkins/catapult-client-build-catapult-project.groovy
```diff
@@ -6,7 +6,8 @@ pipeline {
 	parameters {
 		gitParameter branchFilter: 'origin/(.*)', defaultValue: "${env.GIT_BRANCH}", name: 'MANUAL_GIT_BRANCH', type: 'PT_BRANCH'
 		choice name: 'COMPILER_CONFIGURATION',
-			choices: ['gcc-12', 'gcc-11', 'gcc-westmere', 'clang-13', 'clang-14', 'clang-ausan', 'clang-tsan', 'gcc-code-coverage'],
+			choices: ['gcc-latest', 'gcc-prior', 'gcc-westmere', 'clang-latest', 'clang-prior', 'clang-ausan', 'clang-tsan',
+					  'gcc-code-coverage'],
```
Jaguar0625, line 10: are leading spaces intentional ?

jenkins/catapult/environment.py
```diff
@@ -28,6 +28,18 @@ def system_bin_path(self):
 
 		raise RuntimeError('unable to detect system bin path')
 
+	@property
+	def local_lib_path(self):
+		if self.dry_run:
+			return '<LOCAL_LIB_PATH>'
+
+		for descriptor in [('fedora', '/usr/local/lib64'), ('ubuntu', '/usr/local/lib')]:
```
Jaguar0625, line 36: i guess we don't call this for debian ?

jenkins/catapult/environment.py
```diff
@@ -28,6 +28,18 @@ def system_bin_path(self):
 
 		raise RuntimeError('unable to detect system bin path')
 
+	@property
+	def local_lib_path(self):
```
Jaguar0625, line 32: why did we need to add this ?

jenkins/docker/cpp.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM symbolplatform/symbol-server-build-base:ubuntu-gcc-11-conan
+FROM symbolplatform/symbol-server-build-base:ubuntu-gcc-12-conan
```
Jaguar0625, line 1: should we rename build-base to drop numbered versions too ? 
(can be separate PR ofc)

client/catapult/tests/catapult/utils/CatapultExceptionTests.cpp
Wayonb: The issue is that the namespace information is in the exception message but missing from the boost function.
if GCC 12 then add the missing namespace to the trait so it will match the exception info

jenkins/catapult/jenkins/catapult-client-build-base-image-all.groovy
Wayonb: yea, I added a conan build for previous compiler version.

jenkins/catapult/jenkins/catapult-client-build-catapult-project.groovy
Wayonb: no, editor auto warps :(

jenkins/catapult/environment.py
Wayonb: yea, it does get call for Debian, but it will match ubuntu since it checks for path existing.  Not sure if there is a better way.

jenkins/catapult/environment.py
Wayonb: Due to openssl getting copied to all the test base images.  Fedora also now ships with openssl 3.
Local lib path is different for Fedora vs Ubuntu.

jenkins/docker/cpp.Dockerfile
Wayonb: Yea, I was thinking dropping the numbered version also.  The main issue would be that if the latest image is broken during an upgrade then CI will also be broken

client/catapult/tests/catapult/utils/CatapultExceptionTests.cpp
Jaguar0625: so, you mean it is a bug in boost?

jenkins/catapult/environment.py
Jaguar0625: ok, that's probably fine

jenkins/docker/cpp.Dockerfile
Jaguar0625: how is that different than if base image is broken now?

client/catapult/tests/catapult/utils/CatapultExceptionTests.cpp
Wayonb: maybe. idk.  I thought compiler since it works in gcc11.

jenkins/docker/cpp.Dockerfile
Wayonb: Currently CI will use a specific base image plus conan packages.  If new compiler(new base image) or packages is been tested on a branch it will not affect the other branches.
If the versioned number is removed then when compiler gets updated, all jobs will be using the same base image.
Its only compiler updates so not a big deal.   I can remove update it in another pr.

Jaguar0625 (approved):
👍

PR #366, opened 2022-08-26
[sdk/javascript]: unify sign/verify behavior with python

Jaguar0625:
problem: javascript does not require canonical signature
solution: add canonical signature check
          port additional edge case tests from python


sdk/javascript/src/symbol/KeyPair.js
```diff
@@ -40,6 +41,30 @@ class KeyPair {
 	}
 }
 
+const isCanonicalS = encodedS => {
```
Jaguar0625, line 44: alternatively could promote this and use for both - preference?

gimre-xymcity (approved):

PR #367, opened 2022-08-26
[sdk/python] use native ripemd160 package

gimre-xymcity:
## What's the issue?

Openssl developers figured out it will be great idea to drop ripemd160 hash function in openssl 3.x, that combined with fact that py3.x in ubu is linked against openssl 3.x is even better

This change introduces ripemd native package that we've made by extracting from pycryptodome.

There's a small wrapper that picks either hashlib impl or newly added package

All tests and test vectors are passing.


sdk/python/symbolchain/ripemd160.py
```diff
@@ -0,0 +1,14 @@
+import hashlib
+
+if 'ripemd160' not in hashlib.algorithms_available:
+	from ripemd import ripemd160 as ripemd160_impl
```
Jaguar0625, line 4: does this pass lint?

sdk/python/symbolchain/ripemd160.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+def _factory():
+	return hashlib.new('ripemd160') if 'ripemd160' in hashlib.algorithms_available else ripemd160_impl.new()
+
+
+def ripemd160(data):
```
Jaguar0625, line 11: document

Jaguar0625 (approved):

PR #377, opened 2022-09-05
speed up JS SDK sign+verify operations by writing Rust version compiled to WASM

Jaguar0625:


jenkins/docker/javascript.Dockerfile
```diff
@@ -39,4 +39,11 @@ RUN useradd --uid 1000 -ms /bin/bash ubuntu
 RUN mkdir -p /data/db \
 	&& chown -R ubuntu:ubuntu /data
 
+# install rust and wasm-pack
+ENV PATH=$PATH:/home/ubuntu/.cargo/bin
+RUN curl https://sh.rustup.rs -sSf | CARGO_HOME=/home/ubuntu/.cargo sh -s -- -y \
+	&& . "/home/ubuntu/.cargo/env" \
```
Jaguar0625, line 45: not sure if there's a way to avoid source call
or if nicer way to install this as user (ubuntu) instead of root

sdk/javascript/wasm/src/lib.rs
```diff
@@ -0,0 +1,272 @@
...
+	builder_h.update(&public_key);
+	builder_h.update(&message);
+	let h = hash_to_scalar(builder_h);
+
+	// S = (r + h * a) mod group order
+	let a = copy_and_clamp_private_hash(&private_hash);
+	#[allow(non_snake_case)]
+	let S = &(&h * &a) + &r;
+
+	// serialize and concat R S
```
gimre-xymcity, line 168: we decided not to do additional S check, right?

sdk/javascript/src/impl/ed25519.js
```diff
@@ -0,0 +1,27 @@
...
+const CRYPTO_SIGN_BYTES = 64;
+const CRYPTO_SIGN_PUBLICKEYBYTES = 32;
+const CRYPTO_SIGN_SECRETKEYBYTES = 64;
+
+const ed25519 = {
+	keyPairFromSeed: (hashMode, seed) => {
+		const publicKey = new Uint8Array(CRYPTO_SIGN_PUBLICKEYBYTES);
+		const privateKey = new Uint8Array(CRYPTO_SIGN_SECRETKEYBYTES);
+		for (let i = 0; i < CRYPTO_SIGN_SECRETKEYBYTES; i++)
+			privateKey[i] = seed[i];
```
gimre-xymcity, line 14: I'm not sure if we need privateKey here, I think that:
 * we already require `seed` to be uint8Array
 * I guess we can throw seed into returned dict below (or is there some reason _not_ to?)
  * crypto_sign_keypair does copying to internal sk anyway

sdk/javascript/wasm/src/lib.rs
```diff
@@ -0,0 +1,272 @@
...
+
+/// Generates public key given secret key.
+pub fn crypto_sign_keypair_unboxed(hash_mode: HashMode, sk: &[u8; SECRET_KEY_LENGTH]) -> [u8; PUBLIC_KEY_LENGTH] {
+	set_panic_hook();
+
+	let private_hash = hash_private_key(hash_mode, sk);
+	let key = copy_and_clamp_private_hash(&private_hash);
+	let point = &key * &curve25519_dalek::constants::ED25519_BASEPOINT_TABLE;
+	let compressed = point.compress();
+	compressed.to_bytes()
```
gimre-xymcity, line 140: I'm wondering if we should zero key and/or private_hash after calculating point here... (and similar thing in signing below, zero after calculating S)

sdk/javascript/wasm/src/utils.rs
```diff
@@ -0,0 +1,10 @@
+pub fn set_panic_hook() {
```
gimre-xymcity, line 1: I'm somewhat wondering if we should call it when inside wasm at all...

sdk/javascript/wasm/src/lib.rs
Jaguar0625: which additional S check?

gimre-xymcity (approved):

PR #419, opened 2022-10-17
[catbuffer/schemas]: include versions explicitly in all multiversioned transactions

Jaguar0625:
[problem]: latest version has no postfix (e.g. 'transfer_transaction')
            versions start at 1
[solution]: require latest version to use version postfix (e.g. 'transfer_transaction_v1')
            renumber versions starting at 0


0x6861746366574 (approved):

Jaguar0625:
renumbering starting with version 1

PR #437, opened 2022-11-03
[sdk] fix: updated generation hash post network reset

yilmazbahadir:
Updating testnet generation hash seed in the SDKs(JS and Python) after testnet network reset.


Jaguar0625 (approved):

Jaguar0625 (changes requested):
Just noticed this is causing quite a few tests to fail. Please fix.

yilmazbahadir:
> Just noticed this is causing quite a few tests to fail. Please fix.

Actually raised the PR for us not to forget about it at the time, I'm on it now 👍

Jaguar0625 (approved):

PR #438, opened 2022-11-06
[catapult/client] build: fail on lint and individually disable gcc 12 warnings

Jaguar0625:
problem: lint failures don't fail build
solution: fail build on lint failure

problem: gcc 12 stringop-overread and array-bounds warnings are disabled
solution: enable those warnings and fix / disable individually

### DIFF
--- a/client/catapult/plugins/txes/restriction_mosaic/tests/state/RestrictionValueMapTests.cpp
+++ b/client/catapult/plugins/txes/restriction_mosaic/tests/state/RestrictionValueMapTests.cpp
@@ -142,14 +142,17 @@ namespace catapult { namespace state {
 	TEST(TEST_CLASS, CanRemoveValueNotInMap) {
 		// Arrange:
 		RestrictionValueMap<uint64_t> map;
+		map.set(111, 444);
+		map.set(321, 987);
 
 		// Act:
-		map.remove(111);
+		map.remove(222);
 
 		// Assert:
-		EXPECT_EQ(0u, map.size());
-		AssertCannotGetValue(map, 111);
-		EXPECT_EQ(std::set<uint64_t>(), map.keys());
+		EXPECT_EQ(2u, map.size());
+		AssertCanGetValue(map, 111, 444);
+		AssertCanGetValue(map, 321, 987);
+		EXPECT_EQ(std::set<uint64_t>({ 111, 321 }), map.keys());
 	}
 
 	TEST(TEST_CLASS, CanRemoveSingleValue) {
--- a/client/catapult/src/catapult/model/EntityRange.h
+++ b/client/catapult/src/catapult/model/EntityRange.h
@@ -392,6 +392,9 @@ namespace catapult { namespace model {
 
 		/// Creates an entity range around \a numElements fixed size elements pointed to by \a pData.
 		static Range CopyFixed(const uint8_t* pData, size_t numElements) {
+			if (0 == numElements)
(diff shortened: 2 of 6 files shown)


client/catapult/plugins/txes/restriction_mosaic/tests/state/RestrictionValueMapTests.cpp
```diff
@@ -142,14 +142,17 @@ namespace catapult { namespace state {
 	TEST(TEST_CLASS, CanRemoveValueNotInMap) {
 		// Arrange:
 		RestrictionValueMap<uint64_t> map;
+		map.set(111, 444);
+		map.set(321, 987);
 
 		// Act:
-		map.remove(111);
+		map.remove(222);
```
Jaguar0625, line 149: workaround array-bounds. it does not follow if in remove and warns calling erase on empty vector.

client/catapult/src/catapult/model/EntityRange.h
```diff
@@ -392,6 +392,9 @@ namespace catapult { namespace model {
 
 		/// Creates an entity range around \a numElements fixed size elements pointed to by \a pData.
 		static Range CopyFixed(const uint8_t* pData, size_t numElements) {
+			if (0 == numElements)
+				return Range();
```
Jaguar0625, line 396: stringop-overread does not follow if in memcpy_cond so assumes copy is called when count is 0.

client/catapult/tests/catapult/model/TransactionTests.cpp
```diff
@@ -118,8 +118,18 @@ namespace catapult { namespace model {
 		// Arrange:
 		std::vector<uint8_t> buffer(sizeof(SizePrefixedEntity));
 		auto* pTransaction = reinterpret_cast<Transaction*>(&buffer[0]);
+
+#if defined(__GNUC__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Warray-bounds" // Transaction[0] is partly outside array bounds
+#endif
+
 		pTransaction->Size = sizeof(SizePrefixedEntity);
```
Jaguar0625, line 127: warning is because buffer.size() is smaller than sizeof(Transaction). alternatively, could make buffer larger, but that somewhat weakens the test.

jenkins/catapult/runDockerTestsInnerLint.py
```diff
@@ -110,8 +112,8 @@ def main():
 	environment_manager.chdir(script_path)
 
 	linter_runner = LinterRunner(process_manager, args.out_dir, args.dry_run)
-	run_shell_linters(linter_runner, find_files_with_extension(environment_manager, '.sh'))
-	run_python_linters(linter_runner, find_files_with_extension(environment_manager, '.py'), str(script_path))
+	run_shell_linters(linter_runner, find_files_with_extension(environment_manager, source_dir, '.sh'))
+	run_python_linters(linter_runner, find_files_with_extension(environment_manager, source_dir, '.py'), str(script_path))
```
Jaguar0625, line 116: these were originally scanning files in /scripts (catapult-src/jenkins/catapult) which is wrong directory and indeed has no sh files

Wayonb: Should these linters run in the PR flow?

Wayonb (approved):

gimre-xymcity (approved):

PR #477, opened 2022-12-06
[jenkins] fix: handle testcase decode output failure 

Wayonb:
## What's the issue?
Occasionally catapult tests on Windows fails due to invalid bytes get included in the tests output.

```
15:52:24  catapult-client-build-catapult-project-test-1  | Failed to decode line: b'[----------] 1 test from FutureTests\xfb\xc0\r\n'

15:52:24  catapult-client-build-catapult-project-test-1  |  Exception: 'utf-8' codec can't decode byte 0xfb in position 36: invalid start byte
```

Instead of failing the tests run when decode of the output fails, it will log the failure and continue running the tests.

Ran in Jenkins

### DIFF
--- a/jenkins/catapult/process.py
+++ b/jenkins/catapult/process.py
@@ -9,6 +9,15 @@ class ProcessManager:
 	def __init__(self, dry_run=False):
 		self.dry_run = dry_run
 
+	@staticmethod
+	def _decode_line(line_bin):
+		try:
+			return line_bin.decode('utf-8')
+		except UnicodeDecodeError as ex:
+			err = f'Failed to decode line: {line_bin}\n Exception: {str(ex)}\n'
+			print(err)
+			return err
+
 	def dispatch_subprocess(self, command_line, show_output=True, handle_error=True, redirect_filename=None):
 		self._print_command(command_line)
 
@@ -18,7 +27,7 @@ class ProcessManager:
 		with Popen(command_line, stdout=PIPE, stderr=STDOUT) as process:
 			process_lines = []
 			for line_bin in iter(process.stdout.readline, b''):
-				line = line_bin.decode('utf-8')
+				line = self._decode_line(line_bin)
 
 				if show_output:
 					sys.stdout.write(line)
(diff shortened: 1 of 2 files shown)


jenkins/catapult/process.py
```diff
@@ -9,6 +9,13 @@ class ProcessManager:
 	def __init__(self, dry_run=False):
 		self.dry_run = dry_run
 
+	@staticmethod
+	def _decode_line(line_bin):
+		try:
+			return line_bin.decode('utf-8')
+		except UnicodeDecodeError as ex:
+			return f'Failed to decode line: {line_bin}\n Exception: {str(ex)}\n'
```
Jaguar0625, line 17: is it possible to just strip out / replace binary characters?

gimre-xymcity: imo ignoring the exception is kinda fine, as it should happen only on windows... (unless we spit out some crap, somewhere), however...

@Wayonb hm, the actual reason for that error is that this obviously is _not_ UTF8 sequence, 0xFB is invalid prefix, I'm wondering if this is something GTEST spits out (as this line is likely colored, although I don't see how this should be related...) or is it something else?
You might want to try adding `--gtest_color=no` on windows though, maybe it will help?

The other possible issue (although not sure if this will help or make things worse): is console codepage set to utf8? (there's `chcp` thing: https://superuser.com/a/269857)

Wayonb: @gimre-xymcity You are right in that its something to do with GTEST but it seems random, meaning it wasnt always the same test that cause this error.

I will try turning off the color for gtest on Windows.  Thanks.

Wayonb: I think disabling the color on gtest might have fix the issue.  Its hard to tell since it does happened all the time.

I also left logic in there to just drop characters which it fails to decode.  There is a perf hit with this so maybe just do it for Windows?

Wayonb: Still repo with gtest color disabled

Jaguar0625 (approved):

Jaguar0625 (approved):

PR #484, opened 2022-12-09
[client/rest]: add inflation route(s)

Jaguar0625:
problem: there is no way to determine inflation (block rewards) from rest
solution: add API returning contents of config-inflation
          add API returning inflation given block height

### DIFF
--- a/client/rest/src/routes/networkRoutes.js
+++ b/client/rest/src/routes/networkRoutes.js
@@ -47,6 +47,26 @@ module.exports = {
 			contents => ini.parse(contents)
 		);
 
+		const readAndParseInflationPropertiesFile = () => fileLoader.readOnce(
+			services.config.apiNode.inflationPropertyFilePath,
+			contents => {
+				const inflationObject = ini.parse(contents).inflation;
+				const inflationInflectionPoints = Object.getOwnPropertyNames(inflationObject).map(key => ({
+					startHeight: BigInt(key.substring(key.lastIndexOf('-') + 1)),
+					rewardAmount: BigInt(inflationObject[key])
+				}));
+
+				// sort by start height
+				inflationInflectionPoints.sort((lhs, rhs) => {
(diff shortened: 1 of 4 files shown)


client/rest/src/routes/networkRoutes.js
```diff
@@ -87,6 +107,47 @@ module.exports = {
...
+			}));
+
+		server.get('/network/inflation/at/:height', (req, res, next) => readAndParseInflationPropertiesFile()
+			.then(inflationInflectionPoints => {
+				const height = BigInt(req.params.height);
+
+				const findMatchingPoint = () => {
+					const firstPoint = inflationInflectionPoints[0];
+					if (height < firstPoint.startHeight)
+						return { startHeight: 'N/A', rewardAmount: '0' };
```
gimre-xymcity, line 130: isn't nemesis block falling into this branch?

Jaguar0625: yes, and, imo, it should as it has 0 block rewards

client/rest/src/routes/networkRoutes.js
```diff
@@ -47,6 +47,26 @@ module.exports = {
...
+			services.config.apiNode.inflationPropertyFilePath,
+			contents => {
+				const inflationObject = ini.parse(contents).inflation;
+				const inflationInflectionPoints = Object.getOwnPropertyNames(inflationObject).map(key => ({
+					startHeight: BigInt(key.substring(key.lastIndexOf('-') + 1)),
+					rewardAmount: BigInt(inflationObject[key])
+				}));
+
+				// sort by start height
+				inflationInflectionPoints.sort((lhs, rhs) => {
```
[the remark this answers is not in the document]
Jaguar0625: yes, right now, it needs to be sorted for the client
but idk if the JS properties parser guarantees order so i wanted to explicitly sort here to future proof it

gimre-xymcity (approved):

PR #521, opened 2023-01-17
[jenkins] fix: ccache for catapult Windows build

Wayonb:
Catapult Windows builds takes a long time to build.

## What's the issue?
There was no caching on Windows for the build files.

ccache is now enabled as part of the Catapult build on Windows.  This reduces the build time by 40% to around 25 mins.
By default is it off due to an issue when ccache is installed by a package manager, Scoop or chocolatey, they will install a shim in the PATH instead of the actual binary for ccache.  CMake will copy the shim version.
It also requires using ``/Z7`` instead of ``Zi`` for generating the debug files
https://github.com/ccache/ccache/wiki/MS-Visual-Studio

Also change to use the latest Powershell 7 which uses UTF-8 by default.

By running in Jenkins.

### DIFF
--- a/jenkins/catapult/templates/WindowsTestBaseImage.Dockerfile
+++ b/jenkins/catapult/templates/WindowsTestBaseImage.Dockerfile
@@ -16,5 +16,13 @@ RUN pwsh.exe -ExecutionPolicy RemoteSigned -Command `
 	python3 -m pip install --upgrade pip; `
 	python3 -m pip install --upgrade colorama cryptography gitpython ply pycodestyle pylint pylint-quotes PyYAML
 	
-ENTRYPOINT ["pwsh.exe", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'Continue'; $verbosePreference='Continue';"]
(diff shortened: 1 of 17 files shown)


Jaguar0625 (approved):
Nice job! 👍

Jaguar0625 (approved):

jenkins/catapult/templates/WindowsTestBaseImage.Dockerfile
```diff
@@ -16,5 +16,13 @@ RUN pwsh.exe -ExecutionPolicy RemoteSigned -Command `
 	python3 -m pip install --upgrade pip; `
 	python3 -m pip install --upgrade colorama cryptography gitpython ply pycodestyle pylint pylint-quotes PyYAML
 	
-ENTRYPOINT ["pwsh.exe", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'Continue'; $verbosePreference='Continue';"]
+RUN "'ACP', 'OEMCP', 'MACCP' | Set-ItemProperty 'HKLM:\SYSTEM\CurrentControlSet\Control\Nls\CodePage' -Name { $_ } 65001"; `
```
Jaguar0625, line 19: i guess this isn't working ?

Jaguar0625 (approved):

PR #567, opened 2023-03-06
Missing models for NEM non-verifiable aggregate (multisig) transaction

gimre-xymcity:

There's no model for NEM 'non-verifiable' multisig.

Added proper model + cosignatures are not included inside non-verifiable, as they screw up hash/signature calculations.

Tested in NEM network 😎


Wayonb (approved):
Looks good. 👍🏾

PR #575, opened 2023-03-14
[jenkins] fix: remove openssl 1.1.1f dependency

Wayonb:
## What's the issue?
build of the ci javascript and cpp images are failing due to the removal of libssl1.1_1.1.1f-1ubuntu2.16_amd64.deb from ubuntu
https://jenkins.symboldev.com/job/infra/job/build-ci-image/15/console

Upgrade to mongodb 6.0.4 which does not need openssl by default.

build the images locally to test

### DIFF
--- a/client/catapult/scripts/ci/setup_test.sh
+++ b/client/catapult/scripts/ci/setup_test.sh
@@ -5,7 +5,7 @@ set -ex
 mongod --fork --logpath /data/mongod.log
 
 wait_period=10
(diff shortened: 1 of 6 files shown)


Wayonb:
> Is everything working as expected with this change? (indexes, storing data in db, etc?)

I havent checked indexes yet.  Storing data in db seems fine plus tests are passing.
This change will only affects CI pipeline then next
1. update the Catapult tests
2. update the bootstrap

gimre-xymcity (approved):

Issue #522, opened 2023-01-18
[client/rest] Exclude treasure/ecosystem addresses from circulating supply


cryptoBeliever:
## Current Behaviour

Currently, we are not excluding treasure/ecosystem addresses from the circulating supply.

## Expected Behavior

According to Coinmarketcap documentation (https://docs.google.com/presentation/d/18rJnnx6Fyg8rAcUI1-TBd2_XKLiXi4tOYawF6tk4cyw/edit#slide=id.g10772dbf710_0_11):
> we would typically exclude the addresses balances of wallets that are labeled as private investors, marketing operations, treasury, ecosystem, incentives, and team/advisors/contractors regardless of whether they are unlocked.

Because of that, we have to change `/network/currency/supply/circulating` endpoint to exclude balances from:
- Symbol Protocol Treasury (NCHEST3QRQS4JZGOO64TH7NFJ2A63YA7TPM5PXI)
- Quadratic Funding (NBANKCSJRQ7O34R7ORKSSGV2GDX6UA4J3AHVCRI)
- Harvest Network Fee Sink (NAVORTEX3IPBAUWQBBI3I3BDIOS4AVHPZLCFC7Y)
- Namespace and Mosaic Fee Sink (NCVORTEX4XD5IQASZQEHDWUXT33XBOTBMKFDCLI)

cryptoBeliever:
This change was introduced in PR: https://github.com/symbol/symbol/pull/577

cryptoBeliever:
Works correctly. After updating rest.json with new excluded public keys circulating supply excluded new addresses. Current output:
- circulating: 5760127433.357264
- total: 8194632944.372308
- max: 8999999999.000000

PR #577, opened 2023-03-15
[client/rest]: exclude protocol treasury from the circulating supply

cryptoBeliever:

Issue: https://github.com/symbol/symbol/issues/522

We are not excluding treasure/ecosystem addresses from the circulating supply.

## What's the issue?

According to Coinmarketcap documentation [CMC DOC](https://docs.google.com/presentation/d/18rJnnx6Fyg8rAcUI1-TBd2_XKLiXi4tOYawF6tk4cyw/edit#slide=id.g10772dbf710_0_11)

> we would typically exclude the addresses balances of wallets that are labeled as private investors, marketing operations, treasury, ecosystem, incentives, and team/advisors/contractors, regardless of whether they are unlocked.

Exclude Quadratic Funding and Symbol Protocol Treasury. Currently excluded addresses:

- Symbol Protocol Treasury (NCHEST3QRQS4JZGOO64TH7NFJ2A63YA7TPM5PXI)
- Quadratic Funding (NBANKCSJRQ7O34R7ORKSSGV2GDX6UA4J3AHVCRI)
- Harvest Network Fee Sink (NAVORTEX3IPBAUWQBBI3I3BDIOS4AVHPZLCFC7Y)
- Namespace and Mosaic Fee Sink (NCVORTEX4XD5IQASZQEHDWUXT33XBOTBMKFDCLI)

After approval, the same changes have to be applied to bootstrap configs.


client/rest/resources/rest.json
```diff
@@ -14,7 +14,9 @@
   },
   "uncirculatingAccountPublicKeys": [
     "A4739036FD7EFED2750A51EE9D1D3113BA3F9849E0889213CED7F221B2AA1A20",
-    "2BF1E1F3072E3BE0CD851E4741E101E33DB19C163895F69AA890E7CF177C878C"
+    "2BF1E1F3072E3BE0CD851E4741E101E33DB19C163895F69AA890E7CF177C878C",
+    "E85D10BF47FFBCE2230F70CB43ED2DDE04FCF342524B383972F86EA1FF773C79",
```
gimre-xymcity, line 18: does node.js allow comments within json files? (it'd be quite helpful here...)

gimre-xymcity (approved):
Imo ok, but ping @0x6861746366574

client/rest/resources/rest.json
cryptoBeliever: JSON does not allow comments in `//` form. It's possible in the JSON5 subset, which is not used here. I could add a property like "_comment" before this uncirculatingAccountPublicKeys, but not sure if it will be readable.

Issue #203, opened 2022-05-03
[client/rest] node/health shows database down with MongoClient 4.4.1 


Wayonb:
There is an API change in MongoClient 4.x release which is causing REST to show mongodb as down.

In mongodb 4.4.1, ``db.database.serverConfig`` is undefined
In mongodb 3.7.1, the object is valid.

```
			const dbStatusPromise = new Promise((resolve, reject) => {
				winston.verbose(`connected to mongo with ${db.database.serverConfig}`);
				if (db.database.serverConfig.isConnected())
					resolve();
				else
					reject();
			});
```
```
rest-gateway-0       | verbose: connected to mongo with undefined {"timestamp":"2022-05-03T00:50:30.811Z"}
```

https://github.com/mongodb/node-mongodb-native/blob/HEAD/etc/notes/CHANGES_4.0.0.md#removed-deprecations
``Remove MongoClient.isConnected - calling connect is a no-op if already connected``

Jaguar0625:
@gimre-xymcity  - i was looking at this issue [https://github.com/symbol/symbol/issues/203]

i think the best approach is to replace the check promise db.database.serverConfig.isConnected() with a actual query, like db.chainStatisticCurrent() 

the good news is that it works and will replace database as up or down correctly. 

the bad news is when down, it keeps retrying to make a connection for 30s, so it's impractical. i've been going through the mongo config, but i don't see a way to execute a findOne without retries. 🤔

gimre-xymcity:
@Jaguar0625 our connection happens inside `connector.connectToDatabase()`

current mongo docs say, that:
a) we probably should do "ping" after connection is made (`await client.db("admin").command({ ping: 1 });`): https://www.mongodb.com/docs/drivers/node/current/fundamentals/connection/
and
https://www.mongodb.com/docs/manual/reference/command/ping/

b) Not sure if this will help with anything, but there's `connectTimeoutMS` option when connecting (first link above)

cryptoBeliever:
Works correctly on driver 4.14. I checked on rest images in version 2.4.3 :heavy_check_mark:

PR #579, opened 2023-03-17
[client/rest] fix: pin mongodb version to 3.x until bug fix

Wayonb:
Rest show the database status as down with mongodb 4.x

## What's the issue?
see issue #203

downgrade mongodb to 3.x until bug is fix

Tested with a testnet node


Wayonb:
Closing this since the fix is in 
https://github.com/symbol/symbol/pull/585

PR #585, opened 2023-03-20
[client/rest] fix: rest shows mongodb down 

Wayonb:
Rest endpoint ``/node/health`` shows mongodb database down

## What's the issue?
starting with mongo driver 4.x, mongodb removed the isconnected() method
see #203 

add logic to ping admin db to verify if db is available.
also reduce the connection timeout to 10s but configurable from config.
if mongodb is down it will take 10s for the ``/node/health`` endpoint to return.

Tested on dev box.

### DIFF
--- a/client/rest/resources/rest.json
+++ b/client/rest/resources/rest.json
@@ -41,7 +41,7 @@
     "maxConnectionAttempts": 5,
     "baseRetryDelay": 500,
     "connectionPoolSize": 10,
-    "connectionTimeout": 10000
+    "connectionTimeout": 500
   },
 
   "apiNode": {
--- a/client/rest/resources/rest.json
+++ b/client/rest/resources/rest.json
@@ -40,7 +40,8 @@
     "pageSizeDefault": 20,
     "maxConnectionAttempts": 5,
     "baseRetryDelay": 500,
-    "connectionPoolSize": 10
+    "connectionPoolSize": 10,
+    "connectionTimeout": 10000
   },
 
   "apiNode": {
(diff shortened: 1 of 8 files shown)


client/rest/src/db/CatapultDb.js
```diff
@@ -132,6 +134,15 @@ class CatapultDb {
 		});
 	}
 
+	isConnected() {
+		return this.client.db().admin().ping()
+			.then(() => winston.verbose('Connection to mongodb is established.'))
+			.catch(err => {
+				winston.error(`failed to verify connection to mongodb: ${err}`);
+				return Promise.reject(err);
+			});
```
gimre-xymcity, line 143: not sure, but maybe this should just return promise returned by .ping(), then we could move this logic outside, down into nodeRoutes.

client/rest/resources/rest.json
```diff
@@ -40,7 +40,8 @@
     "pageSizeDefault": 20,
     "maxConnectionAttempts": 5,
     "baseRetryDelay": 500,
-    "connectionPoolSize": 10
+    "connectionPoolSize": 10,
+    "connectionTimeout": 10000
```
gimre-xymcity, line 44: I guess we could make it much smaller, say 500ms (cause during startup, rest should wait for mongo to boot anyway...)

is there any downside of setting it to much lower value/

client/rest/src/db/CatapultDb.js
Wayonb: Ok, let me take a look at this.

client/rest/resources/rest.json
Wayonb: My only concern was with startup. On slower nodes, database startup will take longer.  But should be ok since REST will retry the connection.

PR #596, opened 2023-03-28
[client/catapult] fix: upgrade to openssl 3.1.0 

Wayonb:
## What's the issue?
catapult client is using an older version of openssl

upgrade openssl to version v3.1.0

ran the catapult tests on my dev box.

### DIFF
--- a/client/catapult/src/catapult/crypto/OpensslContexts.cpp
+++ b/client/catapult/src/catapult/crypto/OpensslContexts.cpp
@@ -37,11 +37,15 @@ namespace catapult { namespace crypto {
 	// region OpensslDigestContext
 
 	OpensslDigestContext::OpensslDigestContext() {
-		m_buffer = UniquePtr(EVP_MD_CTX_new(), [](context_type* p) { EVP_MD_CTX_free(p); });
+		std::memset(&m_buffer, 0, CountOf(m_buffer));
+	}
+
+	OpensslDigestContext::~OpensslDigestContext() {
+		reset();
 	}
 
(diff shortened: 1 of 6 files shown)


client/catapult/src/catapult/crypto/OpensslContexts.cpp
```diff
@@ -37,15 +37,11 @@ namespace catapult { namespace crypto {
 	// region OpensslDigestContext
 
 	OpensslDigestContext::OpensslDigestContext() {
-		std::memset(&m_buffer, 0, CountOf(m_buffer));
-	}
-
-	OpensslDigestContext::~OpensslDigestContext() {
-		reset();
+		m_buffer = UniquePtr(EVP_MD_CTX_new(), [](context_type* p) { EVP_MD_CTX_free(p); });
```
Jaguar0625, line 40: this is doing an allocation - that defeats the purpose of these contexts to be stack allocated 😅

Jaguar0625 (changes requested):

client/catapult/src/catapult/crypto/OpensslContexts.cpp
Wayonb: haha, I was wondering this.  I guess you want stack allocation for performance?

Jaguar0625: yep, would prefer to keep that, if possible

Jaguar0625 (approved):
👍

PR #600, opened 2023-03-29
[sdk]: add support for MultisigTransactionV2 transactions

Jaguar0625:
problem: currently user needs to manually change version from 1 to 2
solution: add duplicate transaction with version default 2


gimre-xymcity (changes requested):
Remind me, since when do we have multisig V2 in NEM? 🤐

Jaguar0625:
> Remind me, since when do we have multisig V2 in NEM? 🤐
> 
> disclaimer: _afair_ we don't, we only have aggregate modification, which in symbol terms is "multisig account modification"

we have been using symbol names in NEM sdk so that there is consistent terminology across NEM and Symbol sdks

Jaguar0625:
invalid

PR #607, opened 2023-04-06
[client/catapult]: AddResolvedAddresses is not properly applying resolutions from previous transactions

Jaguar0625:
problem: if a resolution entry is added for a transaction, it should stay in scope for the rest of the block
              AddResolvedAddresses is incorrectly scoping a resolution entry to the transaction that triggered it
    solution: extend the scope of a resolution entry to last until the end of the block

### DIFF
--- a/client/catapult/extensions/addressextraction/src/AddressExtractor.cpp
+++ b/client/catapult/extensions/addressextraction/src/AddressExtractor.cpp
@@ -55,17 +55,38 @@ namespace catapult { namespace addressextraction {
 				model::AddressSet& addresses,
 				uint32_t primaryId,
 				const model::AddressResolutionStatement& resolutionStatement) {
+			auto shouldAddPreviousResolution = true;
 			for (auto i = 0u; i < resolutionStatement.size(); ++i) {
-				const auto& resolutionEntry = resolutionStatement.entryAt(i);
+				const auto& resolutionEntry = resolutionStatement.entryAt(resolutionStatement.size() - i - 1);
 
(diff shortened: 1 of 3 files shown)


client/catapult/extensions/addressextraction/src/AddressExtractor.cpp
```diff
@@ -55,17 +55,38 @@ namespace catapult { namespace addressextraction {
...
+					// there is not enough information to determine whether or not 1 has occured:
+					//     1. old alias is used
+					//     2. alias is unlinked
+					//     3. new alias is linked
+					//     4. new alias is used
+					// in this ambiguous situation, include the old alias (potential false positive)
+
+					// for a non-aggregate transaction, a single unresolved address can never map multiple resolved addresses
+					if (!resolutionEntry.Source.SecondaryId)
+						shouldAddPreviousResolution = false;
```
gimre-xymcity, line 80: hmm, why is this needed?

Jaguar0625: in the case it is a non-aggregate transaction that is using an alias, either:
1. the transaction will be first one using the alias and have a resolution entry with matching primaryId
2. the transaction will NOT be first one using the alias (previous transaction in block would have used it and created resolution entry), so it should use previous resolution entry for address

in case of 1, previous resolution entry isn't needed, so shouldAddPreviousResolution is set to false

gimre-xymcity (approved):

PR #621, opened 2023-04-13
[sdk]: state proof validation is incorrectly building state path 

Jaguar0625:
problem: when node has odd length path, an extra trailing 0 is appended
    solution: properly handle nodes with odd length paths


sdk/javascript/src/symbol/merkle.js
```diff
@@ -331,7 +352,7 @@ export const provePatriciaMerkle = (encodedKey, valueToTest, merklePath, stateHa
 		}
 
 		childHash = nodeHash;
-		actualPath = `${formattedLinkIndex}${uint8ToHex(node.path.path)}${actualPath}`;
+		actualPath = `${formattedLinkIndex}${node.hexPath}${actualPath}`;
```
Jaguar0625, line 355: this is location of bug

in the case node path had odd nibble, original code would return full byte (resulting in extra 0 nibble in path).

correction is to trim it, which is now happening in hexPath

similar bug (and fix) applies to python sdk

gimre-xymcity (approved):

PR #622, opened 2023-04-13
[client/catapult] fix: catapult client ionet failing on Windows

Wayonb:
catapult ionet tests occasionally fails on Windows

## What's the issue?
catapult client ionet tests are failing on Windows due to race condition due to the server socket getting closed before the read is complete by the client.

add synchronization between the server and client threads so the server thread waits for read to be completed before the server thread exit.

tested local on Windows and Linux nodes.
Havent been able to reproduce the issue on Windows.

### DIFF
--- a/client/catapult/tests/catapult/ionet/PacketSocketTests.cpp
+++ b/client/catapult/tests/catapult/ionet/PacketSocketTests.cpp
@@ -66,7 +66,7 @@ namespace catapult { namespace ionet {
 					writeCode = code;
 				});
 
-				WAIT_FOR(readComplete);
+				test::waitForReadCompleteOrLog(readComplete);
 			});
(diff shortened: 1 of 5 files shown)


Jaguar0625:
> server socket getting closed before the read is complete by the client.

do you know what IS read by the client in that case ?

Wayonb:
> do you know what IS read by the client in that case ?

The will read only part of the buffer and you will see a message like 
```
2023-04-12 17:30:48.811176 0x00004060: <debug> (net::ClientSocket.cpp@181) client socket read 118792 bytes - with code An existing connection was forcibly closed by the remote host
```

If successful then this message is logged.
```
2023-04-13 04:16:48.891266 0x000038d0: <debug> (net::ClientSocket.cpp@181) client socket read 204800 bytes - success!
```

Jaguar0625 (approved):

Jaguar0625 (approved):
👍

PR #664, opened 2023-05-12
[sdk/python]: add special handling for encrypted messages created by Symbol wallets

Jaguar0625:
problem: old Symbol TS SDK hex encodes binary data before encrypting it
solution: add deprecated functions so SDK users can easily interop with wallet encrypted messages


sdk/python/symbolchain/symbol/MessageEncoder.py
```diff
@@ -57,3 +62,36 @@ def encode(self, recipient_public_key: PublicKey, message: bytes):
...
+			'This function is only provided for compatability with the original Symbol wallets.\n'
+			'Please use `try_decode` in any new code.',
+			category=DeprecationWarning,
+			stacklevel=2)
+
+		if 1 == encoded_message[0]:
+			try:
+				# wallet additionally hex encodes
+				return self.try_decode(recipient_public_key, bytes([1]) + unhexlify(encoded_message[1:].decode('utf8')))
+			except (UnicodeDecodeError, BinasciiError):
```
gimre-xymcity, line 82: what about `TypeError: Non-hexadecimal digit found` ?

Jaguar0625: that is raised as `binascii.Error`

PR #672, opened 2023-05-18
[jenkins] fix: update multibranch jobs to use the full folder path in the job name

Wayonb:
 Currently Jenkins uses the project's folder name as the name of the Jenkins multibranch job.

## What's the issue?
 if there are two projects in the same repo with the same leaf folder name, there will be a conflict in Jenkins multibranch job name.

create the Jenkins job name from the full folder path from the root but replace '/' with '-'

Will test in Jenkins


jenkins/shared-library/vars/createMonorepoMultibranchJobs.groovy
```diff
@@ -49,8 +50,12 @@ void generatePackageMultibranchJobs(Object buildConfiguration, Map jobConfigurat
 		jobConfiguration.displayName = build.name
 		jobConfiguration.packageFolder = Paths.get(jobConfiguration.repositoryName.toString()).resolve(build.path).toString()
 		jobConfiguration.daysToKeep = build.logRotation
+		renameJobIfDisplayNamePresent(jobConfiguration.fullBranchFolder, jobConfiguration.displayName, jobConfiguration.jobName)
```
Jaguar0625, line 53: do we need this ?

Wayonb: Thought it would be good to keep job history.
If you dont think its necessary will remove it.  It will make it cleaner also.

Jaguar0625: that was my guess, just wanted to confirm

Jaguar0625 (approved):

PR #687, opened 2023-06-04
[jenkins] fix: building catapult test image failure

Wayonb:
Catapult Ubuntu/Debian test image fails to build perf module from the Linux kernel due to libtraceevent is required.

## What's the issue?
The latest Linux kernel perf module defaults to requiring libtraceevent library

disable the requirement for libtraceevent when building the perf module from Linux kernel.

Built the test image locally


Wayonb:
It seem that these Dockerfiles can be combined since they are pretty similar.

Jaguar0625 (approved):

PR #708, opened 2023-06-20
[jenkins] feat: Add ARM support in build infrastructure

Wayonb:
Our build system does not support arm64.  Currently only amd64 architecture is supported.

## What's the issue?
Need to support Symbol/NEM running on arm64 architecture

Jenkins job has architecture option to choose whether to build amd64 or arm64.  Currently the default is to build amd64.
When building an image release, it will now create two tags.
1. An architecture specific image which will end with either arm64 or amd64.
e.g. I built a test arm64 image for the catapult server is here 
https://hub.docker.com/layers/symbolplatform/symbol-server/gcc-test-1.0.3.6-arm64/images/sha256-73be3e44a762f5024bc5c1fd3d7507bbdd52d99eaca0056821dd62740d8cebbc?context=repo

2. A multi-architecture tag is created that will contains an image for all the architectures in the release
https://hub.docker.com/layers/symbolplatform/symbol-server/gcc-test-1.0.3.6/images/sha256-1c015c264be4fb54f865e9466299f5590330e2e22df2fec8484be4dc897fa044?context=repo

Each architecture job is run independently and is build on native host architecture.  Jenkins will create the multi-architecture tag with the last images for each architecture.

Update the catapult base scripts to build deps for arm64.

Jenkinsfile also updated to separate OS and instance size.

Tested in Jenkins.


jenkins/shared-library/vars/dockerHelper.groovy
```diff
@@ -0,0 +1,78 @@
+void updateDockerImage(String targetImageName, String sourceImageName, String architecture) {
```
Jaguar0625, line 1: looks good, but what changed that we need this additional logic now?

jenkins/shared-library/vars/dockerHelper.groovy
```diff
@@ -0,0 +1,78 @@
...
+		}
+		else {
+			// This is a multi-architecture image
+			List<String> imageDigests = resolveDockerImageDigests(packageJson, sourceImageName, architecture)
+			String digestList = imageDigests.join(' ')
+			print("Digest list: ${digestList}")
+			runScript("docker buildx imagetools create --tag ${targetImageName} ${digestList}")
+		}
+
+		runScript("docker buildx imagetools inspect ${targetImageName}")
```
Jaguar0625, line 31: is this just for diagnostics?

jenkins/catapult/jenkins/catapult-client-build-base-image-all.groovy
```diff
@@ -28,120 +28,208 @@ pipeline {
 
 		stage('build base images') {
 			parallel {
-				stage('gcc prior') {
+				stage('gcc prior - amd64') {
```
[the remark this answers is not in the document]
Wayonb: You mean move all the stages to a function that takes amd64|arm64?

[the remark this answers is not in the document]
Wayonb: or I was thinking of configuring the job to run either amd64|arm64 and trigger them on different day.

jenkins/catapult/jenkins/catapult-client-build-base-image.groovy
```diff
@@ -101,13 +108,38 @@ pipeline {
 						}
 					}
 				}
+				stage('build final image') {
```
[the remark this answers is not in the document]
Wayonb: I guess its not really needed.  I added it to push the final docker image.  Let see if I can move it into the stage it was build

jenkins/catapult/runDockerTests.py
```diff
@@ -74,9 +75,11 @@ def main():
 	compose_filename += '.yaml'
 	compose_template_filepath = Path(__file__).parent / 'templates' / compose_filename
 	print(f'processing template from {compose_template_filepath}')
+	compiler_filepath = Path(args.compiler_configuration)
+	compiler_base_path = f'{compiler_filepath.parent.name}/{compiler_filepath.name}'
```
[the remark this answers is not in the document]
Wayonb: compiler_filepath is the full path from the git root
compiler_base_path is the compiler filename and its parent path.  so it would be like ``amd64/gcc-latest.yaml``

jenkins/shared-library/vars/dockerHelper.groovy
Wayonb: This function is to create the multi-architecture image.
ofc, docker support for this is weak and not well design, imo. 😅 
The docker buildx command will append to an existing image but it doesnt override the current platform if it exists already.  So have to manually create the image in certain case.

Jaguar0625: consider adding comment to that effect in the code

jenkins/shared-library/vars/dockerHelper.groovy
Jaguar0625: consider adding comment

jenkins/catapult/jenkins/catapult-client-build-compiler-image-all.groovy
```diff
@@ -33,58 +46,68 @@ pipeline {
...
 				stage('clang latest') {
 					steps {
 						script {
-							dispatchBuildCompilerImageJob('clang-latest', 'ubuntu')
+							dispatchBuildCompilerImageJob('clang-latest', 'ubuntu', "${ARCHITECTURE}")
 						}
 					}
 				}
 
 				stage('msvc latest') {
```
Jaguar0625, line 90: should you make these amd64 only tasks conditional on arch in daily and weekly workflows, like in image workflows?

Wayonb: yea, my thoughts behind tests were to run all the jobs independent of architecture.  So we ran mainly arm64 tests during the week, the sans tests will still get executed.  Currently the default is still run jobs on amd64.

Jaguar0625 (approved):

PR #709, opened 2023-06-21
[jenkins] fix: postgres docker image fails to build

Wayonb:
postgres docker image fails to build since its not able to find Java 11 package to install.

## What's the issue?
This seems postgres:15 update the base image to debian:bookworm-slim which only have Java 17.

switch the postgres ci image to use the postgres:bulleye as the base image.

Also add missing checkout step in the CI job.

Test by building the docker image local

### DIFF
--- a/jenkins/infra/jenkins/build-ci-image.groovy
+++ b/jenkins/infra/jenkins/build-ci-image.groovy
@@ -45,6 +45,16 @@ pipeline {
 				}
 			}
 		}
+		stage('checkout') {
+			when {
+				triggeredBy 'UserIdCause'
+			}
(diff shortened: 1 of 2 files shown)


jenkins/infra/jenkins/build-ci-image.groovy
```diff
@@ -45,6 +45,16 @@ pipeline {
 				}
 			}
 		}
+		stage('checkout') {
```
Jaguar0625, line 48: seems unrelated to postgres, so would probably merge as two commits

Jaguar0625 (approved):

PR #722, opened 2023-07-06
[jenkins] fix: docker build hangs due to network

Wayonb:
npm install hangs when trying to connect sites to install packages during docker build

## What's the issue?
The newer version of npm tries to use ipv6 first which is configured in docker but not the host.

add --network host flag to use the host network which does not have IPv6 configured.  

Ran a test locally in docker.


jenkins/shared-library/vars/publish.groovy
```diff
@@ -30,7 +30,9 @@ void dockerPublisher(Map config, String phase) {
 	final String imageVersionName = "${config.dockerImageName}:${version}"
 	final String archImageName = imageVersionName + "-${ARCHITECTURE}"
 	dockerHelper.loginAndRunCommand(DOCKER_CREDENTIALS_ID) {
-		dockerHelper.dockerBuildAndPushImage(archImageName, config.dockerBuildArgs ?: '.')
+		String args = config.dockerBuildArgs ?: '.'
+		args = '--network host ' + args
```
Jaguar0625, line 34: is `--network host .` valid ?

Wayonb: it will run as ``docker build -t $archImageName --network host .``

Jaguar0625: i was more asking if '.' was valid, since it seems to not be added when there are args? or is it intended to be part of args

Wayonb: yea ``docker build`` takes a path for its current working directory.  The default is '.' but if a project override with docker args then the Jenkinsfile need to also add this path because it can be different.

Jaguar0625 (approved):

PR #748, opened 2023-07-24
[jenkins]: add support to build different version of python images

Wayonb:
Test are only run in Ubuntu 22.04

## What's the issue?
shoestring is run on multiple version of Ubuntu which leads to failures for end users on Ubuntu 20.04 and Ubuntu 23.04.

create test images for Ubuntu 20.04 and 23.04 which will be run on a weekly schedule.

run docker build locally to test.

### DIFF
--- a/jenkins/docker/python.Dockerfile
+++ b/jenkins/docker/python.Dockerfile
@@ -1,4 +1,6 @@
-FROM ubuntu:22.04
+ARG FROM_IMAGE='ubuntu:22.04'
+
+FROM ${FROM_IMAGE}
 
 # install tzdata first to prevent 'geographic area' prompt
 RUN apt-get update >/dev/null \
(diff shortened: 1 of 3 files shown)


jenkins/docker/python.Dockerfile
```diff
@@ -31,7 +31,14 @@ RUN ARCH=$([ "$(uname -m)" = "x86_64" ] && echo "linux" || echo "aarch64") \
...
 
 # add ubuntu user (used by jenkins)
-RUN useradd --uid 1000 -ms /bin/bash ubuntu
-ENV PATH=$PATH:/home/ubuntu/.local/bin
-
+RUN id -u "ubuntu" || useradd --uid 1000 -ms /bin/bash ubuntu
+USER ubuntu
 WORKDIR /home/ubuntu
+
+# create virtual environment which is required by Ubuntu 23.04
```
Jaguar0625, line 38: , before which

jenkins/infra/jenkins/build-ci-image-all.groovy
```diff
@@ -43,46 +43,60 @@ pipeline {
 
 		stage('build ci images') {
 			parallel {
-				stage('cpp') {
+				stage('cpp - ubuntu:22.04') {
 					steps {
 						script {
-							dispatchBuildCiImageJob('cpp')
+							dispatchBuildCiImageJob('cpp', 'ubuntu:22.04')
```
Jaguar0625, line 49: can you not use default argument instead of specifying 22.04 everywhere?

jenkins/infra/jenkins/build-ci-image.groovy
```diff
@@ -7,6 +7,9 @@ pipeline {
 		choice name: 'ARCHITECTURE',
 			choices: ['amd64', 'arm64'],
 			description: 'Computer architecture'
+		choice name: 'BASE_IMAGE',
+				choices: ['ubuntu:22.04', 'ubuntu:20.04', 'ubuntu:23.04'],
+				description: 'Base image'
```
Jaguar0625, line 12: extra indent

jenkins/infra/jenkins/build-ci-image-all.groovy
Wayonb: Yea, I had a default parameter in first commit but then change it to make it more readable.
Since the stage names are updated, it make sense to use the default parameter.

Jaguar0625 (approved):

jenkins/infra/jenkins/build-ci-image-all.groovy
```diff
@@ -43,49 +43,63 @@ pipeline {
 
 		stage('build ci images') {
 			parallel {
-				stage('cpp') {
+				stage('cpp - ubuntu:22.04') {
```
Jaguar0625, line 46: not sure if we need `ubuntu:22.04` in description vs something like more general like `default` or nothing, as latter would make updating default Ubuntu OS fewer changes

Wayonb: yea, I thought the same also.

Jaguar0625 (approved):

PR #749, opened 2023-07-24
[sdk]: add function for looking up friendly transaction name given type and version

Jaguar0625:
f(type, version) => friendly name

requested by community

### DIFF
--- a/catbuffer/schemas/nem/transaction_type.cats
+++ b/catbuffer/schemas/nem/transaction_type.cats
@@ -16,7 +16,7 @@ enum TransactionType : uint32
 	MULTISIG_COSIGNATURE = 0x1002
 
 	# multisig transaction
-	MULTISIG_TRANSACTION = 0x01004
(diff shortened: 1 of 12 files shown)


catbuffer/schemas/nem/transaction_type.cats
```diff
@@ -16,7 +16,7 @@ enum TransactionType : uint32
 	MULTISIG_COSIGNATURE = 0x1002
 
 	# multisig transaction
-	MULTISIG_TRANSACTION = 0x01004
+	MULTISIG = 0x01004
```
gimre-xymcity, line 19: hmm, not that I think any code used it, but isn't this breaking change?

Jaguar0625: it is breaking if anyone is using the enum value directly, but the name used to create tx is unchanged as multisig_transaction_v#

gimre-xymcity (approved):

PR #763, opened 2023-08-08
[jenkins]: Add a custom publish script

Wayonb:
Publishing a package is not customizable by each project.

## What's the issue?
Projects can only publish one artifact from each project.

A project can provide a publishing script that will allow custom steps to publish an artifact.

Testing in Jenkins

### DIFF
--- a/jenkins/shared-library/vars/publish.groovy
+++ b/jenkins/shared-library/vars/publish.groovy
@@ -77,9 +77,6 @@ void pythonPublisher(Map config, String phase) {
 	}
 
 	withCredentials([string(credentialsId: credentialsId, variable: 'PYPI_TOKEN')]) {
-		env.TWINE_USERNAME = '__token__'
-		env.TWINE_PASSWORD = env.PYPI_TOKEN
-
 		publishArtifact {
 			Object requirementsFile = readFile 'requirements.txt'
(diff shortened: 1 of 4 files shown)


jenkins/shared-library/vars/publish.groovy
```diff
@@ -72,22 +76,27 @@ void pythonPublisher(Map config, String phase) {
...
-		if (isAlphaRelease(phase)) {
-			runScript('poetry config "repositories.test" "https://test.pypi.org/legacy/"')
-			runScript("poetry config 'pypi-token.test' ${POETRY_PYPI_TOKEN_PYPI}")
-			runScript('poetry publish --repository "test"')
-		} else {
-			runScript("poetry config 'pypi-token.pypi' ${POETRY_PYPI_TOKEN_PYPI}")
-			runScript('poetry publish')
+	withCredentials([string(credentialsId: credentialsId, variable: 'PYPI_TOKEN')]) {
+		env.TWINE_USERNAME = '__token__'
+		env.TWINE_PASSWORD = env.PYPI_TOKEN
```
Jaguar0625, line 81: if possible, it would be better to set these in publish.sh to avoid the twine knowledge here

if not possible, ok, but might add comment

Jaguar0625 (approved):
👍

PR #779, opened 2023-09-01
[jenkins] feat: add Windows ci builds

Wayonb:
Projects can only run CI in Ubuntu image

## What's the issue?
Projects are not able to run Windows CI builds.

Add support for Windows CI builds 
Add Windows CI build images
Windows image has Cygwin installed so scripts can be run with bash

Testing in Jenkins.


jenkins/docker/windows/javascript.Dockerfile
```diff
@@ -0,0 +1,29 @@
+# escape=`
+
+ARG FROM_IMAGE=symbolplatform/symbol-server-compiler:windows-msvc-17
+
+FROM ${FROM_IMAGE}
+LABEL maintainer="Catapult Development Team"
+
+SHELL ["pwsh","-command", "$ErrorActionPreference = 'Stop';"]
```
Jaguar0625, line 8: space after first comma?  (py too)

jenkins/docker/windows/python.Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+	$command='c:\scoop.ps1 -RunAsAdmin'; `
+	iex $command; `
+	del c:\scoop.ps1; `
+	scoop install python git shellcheck openssl cmake cygwin; `
+	python3 -m pip install --upgrade pip; `
+	python3 -m pip install --upgrade gitlint wheel
+
+# Install codecov
+RUN Invoke-WebRequest -Uri https://uploader.codecov.io/latest/windows/codecov.exe -Outfile c:\Windows\System32\codecov.exe
+
```
Jaguar0625, line 25: i guess you don't need to restore default shell because no more commands are run here?

jenkins/shared-library/vars/defaultCiPipeline.groovy
```diff
@@ -212,10 +212,10 @@ void call(Closure body) {
 							allOf {
 								expression {
 									// The branch indexing build TEST_MODE = null
-									return env.TEST_MODE == null || 'code-coverage' == env.TEST_MODE
+									return null == params.TEST_MODE || 'code-coverage' == params.TEST_MODE
```
Jaguar0625, line 215: why this change from env to params?

jenkins/shared-library/vars/weeklyBuildPipeline.groovy
```diff
@@ -135,11 +135,11 @@ void triggerAllJobs(
 
 		environments.each { environment ->
 			String stageName = "${displayName} (${environment})"
+			String osValue = environment.split('-')[1]
```
Jaguar0625, line 138: i thought tags were {operating-system}-{base-image} (from baseImages.yaml) or is this something else?

jenkins/docker/windows/python.Dockerfile
Wayonb: Yeah, in the Javascript Dockerfile to set the PATH with ``setx`` the shell needs to be cmd.
The entry point is the same for both images

jenkins/shared-library/vars/defaultCiPipeline.groovy
Wayonb: This was more of a cleanup of code.  Should probably be in a different PR 😅 
Parameters selected by the users can be accessed either by ``env`` or ``params``.  I am switching these to use just ``params`` for readability.

jenkins/shared-library/vars/weeklyBuildPipeline.groovy
Wayonb: yeah, that's the tag for the base image.  The environment tag includes the environment name, so it would be {environmentName}-{operating-system}-{base-image}.  Something like ``Python-Windows-lts`` or ``Javascript-ubuntu-lts``.
Let me know if you have a better idea

Wayonb: I refactor a bit to simplify and hopefully better readability.  Let me know if this is better.

Jaguar0625 (approved):
👍

PR #792, opened 2023-09-14
[sdk] fixes for node 18 AND  fixes for python 3.11

Jaguar0625:
[sdk/python]: fixes for python 3.11
    
     problem: 3.11+ removes named zero flags from Flag class
    solution: manually add 'none' => 0 mapping in add_flags_parser

    [sdk/javascript]: fixes for Node 18
    
     problem: exceptions filtered by MessageEncoder (NEM) are too specific
    solution: use more generic filters that work with both Node 16 and 18


Wayonb (commented):
👍🏾

Wayonb (approved):

PR #803, opened 2023-09-27
[jenkins] fix: only run unit test during CI

Wayonb:
All tests are run during CI.

## What's the issue?
Some test suite(such as example & vector) takes longer to run.

Only run the unit tests during CI.
All the tests are executed during the nightly builds.

Tested in Jenkins.


Jaguar0625 (approved):
👍

PR #804, opened 2023-09-27
[jenkins] feat: add generic AWS publisher to Jenkins

Wayonb:
Jenkins only has permission to its own AWS account.

## What's the issue?
Jenkins cannot publish to different AWS accounts.

Add a generic AWS publisher which will set up credentials for a specific AWS account.
the published script will run to all specific operations on the AWS account.

locally on my dev box.

### DIFF
--- a/jenkins/shared-library/vars/publish.groovy
+++ b/jenkins/shared-library/vars/publish.groovy
@@ -198,13 +198,13 @@ boolean isGitHubRepositoryPublic(String orgName, String repoName) {
 }
 
 void awsPublisher(Map config, String phase) {
-	if (config.publisher != 'aws') {
+	if (config.publisher != 'aws' || !isRelease(phase)) {
 		return
 	}
 
 	withCredentials([usernamePassword(credentialsId: config.awsCredentialId,
 			usernameVariable: 'AWS_ACCESS_KEY_ID',
 			passwordVariable: 'AWS_SECRET_ACCESS_KEY')]) {
-		publishArtifact {}
+		publishArtifact { }
 	}
 }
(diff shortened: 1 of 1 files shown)


jenkins/shared-library/vars/publish.groovy
```diff
@@ -195,3 +196,15 @@ boolean isGitHubRepositoryPublic(String orgName, String repoName) {
 		return false
 	}
 }
+
+void awsPublisher(Map config, String phase) {
```
Jaguar0625, line 200: probably would move this up after githubpagespublisher

jenkins/shared-library/vars/publish.groovy
```diff
@@ -195,3 +196,15 @@ boolean isGitHubRepositoryPublic(String orgName, String repoName) {
...
+	if (config.publisher != 'aws' || !isRelease(phase)) {
+		return
+	}
+
+	withCredentials([usernamePassword(credentialsId: config.awsCredentialId,
+			usernameVariable: 'AWS_ACCESS_KEY_ID',
+			passwordVariable: 'AWS_SECRET_ACCESS_KEY')]) {
```
Jaguar0625, line 207: notice indentation in file is inconsistent - in some cases these are indented twice, others once

Jaguar0625 (approved):

Jaguar0625:
This is needed because we expect to have one AWS account for jenkins / builds and another for production?

jenkins/shared-library/vars/publish.groovy
Wayonb: yea, its my editor doing auto-format. 😬 
I need to switch back to vi 😄

Wayonb:
> This is needed because we expect to have one AWS account for jenkins / builds and another for production?

Yeah, currently Jenkins is in its own AWS account.  To perform any operation in any other account whether testnet/production/strix requires some form of creds for authentication.

PR #805, opened 2023-09-27
[jenkins] fix: add custom workspace for CI builds

Wayonb:
Jenkins includes the branch name as part of the workspace folder name.

## What's the issue?
C++ compilation which uses marco like ``__FILE__``, will cause ccache to have a lot of cache misses due to the differences in path.  This leads to longer compile times for Catapult.
https://ccache.dev/manual/latest.html#_compiling_in_different_directories

Add a custom workspace for CI builds.  This will allow faster builds for Catapult client.

Tested locally and it seems to be faster.


Jaguar0625 (approved):
👍

PR #806, opened 2023-09-29
[jenkins] fix: combine dependapot prs

Wayonb:
Dependabot PRs are merged individually 

## What's the issue?
A lot of Dependabot PRs will cause a lot of jobs in Jenkins which will lead to the starvation of user PR jobs.

After dependabot PR jobs run successfully, they will be combined into one PR and merged.
This will prevent the rebase and rerun of the Jenkins job after each merge.

Ran the workflow locally


Wayonb:
With this change, auto approval does not work 😅 
Since the GitHub Action Bot is creating the PR, it cannot approve its own PR.
I have been looking for a solution but so far it seems it will require adding a personal access token to GitHub.
I have been trying hard not to use a bot 😄

Jaguar0625 (approved):

PR #838, opened 2023-10-27
[jenkins] fix: add a controller job to manage CI builds

Wayonb:
All the CI jobs are triggered by GitHub calling into Jenkins.

## What's the issue?
For new branch/PR, Jenkins will do branch indexing where it does not have knowledge of the change list.
This will cause all the CI jobs to run.

A controller job has been added to manage which CI jobs to run.  The controller job will get the files in the commit and trigger the required CI jobs.
The controller job will get changed list in the following order
1. Check the ``currentBuild.changeSets`` Jenkins object for the files needs to build.
2. If pull request, get the file list from the PR
3. If all else fails, get the list of files from the last commit.

The file list is used to build a list of CI job that needs to be run.  The controller will trigger the jobs and wait for complete.
The controller job will now also create and update the builds when needed.
The seed job is only create the controller job.

Will test in Jenkins.

### DIFF
--- a/jenkins/shared-library/vars/changedSetHelper.groovy
+++ b/jenkins/shared-library/vars/changedSetHelper.groovy
@@ -1,0 +1,52 @@
+// This will return list of files Jenkins thinks have changed, except for new branches where it will return an empty list.
+List<String> jenkinsChangedFiles() {
+	List<String> files = currentBuild.changeSets*.items*.affectedPaths
+	return files.flatten().unique()
+}
+
+// This will return all the files in the pull request
+List<String> pullRequestChangedFiles() {
+	// CHANGE_ID is set only for pull requests
+	return env.CHANGE_ID ? pullRequest.files*.filename : []
+}
+
+// Jenkins native interface to retrieve changes, i.e. `currentBuild.changeSets`, returns an empty list for newly
+// created branches (see https://issues.jenkins.io/browse/JENKINS-14138), so let's use `git` instead.
+List<String> lastCommitChangedFiles() {
+	return runScript('git diff-tree --no-commit-id --name-only -r HEAD', true).trim().split('\n')
+}
+
+List<String> resolveChangedList() {
+	List<Closure> changedFileHandler = [
+		this.&jenkinsChangedFiles,
+		this.&pullRequestChangedFiles,
+		this.&lastCommitChangedFiles
+	]
+
+	for (Closure handler in changedFileHandler) {
+		List<String> changedFiles = handler()
+		if (!changedFiles.isEmpty()) {
+			return changedFiles
+		}
+	}
+
+	return []
+}
+
+boolean isFileInChangedSet(String fileName) {
+	return resolveChangedList().find { changedFilePath -> changedFilePath == fileName }
+}
+
+Map<String, String> findRelevantMultibranchPipelines(List<String> changedFilesPath, Map<String, String> jenkinsBuilds) {
+	return jenkinsBuilds.findAll { build ->
+		changedFilesPath.find { changedFilePath -> changedFilePath.startsWith(build.value) }
+	}
+}
+
+Map<String, String> findMultibranchPipelinesToRun(Map<String, String> jenkinsBuilds) {
+	return findRelevantMultibranchPipelines(resolveChangedList(), jenkinsBuilds)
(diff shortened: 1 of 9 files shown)


jenkins/shared-library/vars/changedSetHelper.groovy
```diff
@@ -0,0 +1,52 @@
...
+	}
+}
+
+Map<String, String> findMultibranchPipelinesToRun(Map<String, String> jenkinsBuilds) {
+	return findRelevantMultibranchPipelines(resolveChangedList(), jenkinsBuilds)
+}
+
+Map <String, String> resolveJenkinsfilesJobToRun(Map<String, String> allJenkinsfileMap) {
+	return findMultibranchPipelinesToRun(allJenkinsfileMap)
+}
```
Jaguar0625, line 52: do you need this? isn't declaration the same as previous function?

jenkins/shared-library/vars/createMultibranchJob.groovy
```diff
@@ -1,5 +1,21 @@
...
+					skipInitialBuildOnFirstBranchIndexing()
+				}
+				strategy {
+					defaultBranchPropertyStrategy {
+						props {
+							noTriggerBranchProperty()
+						}
+					}
+				}
+	'''
```
Jaguar0625, line 16: is this over indented ?

jenkins/shared-library/vars/createMultibranchJob.groovy
```diff
@@ -78,20 +94,15 @@ void call(Map jobConfiguration) {
 											parentCredentials(true)
 										}
 									}
+									gitHubStatusChecks {
```
Jaguar0625, line 97: what does this do?

jenkins/shared-library/vars/monorepoControllerPipeline.groovy
```diff
@@ -0,0 +1,171 @@
+import java.nio.file.Paths
```
Jaguar0625, line 1: i see this is controlled by controller.groovy, but how/when is that triggered? 

each branch/pr creation?

Jaguar0625 (approved):

jenkins/shared-library/vars/changedSetHelper.groovy
Wayonb: yea, not needed. will refactor 👍🏾

jenkins/shared-library/vars/createMultibranchJob.groovy
Wayonb: yea, I was leaving the indent to match where it got inserted below.  but this is probably not needed since no one sees the final output. 😅

jenkins/shared-library/vars/createMultibranchJob.groovy
Wayonb: Sorry, I forgot to mention this.
By default, all the GitHub check status gets reported under ``Jenkins``.  This causes issues with GitHub since the status check will overwrite each other since Jenkins does not combine them.
This will allow each job to report their status to GitHub under a different name.  You will see a lot more information in GitHub. 😄

jenkins/shared-library/vars/monorepoControllerPipeline.groovy
Wayonb: yea, the controller will be triggered for every push/pr.  Then it will figure out which CI job to run.

Jaguar0625 (approved):
👍

jenkins/shared-library/vars/helper.groovy
```diff
@@ -109,3 +109,19 @@ boolean tryRunCommand(Closure command) {
 String resolveWorkspacePath(String os) {
 	return 'windows' == os ? 'C:\\Users\\Administrator\\jenkins\\workspace\\' : '/home/ubuntu/jenkins/workspace/'
 }
+
+boolean isGitHubRepositoryPublic(String orgName, String repoName) {
```
Jaguar0625: this is already in githubHelper ?

Wayonb: Yeah, this PR was before it got refactored into the gitHubHelper, which should be its final move.

PR #850, opened 2023-11-07
[jenkins] fix: CI and Catapult images can be stale

Wayonb:
## What's the issue?
Alternating builds between arm64 and amd64 are not working correctly due to the ``ARCHITECTURE = helper.determineArchitecture()`` which sets a local variable instead of the env value.
This would cause one of the architecture images not to build and become stale.

Set the ``env.ARCHITECTURE`` when calling the helper.determineArchitecture().

Will test in Jenkins


Jaguar0625 (commented):
do you need to update in other places too, like `catapult-client-release-build` or anywhere else ${ARCHITECTURE} show up?

Wayonb:
> helper.determineArchitecture()

The ``ARCHITECTURE`` is only overridden for the image and daily builds.
The problem is that ``ARCHITECTURE = helper.determineArchitecture()`` was setting a local scoped value of ``ARCHITECTURE`` and not the ``env.ARCHITECTURE`` which sets the global value.

Jaguar0625 (approved):

PR #852, opened 2023-11-08
[jenkins] fix: ignore codecov upload failures for private repo

Wayonb:
Jenkins job for private repo will fail on Codecov uploads which we cannot fix.

## What's the issue?
Codecov has a 250 upload limit for private repo after which uploads will fail.
Codecov also throttles uploads when they are too frequent.

For private repo, do not pass the ``--nonZero`` flag to Codecov which forces a returned code on failure.

Tested in Jenkins.


Jaguar0625 (approved):
👍

PR #857, opened 2023-11-14
[sdk] fix: Update the Javascript and Python SDKs to run on Windows.

Wayonb:
## What's the issue?
1. Windows by default installs Python 3.12 which is not supported by aiohttp
    ``https://github.com/aio-libs/aiohttp/issues/7646``
2. build scripts fail to diff current JSON files with the generated file on Windows due to extra ``cr``.
3. Javascript SDK is failing to build on Windows due to
         a. build using the linker from the git package instead of VS
         b. run of the vector scripts failed since Windows shell does not know BLOCKCHAIN env.  Windows shell requires %BLOCKCHAIN% vs ${BLOCKCHAIN}
         c. import file does match URL format(https://nodejs.org/api/esm.html#file-urls)
     
1. Windows install Python 3.11 which is supported by aiohttp
2. Ignore ``cr`` on the diff of JSON files.
3. Javascript SDK updates
         a. Put the VS tool first in the PATH in the user profile.
         b. run the vector scripts with bash so that the environment variable(BLOCKCHAIN) will get processed.
         c. convert the import file pathToFileURL to URL format

Tested in Jenkins.

### DIFF
--- a/jenkins/docker/windows/javascript.Dockerfile
+++ b/jenkins/docker/windows/javascript.Dockerfile
@@ -12,7 +12,9 @@ RUN Set-ExecutionPolicy -ExecutionPolicy RemoteSigned -Scope CurrentUser; `
 	$command='c:\scoop.ps1 -RunAsAdmin'; `
 	iex $command; `
 	del c:\scoop.ps1; `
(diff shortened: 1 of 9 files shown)


Jaguar0625 (approved):
👍

PR #885, opened 2023-12-29
block_filterd_timestamp

0x070696E65:
Could not filter blocks by timestamp

Blocks can be filtered by timestamp

### DIFF
--- a/client/rest/src/db/CatapultDb.js
+++ b/client/rest/src/db/CatapultDb.js
@@ -204,11 +204,13 @@ class CatapultDb {
 	 * Retrieves filtered and paginated blocks.
 	 * @param {Uint8Array} signerPublicKey Filters by signer public key
 	 * @param {Uint8Array} beneficiaryAddress Filters by beneficiary address
+	 * @param {uint64} fromTimestamp Filters by fromTimestamp
+	 * @param {uint64} toTimestamp Filters by toTimestamp
 	 * @param {object} options Options for ordering and pagination. Can have an `offset`, and must contain the `sortField`, `sortDirection`,
 	 * `pageSize` and `pageNumber`. 'sortField' must be within allowed 'sortingOptions'.
 	 * @returns {Promise.<object>} Blocks page.
 	 */
-	blocks(signerPublicKey, beneficiaryAddress, options) {
+	blocks(signerPublicKey, beneficiaryAddress, fromTimestamp, toTimestamp, options) {
 		const sortingOptions = {
 			id: '_id',
 			height: 'block.height'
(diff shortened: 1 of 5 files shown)


client/rest/src/db/CatapultDb.js
```diff
@@ -204,11 +204,13 @@ class CatapultDb {
 	 * Retrieves filtered and paginated blocks.
 	 * @param {Uint8Array} signerPublicKey Filters by signer public key
 	 * @param {Uint8Array} beneficiaryAddress Filters by beneficiary address
+	 * @param {uint64} fromTimestamp Filters by fromTimestamp
+	 * @param {uint64} toTimestamp Filters by toTimestamp
```
Jaguar0625, line 208: try something a little more descriptive, e.g.

Filters blocks by only including blocks with a timestamp greater than or equal to provided value
Filters blocks by only including blocks with a timestamp less than or equal to provided value

client/rest/test/db/CatapultDb_spec.js
```diff
@@ -401,7 +411,45 @@ describe('catapult db', () => {
...
-				return runTestAndVerifyIds(dbBlocks, db => db.blocks(undefined, account1.address, paginationOptions), [10]);
+				return runTestAndVerifyIds(dbBlocks, db =>
+					db.blocks(undefined, account1.address, undefined, undefined, paginationOptions), [10]);
+			});
+
+			it('fromTimestamp', () => {
+				// Arrange:
+				const dbBlocks = [
+					createBlock(10, 1, account1.publicKey, account1.address, 23456),
+					createBlock(20, 1, account1.publicKey, account1.address, 12345)
```
Jaguar0625, line 422: for this test, would prefer if you add a third block with height > 23456

test is currently testing:
1. == fromTimestamp included
2. < fromTimestamp not included

but is missing
3. > fromTimestamp included

and then make similar changes to other NEW tests

client/rest/test/db/utils/dbTestUtils.js
```diff
@@ -162,7 +162,7 @@ const createDbTransaction = (id, signerPublicKey, recipientAddress, options) =>
 		message: { size: 12, payload: new Binary(test.random.bytes(12)) },
 		mosaics: []
 	};
-	for (let j = 0; 3 < j; ++j)
+	for (let j = 0; 3 > j; ++j)
```
Jaguar0625, line 165: what is purpose of this change? 🤔

client/rest/src/db/CatapultDb.js
0x070696E65: I used your text verbatim 🙏

client/rest/test/db/CatapultDb_spec.js
0x070696E65: I added some tests!

client/rest/test/db/utils/dbTestUtils.js
0x070696E65: Sorry, this is a mistake. It has been restored to its original state.

Jaguar0625 (approved):
Nice job!

Wayonb (approved):

PR #896, opened 2024-01-09
[jenkins] fix: add virtual python environment

Wayonb:
The cpp image does not create a virtual environment for Python.

## What's the issue?
Without a virtual Python environment, there will be package conflicts on installation.

 Create a virtual python environment for the cpp image.

It is tested locally on dev box.


Jaguar0625 (approved):
👍

PR #904, opened 2024-01-24
[sdk/javascript]: remove logic depdendent on constructor.name 

Jaguar0625:
problem: `constructor.name` is used in two places in the sdk, but it is dropped during minification
solution: rewrite those sections to no longer use `constructor.name`


sdk/javascript/src/RuleBasedTransactionFactory.js
```diff
@@ -40,8 +40,11 @@ const typeConverterFactory = (module, customTypeConverter, value) => {
 	if (customTypeConverter && customTypeConverter(value))
 		return customTypeConverter(value);
 
-	if (value instanceof ByteArray)
-		return new module[value.constructor.name](value.bytes);
+	if (value instanceof ByteArray) {
+		/** @type object */
```
Jaguar0625, line 44: since this is in `instanceof` block, it must be a `ByteArray` type with NAME static property

but, TS still detects `value.constructor` as `function` (not `ByteArray`) and disallows assignment because `function` may not have `NAME`.

Wayonb (approved):
👍🏾

Jaguar0625:
merged 4569b2f40204c98d1c7889b52976e7f09231ab3b

PR #906, opened 2024-01-26
[monorepo] fix: update latest test image to use python 3.12

Wayonb:
Tests are currently not run with Python 3.12

## What's the issue?
Python 3.12 is now the default version on some Linux distributions.

Use Python 3.12 for testing in the latest image.
Move to Pylint 3, which supports Python 3.12
Remove Pylint-Quotes since does not support Pylint 3

Will run tests in Jenkins.

### DIFF
--- a/jenkins/docker/ubuntu/javascript.Dockerfile
+++ b/jenkins/docker/ubuntu/javascript.Dockerfile
@@ -17,7 +17,7 @@ RUN apt-get install -y wget gnupg \
 
 # nodejs
 ENV NODE_OPTIONS="--dns-result-order=ipv4first"
-ARG NODEJS_VERSION=18
+ARG NODEJS_VERSION=20
 RUN apt-get install -y ca-certificates curl gnupg \
 	&& mkdir -p /etc/apt/keyrings \
 	&& curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg \
--- a/jenkins/docker/ubuntu/javascript.Dockerfile
+++ b/jenkins/docker/ubuntu/javascript.Dockerfile
@@ -17,11 +17,11 @@ RUN apt-get install -y wget gnupg \
 
 # nodejs
 ENV NODE_OPTIONS="--dns-result-order=ipv4first"
-ARG NODE_MAJOR=18
+ARG NODEJS_VERSION=18
 RUN apt-get install -y ca-certificates curl gnupg \
 	&& mkdir -p /etc/apt/keyrings \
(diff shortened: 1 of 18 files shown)


jenkins/docker/ubuntu/javascript.Dockerfile
```diff
@@ -17,11 +17,11 @@ RUN apt-get install -y wget gnupg \
 
 # nodejs
 ENV NODE_OPTIONS="--dns-result-order=ipv4first"
-ARG NODE_MAJOR=18
+ARG NODEJS_VERSION=18
```
Jaguar0625, line 20: fyi - 18 is in maintenance, maybe we should look into upgrading to active (20) in separate PR

jenkins/docker/windows/javascript.Dockerfile
```diff
@@ -12,9 +12,7 @@ RUN Set-ExecutionPolicy -ExecutionPolicy RemoteSigned -Scope CurrentUser; `
 	$command='c:\scoop.ps1 -RunAsAdmin'; `
 	iex $command; `
 	del c:\scoop.ps1; `
-	scoop install git shellcheck nodejs-lts cygwin rustup; `
-	scoop bucket add versions; `
-	scoop install versions/python311; `
+	scoop install git shellcheck nodejs-lts cygwin rustup python3; `
```
Jaguar0625, line 15: what python version is that? 3.11?

linters/python/.pylintrc
```diff
@@ -276,7 +276,7 @@ missing-member-max-choices=1
 
 # List of additional names supposed to be defined in builtins. Remember that
 # you should avoid to define new builtins when possible.
-additional-builtins=_
+additional-builtins=
```
Jaguar0625, line 279: why this change?

Jaguar0625 (approved):

jenkins/docker/ubuntu/javascript.Dockerfile
Wayonb: Yeah, I will also update the default value.  This will be overwritten by values in the ``baseImages.yaml``

jenkins/docker/windows/javascript.Dockerfile
Wayonb: Scoop will always install the latest Python version on Windows.  Currently, this is Python 3.12.1.

This is an excellent way to find out when our Python SDK is broken with the latest release 😅

linters/python/.pylintrc
Wayonb: This is a mistake, will revert 👍🏾

Jaguar0625 (approved):

PR #911, opened 2024-01-29
[client/rest]: fix failing bootstrapper tests

Jaguar0625:
problem: tests are failing because requests are implicitly being sent with keep-alive
solution: explicitly send requests with `Connection: close` header


Wayonb (approved):
👍🏾

PR #919, opened 2024-01-31
[catapult/client]: upgrade boost to 1.83

Jaguar0625:
- fix duplicate libraries linker warning
- workaround boost versions 1.81-1.83
- increment boost target


Wayonb (approved):
👍🏾

PR #928, opened 2024-02-06
[sdk/javascript]: fix failing test vectors after constructor.name change

Jaguar0625:
problem: NAME is not added to generated model classes, so they can not be used in descriptors
solution: add fallback logic when NAME is not provided

### DIFF
--- a/sdk/javascript/src/RuleBasedTransactionFactory.js
+++ b/sdk/javascript/src/RuleBasedTransactionFactory.js
@@ -43,7 +43,11 @@ const typeConverterFactory = (module, customTypeConverter, value) => {
 	if (value instanceof ByteArray) {
 		/** @type object */
 		const ByteArrayClass = value.constructor;
-		return new module[ByteArrayClass.NAME](value.bytes);
+
+		if (ByteArrayClass.NAME in module)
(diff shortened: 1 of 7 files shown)


sdk/javascript/src/RuleBasedTransactionFactory.js
```diff
@@ -43,7 +43,11 @@ const typeConverterFactory = (module, customTypeConverter, value) => {
 	if (value instanceof ByteArray) {
 		/** @type object */
 		const ByteArrayClass = value.constructor;
-		return new module[ByteArrayClass.NAME](value.bytes);
+
+		if (ByteArrayClass.NAME in module)
+			return new module[ByteArrayClass.NAME](value.bytes);
+
+		return new ByteArrayClass(value.bytes);
```
Wayonb, line 50: Is there a case where the ``NAME`` is not set in the class?

Wayonb (approved):

sdk/javascript/src/RuleBasedTransactionFactory.js
Jaguar0625: yes - it is only set in the public types, but not the generated model types

PR #929, opened 2024-02-07
[jenkins] fix: always run the vector tests

Wayonb:
problem: currently, vector tests are not run by default.
solution: always run vector tests


Jaguar0625 (approved):
good for now, but we should probably add some concept of integration / dev-only stage for the docs tests, as it would be useful to run the simple examples in all cases too

PR #951, opened 2024-03-02
[monorepo] fix: updated the CodeQL action

Wayonb:
problem: CodeQL is not running with the latest configuration
solution: update CodeQL workflow to the latest

### DIFF
--- a/.github/workflows/codeql-analysis.yaml
+++ b/.github/workflows/codeql-analysis.yaml
@@ -10,10 +10,10 @@ name: Code Scanning - Action
     - cron: '44 11 * * 3'
 
 jobs:
-  CodeQL-Build:
-    # CodeQL runs on ubuntu-latest, windows-latest, and macos-latest
-    runs-on: ubuntu-latest
-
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yaml
```diff
@@ -10,10 +10,10 @@ name: Code Scanning - Action
     - cron: '44 11 * * 3'
 
 jobs:
-  CodeQL-Build:
-    # CodeQL runs on ubuntu-latest, windows-latest, and macos-latest
-    runs-on: ubuntu-latest
-
+  analyze:
+    name: Analyze
+    runs-on: ${{ (matrix.language == 'swift' && 'macos-latest') || 'ubuntu-latest' }}
```
Jaguar0625, line 15: since we don't have swift, i guess this is just template from somewhere?

Wayonb: yes, generated by GitHub we only use Ubuntu, so it can be removed 👍🏾

PR #952, opened 2024-03-02
[jenkins] feat: add support for build and test on Fedora

Wayonb:
problem: there is no support for building and testing on Fedora.
solution: add Fedora CI images for testing


Jaguar0625 (approved):
👍

PR #956, opened 2024-03-06
[sdk/javascript]: use globalThis instead of global

Jaguar0625:
problem: sdk can't run in browser where global is not defined
solution: replace global with globalThis (resolves to window)


Wayonb (approved):
👍🏾

PR #957, opened 2024-03-08
[jenkins] fix: install rust in Javascript image

Wayonb:
problem: Javascript image does not have rust installed
solution: run rustup during Javascript image creation to install the latest rust


Jaguar0625 (approved):
👍

PR #958, opened 2024-03-08
[sdk/javascript]: improve codegen to significantly reduce size of generated models

Jaguar0625:
problem: generated code is very duplicative
solution: add concept of base class to generator to reduce duplication

### DIFF
--- a/sdk/javascript/generator/AbstractTypeFormatter.py
+++ b/sdk/javascript/generator/AbstractTypeFormatter.py
@@ -16,6 +16,10 @@ class AbstractTypeFormatter(ABC):
 	def typename(self):
 		raise NotImplementedError('need to override method')
 
+	@property
+	def is_type_abstract(self):
+		return False
+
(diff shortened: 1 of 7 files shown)


sdk/javascript/generator/AbstractTypeFormatter.py
```diff
@@ -16,6 +16,10 @@ class AbstractTypeFormatter(ABC):
 	def typename(self):
 		raise NotImplementedError('need to override method')
 
+	@property
+	def is_abstract(self):
+		return False
```
Wayonb, line 21: At first, I thought this referred to the formatted class....maybe ``is_abstract_type/class``?

Wayonb (approved):
Nice use of inheritance 👍🏾

Jaguar0625:
@Wayonb - let me know if you have any additional comments after the rename 🙏

Wayonb (approved):
👍🏾

PR #959, opened 2024-03-08
[sdk/python] fix: ByteString is deprecated in Python 3.9

Wayonb:
problem: Pylint 3.1 fails in Python 3.12 due to ByteString is deprecated
         https://docs.python.org/3.12/library/typing.html#typing.ByteString
solution: Replace the ByteString with ``bytes | bytearray | memoryview``


Jaguar0625 (approved):
👍

PR #964, opened 2024-03-12
[sdk/python]: improve codegen to significantly reduce size of generated models

Jaguar0625:
problem: generated code is very duplicative
    solution: add concept of base class to generator to reduce duplication

### DIFF
--- a/sdk/python/symbolchain/sc/__init__.py
+++ b/sdk/python/symbolchain/sc/__init__.py
@@ -372,18 +372,19 @@ class Mosaic:
 	@classmethod
 	def deserialize(cls, payload: bytes | bytearray | memoryview) -> Mosaic:
 		buffer = memoryview(payload)
+		instance = Mosaic()
 		mosaic_id = MosaicId.deserialize(buffer)
 		buffer = buffer[mosaic_id.size:]
 		amount = Amount.deserialize(buffer)
 		buffer = buffer[amount.size:]
 
-		instance = Mosaic()
+		# pylint: disable=protected-access
 		instance._mosaic_id = mosaic_id
 		instance._amount = amount
 		return instance
 
 	def serialize(self) -> bytes:
-		buffer = bytes()
+		buffer = bytearray()
 		buffer += self._mosaic_id.serialize()
 		buffer += self._amount.serialize()
 		return buffer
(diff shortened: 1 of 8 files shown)


sdk/python/symbolchain/sc/__init__.py
```diff
@@ -635,12 +637,11 @@ def size(self) -> int:
 		return size
 
 	@classmethod
-	def deserialize(cls, payload: bytes | bytearray | memoryview) -> Transaction:
-		buffer = memoryview(payload)
+	def _deserialize(cls, buffer: memoryview, instance) -> (int, int):
+		size_ = len(buffer)
 		size_ = int.from_bytes(buffer[:4], byteorder='little', signed=False)
 		buffer = buffer[4:]
 		buffer = buffer[:size_ - 4]
```
Wayonb, line 644: Can this be rewritten as ``buffer = buffer[4:size_ - 4]``

sdk/python/symbolchain/sc/__init__.py
```diff
@@ -662,18 +663,22 @@ def deserialize(cls, payload: bytes | bytearray | memoryview) -> Transaction:
...
+		# pylint: disable=protected-access
 		instance._signature = signature
 		instance._signer_public_key = signer_public_key
 		instance._version = version
 		instance._network = network
 		instance._type_ = type_
 		instance._fee = fee
 		instance._deadline = deadline
-		return instance
+		return (size_ - len(buffer), size_)
```
Wayonb, line 674: Should ``len(buffer)`` be zero at end? 🤔

Wayonb (changes requested):

sdk/python/symbolchain/sc/__init__.py
```diff
@@ -635,12 +637,11 @@ def size(self) -> int:
 		return size
 
 	@classmethod
-	def deserialize(cls, payload: bytes | bytearray | memoryview) -> Transaction:
-		buffer = memoryview(payload)
+	def _deserialize(cls, buffer: memoryview, instance) -> (int, int):
+		size_ = len(buffer)
```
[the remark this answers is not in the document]
Jaguar0625: this was added for NEM, where structs don't have intrinsic side fields. i will add a better filter.

sdk/python/symbolchain/sc/__init__.py
Jaguar0625: i can tweak the generator a bit 👍

sdk/python/symbolchain/sc/__init__.py
Jaguar0625: no - buffer is entire *remaining* buffer. so, we need to effectively shift (and mask) it like in the JS implementation:
```
		view.shiftRight(4);
		view.shrink(size - 4);
```
remember, this is base class (Transaction) and size includes both base + derived fields. at this point, buffer should contain only derived part. `size_ - len(buffer) ` let's us skip to derived part. `size_` masks off the next entity(ies), if present [important for aggregates, which have cosignatures expanding to fill remaining space].

Wayonb (approved):
👍🏾

PR #965, opened 2024-03-13
[sdk/javascript]: split single entry point into default, nem and symbol

Jaguar0625:
problem: sdk does not support named imports
    solution: restructure and split up module entry points
    
    allow usage like:
    ```js
    import { PrivateKey } from 'symbol-sdk';
    import { SymbolFacade } from 'symbol-sdk/symbol';
    ```


Wayonb:
Changes look good but examples need to be updated 👍🏾

Wayonb (approved):
👍🏾

PR #968, opened 2024-03-15
[sdk/javascript]: run ts compiler over examples

Jaguar0625:
problem: examples are not TS compliant
solution:
          - add static getter that can be used instead of constructor in TS
          - update examples


Wayonb (approved):
👍🏾

PR #981, opened 2024-03-29
[sdk]: remove duplicate receipt test vector

Jaguar0625:
problem: vectors contain duplicate MosaicRentalFeeReceipt entry
solution: remove it


Wayonb (approved):
👍🏾

PR #987, opened 2024-04-01
[jenkins] fix: the latest docker requires subnet for static ip

Wayonb:
problem: the latest docker version requires a subnetwork when a static IP address is used.
solution: specify the network that has the static IP address


Jaguar0625 (approved):
👍

PR #988, opened 2024-04-01
[jenkins] fix: escape character from backslach to backtick for Windows Dockerfile

Wayonb:
problem: escape character for Windows Dockerfile should be backtick and not backslash
solution: update the escape character to the backtick


Jaguar0625 (approved):
👍

PR #989, opened 2024-04-01
[jenkins] fix: nightly and weekly jobs passes the branch name ci job

Wayonb:
problem: nightly and weekly jobs pass the branch name ci job when it is not needed.
         The CI job ignores the branch name, which leads to many warnings in the Jenkins logs.
solution: do not pass the branch name to the CI jobs.


Jaguar0625 (approved):
👍

PR #990, opened 2024-04-05
[client/catapult]: add configuration for managing RockDB log files

Jaguar0625:
problem: RocksDB logs files use a lot of disk space
solution: allow node operator to limit log resource use


Wayonb (approved):
👍🏾

PR #991, opened 2024-04-05
[jenkins] fix: remove hardcode repo in the Jenkinsfile

Wayonb:
problem: catapult project has repo hardcoded, which is not needed
solution: remove the hardcoded repo sync; this is managed in the Jenkins job


Jaguar0625 (approved):
👍

PR #992, opened 2024-04-05
[client/catapult] fix: update _FORTIFY_SOURCE=3 for gcc on Ubuntu 24.04

Wayonb:
problem: GCC on Ubuntu 24.04 updated the default _FORTIFY_SOURCE to 3
         This causes refined warning when building catapult release
solution: update the _FORTIFY_SOURCE=3


Jaguar0625 (approved):
👍

PR #1001, opened 2024-04-09
[jenkins] fix: update helper name

Wayonb:
problem: git helper function to their own file, but some ref was not updated
solution: update the git function to the correct class.


Jaguar0625 (approved):
👍

PR #1004, opened 2024-04-09
[sdk/javascript] fix: set the git repo and homepage URL

Wayonb:
problem: git repo and homepage URL are not set in the package.json
solution: set the git repo and homepage URL in package.json


Jaguar0625 (approved):
👍

PR #1005, opened 2024-04-10
[sdk/javascript]: '| 0' truncation trick doesn't work with large integers

Jaguar0625:
problem: '| 0' is a bitwise operation that only works with signed 32-bit integers
          since DateTime timestamps are larger, it is inappropriate for these numbers
solution: use Math.trunc instead, which properly handles larger numbers


AnthonyLaw (approved):
👍🏼

Wayonb (approved):
👍🏾

PR #1007, opened 2024-04-11
[jenkins] fix: upgrade to clang 18

Wayonb:
problem: tsan run is failing 'unexpected memory mapping'
         this is due to a change in the latest Linux kernel.
solution: upgrade to the latest clang 18, which should have the fix
``https://github.com/google/sanitizers/issues/1716#issuecomment-2010399341``


Jaguar0625 (approved):
👍

Jaguar0625 (approved):
👍

PR #1008, opened 2024-04-12
[jenkins] fix: clang 18 remove the 'avx' options from ARM

Wayonb:
problem: libzmq is failing to build on ARM due to 'avx' was removed in Clang 18
solution: only add the 'avx' options on intel architecture


Jaguar0625 (approved):
👍

PR #1019, opened 2024-04-22
[sdk/javascript]: fix package exports

Jaguar0625:
problem: subpath exports are not working correctly with certain JS/TS environments
    solution: for max compatibility, add 'import' and move 'default' last
              add README for min TS requirements


Wayonb (approved):
👍🏾

PR #1020, opened 2024-04-22
[sdk/javascript] fix: skip the node_modules folder for Python linting

Wayonb:
problem: Python lint script searches the node_modules folder during linting.
         the issue is some node packages contain Python files which leads to failure.
solution: skip the node_modules folder for the Python lint script.


Jaguar0625 (approved):
👍

PR #1023, opened 2024-04-24
[jenkins] fix: update docker images

Wayonb:
problem: docker images are not using the latest OS
solution: upgrade the base images to the latest
          add zip and AWS CLI to the JS and Python images


Jaguar0625 (approved):
👍

Jaguar0625 (approved):
👍

PR #1024, opened 2024-04-25
[jenkins] fix: linting failure due to rmtree onerror deprecation

Wayonb:
problem: in Python 3.12, shutil.rmtree onerror parameter is deprecated
solution: in Python 3.12 use onexc parameter instead of onerror

### DIFF
--- a/jenkins/catapult/environment.py
+++ b/jenkins/catapult/environment.py
@@ -4,11 +4,15 @@ import sys
 from pathlib import Path
 
 
-def rm_failure_handler(func, path, excinfo):
+def rm_onerror_handler(func, path, excinfo):
+	rm_onexc_handler(func, path, excinfo[1])
+
+
+def rm_onexc_handler(func, path, exception):
 	del func
 	del path
-	if excinfo[0] != FileNotFoundError:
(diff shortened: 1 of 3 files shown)


jenkins/catapult/environment.py
```diff
@@ -71,7 +71,8 @@ def rmtree(self, path):
 		if self.dry_run:
 			return
 
-		shutil.rmtree(path, onerror=rm_failure_handler)
+		kwargs = {'onexc' if sys.version_info >= (3, 12) else 'onerror': rm_failure_handler}
+		shutil.rmtree(path, **kwargs)
```
Jaguar0625, line 75: i believe callback parameters changed, so callback needs updating too
https://docs.python.org/3/library/shutil.html#shutil.rmtree

Jaguar0625 (changes requested):

jenkins/catapult/environment.py
Wayonb: yea, it seems the third parameter has change

Jaguar0625 (approved):
👍

PR #1025, opened 2024-04-25
[jenkins] fix: upgrade catapult os and mongodb version

Wayonb:
problem: catapult is not using the latest Fedora, Ubuntu and MongoDB versions
solution: update catapult docker images to use the latest.


Jaguar0625 (approved):
👍

PR #1026, opened 2024-04-26
[sdk/javascript]: mocha runner is being used but jest types are being used

Jaguar0625:
[sdk/javascript]: mocha runner is being used but jest types are being used
    
     problem: mismatch between test runner and types
    solution: replace jest types with mocha types

    [sdk/javascript]: refresh package-lock.json
    
     problem: package-lock.json is stale
    solution: regenerate it


Wayonb (approved):
👍🏾

PR #1027, opened 2024-04-26
[sdk/javascript]: allow BaseValue to coerce in bounds inputs into correct types

Jaguar0625:
problem: NetworkTimestamp.timestamp is BigInt but models.Timestamp (NEM) is Number
          latter cannot be created around NetworkTimestamp.timestamp due to type check
          and mismatch (BigInt != Number)
solution: add logic to BaseValue to automatically coerce input of Number|BigInt
          to desired underlying type (assuming it is in bounds of target)


Wayonb (approved):
👍🏾

PR #1031, opened 2024-05-01
[sdk/javascript]: improve typescript support by generating type annotated descriptors

Jaguar0625:
these can be used to generate descriptor maps passed to create functions in a typesafe way with intellisense support.

this is part one of a multipart change.
a followup change will integrate these descriptors with the facades.

 problem: sdk does not support setting transaction parameters all at once in a typesafe way
solution: generate a suite of descriptors that accept strongly typed arguments


sdk/javascript/src/nem/models_ts.js
```diff
@@ -0,0 +1,632 @@
...
+	}
+
+	/**
+	 * Builds a representation of this descriptor that can be passed to a factory function.
+	 * @returns {object} Descriptor that can be passed to a factory function.
+	 */
+	toMap() {
+		return this.rawDescriptor;
+	}
+}
```
Wayonb, line 33: missing a space?

sdk/javascript/src/symbol/models_ts.js
```diff
@@ -0,0 +1,1039 @@
...
+	 * @param {models.Cosignature[]|undefined} cosignatures Cosignatures data.
+	Fills up remaining body space after transactions.
+	 */
+	constructor(transactionsHash, transactions = undefined, cosignatures = undefined) {
+		this.rawDescriptor = {
+			type: 'aggregate_complete_transaction_v2',
+			transactionsHash
+		};
+
+		if (transactions)
```
Wayonb, line 232: What is the use case when there is no transaction in the aggregate tx?

sdk/javascript/src/symbol/models_ts.js
```diff
@@ -0,0 +1,1039 @@
...
+export class NamespaceRegistrationTransactionV1Descriptor {
+	/**
+	 * Creates a descriptor for NamespaceRegistrationTransactionV1.
+	 * @param {models.NamespaceId} id Namespace identifier.
+	 * @param {models.NamespaceRegistrationType} registrationType Namespace registration type.
+	 * @param {models.BlockDuration|undefined} duration Number of confirmed blocks you would like to rent the namespace for. Required for root namespaces.
+	 * @param {models.NamespaceId|undefined} parentId Parent namespace identifier. Required for sub-namespaces.
+	 * @param {Uint8Array|string|undefined} name Namespace name.
+	 */
+	constructor(id, registrationType, duration = undefined, parentId = undefined, name = undefined) {
```
Wayonb, line 805: Should there be a constructor for creating the root vs sub namespace? This would ensure that the correct values are passed in.

Wayonb (approved):
👍🏾

sdk/javascript/src/nem/models_ts.js
Jaguar0625: where? 🤔

sdk/javascript/src/symbol/models_ts.js
Jaguar0625: there is none, transaction will get rejected with `Failure_Aggregate_No_Transactions` by network 

the generator is merely flagging all arrays and conditional arguments as optional

sdk/javascript/src/symbol/models_ts.js
Jaguar0625: good suggestion, i will look into this

sdk/javascript/src/nem/models_ts.js
Wayonb: Between the generated classes.  So between the `}` and the start of the comment for the next class.

```
	toMap() {
		return this.rawDescriptor;
	}
}
/**
 * Type safe descriptor used to generate a descriptor map for NamespaceIdDescriptor.
 *
 * binary layout for a namespace id
 */
export class NamespaceIdDescriptor {
```

should be 

```
	toMap() {
		return this.rawDescriptor;
	}
}

/**
 * Type safe descriptor used to generate a descriptor map for NamespaceIdDescriptor.
 *
 * binary layout for a namespace id
 */
export class NamespaceIdDescriptor {
```

sdk/javascript/src/symbol/models_ts.js
Jaguar0625: since JS doesn't support overloaded ctors, we could potentially add factory functions.

however, that probably doesn't make sense because it would be a different paradigm just for that one class, and the main goal of this is to simplify and streamline usage.

PR #1032, opened 2024-05-01
[sdk/javascript]: extend facades to simplify usage of typed descriptors

Jaguar0625:
problem: using typed descriptors requires some extra knowledge
              in order to create valid transactions
    solution: simplify usage by adding helper functions to facades


Wayonb (approved):
👍🏾

PR #1033, opened 2024-05-01
[sdk]:  add deserialize function to TransactionFactory in sdk

Jaguar0625:
[sdk/python]: add deserialize function to TransactionFactory in sdk
    
     problem: deserialize is difficult to discover because it is only on model factory
    solution: expose deserialize on sdk TransactionFactory too

    [sdk/javascript]: add deserialize function to TransactionFactory in sdk
    
     problem: deserialize is difficult to discover because it is only on model factory
    solution: expose deserialize on sdk TransactionFactory too


Wayonb (approved):
👍🏾

PR #1039, opened 2024-05-08
[sdk/javascript]: add cosignatureCount optional parameter to SymbolFacade createTransactionFromTypedDescriptor

Jaguar0625:
problem: when calculating fee for aggregate transactions the total number of cosignatures needs to be taken into account
          currently, this is not easy to configure via createTransactionFromTypedDescriptor
solution: add an optional cosignatureCount parameter


Wayonb (approved):
👍🏾

PR #1040, opened 2024-05-10
[jenkins] fix: catapult project fails on initial create

Wayonb:
problem: catapult jobs will get updated when build configuration changes.
         this causes catapult project jobs to fail due to defaults values are not set.
solution: set the default values for each parameter for the catapult project.


Jaguar0625 (approved):
👍

PR #1048, opened 2024-05-13
[sdk]: add createAccount to facades

Jaguar0625:
[sdk/python]: add create_account to facades
    
     problem: deriving an address from a private key is a multistep process
    solution: add facade function that will derive public key and address from private key

    [sdk/javascript]: add createAccount to facades
    
     problem: deriving an address from a private key is a multistep process
    solution: add facade function that will derive public key and address from private key


AnthonyLaw (approved):
👍🏼

Wayonb (approved):
👍🏾

PR #1049, opened 2024-05-13
[jenkins] fix: add coverage to JavaScript image

Wayonb:
problem: some Python projects use Javascript image since npm is required
solution: add coverage tool to Javascript image to allow Python projects to create code coverage


Jaguar0625 (approved):
👍

PR #1050, opened 2024-05-16
[monorepo]: add .python-version to gitignore

Jaguar0625:
problem: .python-version can be used to map python environments to directory trees but these shouldn't be committed
solution: add .python-version to .gitignore


Wayonb (approved):
👍🏾

PR #1052, opened 2024-05-17
[sdk]: add public account and behaviors to account classes

Jaguar0625:
problem: sdk user needs to interact with facade too much
    solution: expand account class and add public account class


Wayonb (approved):
👍🏾

PR #1064, opened 2024-05-20
[sdk/javascript]: export PublicAccount and Account from sdk

Jaguar0625:
problem: account types are not exported but TS needes them for typing
solution: export account types


Wayonb (approved):
👍🏾

AnthonyLaw (approved):

PR #1066, opened 2024-05-21
[sdk]: add message-encoder to account class 

Jaguar0625:
[sdk/python]: add message_encoder to account classes
    
     problem: user needs to know about MessageEncoder to encrypt/encode messages
    solution: add a helper function to accounts to create it to improve discoverability

    [sdk/javascript]: add messageEncoder to account classes
    
     problem: user needs to know about MessageEncoder to encrypt/encode messages
    solution: add a helper function to accounts to create it to improve discoverability


Wayonb (approved):
👍🏾

PR #1067, opened 2024-05-21
[sdk]: add metadata-generate-key

Jaguar0625:
[sdk/python]: add metadata_generate_key
    
     problem: there is no built in way to convert a string to a metadata key
    solution: add metadata_generate_key and make it compatible with SDK V2 implementation

    [sdk/javascript]: add metadataGenerateKey
    
     problem: there is no built in way to convert a string to a metadata key
    solution: add metadataGenerateKey and make it compatible with SDK V2 implementation


Wayonb (approved):
👍🏾

PR #1068, opened 2024-05-21
[sdk]: add additional Symbol address utilities

Jaguar0625:
[sdk/python]: add additional Symbol address utilities
    
     problem: some useful functions from the SDK V2 are missing
    solution: add missing functions:
              - create address alias from namespace id
              - extract namespace id from address alias
              - create address from decoded address hex string (REST format)

    [sdk/javascript]: add additional Symbol address utilities
    
     problem: some useful functions from the SDK V2 are missing
    solution: add missing functions:
              - create address alias from namespace id
              - extract namespace id from address alias
              - create address from decoded address hex string (REST format)


Wayonb (approved):
👍🏾

PR #1070, opened 2024-05-22
[sdk]: add generation for class level documentation

Jaguar0625:
🚨 there is new `./scripts/generate-docs.sh` in each SDK. they will produce documents with index file in `docs/build/html/index.html`. we should probably publish these somewhere...
    
     [sdk/javascript]: add typedoc documentation support
    
     problem: there is no class level documentation
    solution: use typedoc to generate basic TS-style documentationx

     [sdk/python]: add sphinx documentation support
    
     problem: there is no class level documentation
    solution: use sphinx to generate basic readthedocs-style documentation


Wayonb (approved):
👍🏾

PR #1071, opened 2024-05-23
[jenkins] fix: curl command failing on Windows

Wayonb:
problem: the curl command is failing on Windows due to multiple issues
       1. Windows cmd shell does not use backslash for multi-line command
       2. in the curl write-out format the % was getting removed
solution: 1. use StringBuilder to format the curl command
              2. remove the write-out option curl will return failure to shell

### DIFF
--- a/jenkins/shared-library/vars/githubHelper.groovy
+++ b/jenkins/shared-library/vars/githubHelper.groovy
@@ -12,6 +12,7 @@ boolean isGitHubRepositoryPublic(String orgName, String repoName) {
 	}
 }
 
+// groovylint-disable-next-line FactoryMethodName
 String buildCurlCommand(String token, String url, String data = null, Boolean post = false) {
 	return [
 		'curl -L',
--- a/jenkins/shared-library/vars/githubHelper.groovy
+++ b/jenkins/shared-library/vars/githubHelper.groovy
@@ -13,13 +13,9 @@ boolean isGitHubRepositoryPublic(String orgName, String repoName) {
 }
 
 Object getRepositoryInfo(String token, String ownerName, String repositoryName) {
-	final String getRepoCommand = """
-		curl -L \\
-		  -H "Accept: application/vnd.github+json" \\
-		  -H "Authorization: Bearer ${token}" \\
-		  -H "X-GitHub-Api-Version: 2022-11-28" \\
-		  https://api.github.com/repos/${ownerName}/${repositoryName} \\
-	"""
+	final String getRepoCommand = new StringBuilder().append('curl -L -H "Accept: application/vnd.github+json"')
+		.append(""" -H "Authorization: Bearer ${token}" """)
+		.append("""-H "X-GitHub-Api-Version: 2022-11-28"  https://api.github.com/repos/${ownerName}/${repositoryName}""")
 
 	final String getRepoResponse = executeGithubApiRequest(getRepoCommand)
 	return yamlHelper.readYamlFromText(getRepoResponse)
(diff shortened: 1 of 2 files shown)


jenkins/shared-library/vars/githubHelper.groovy
```diff
@@ -29,12 +25,8 @@ String executeGithubApiRequest(String command) {
 	String results = ''
 	helper.withTempDir {
 		String file = "./${System.currentTimeMillis()}"
-		final String response = runScript("${command} -s -w '%{http_code}\\n' -o ${file}", true).trim()
+		runScript("${command} -s -o ${file}")
```
Jaguar0625, line 28: there is exception on failure now?

jenkins/shared-library/vars/githubHelper.groovy
```diff
@@ -13,13 +13,9 @@ boolean isGitHubRepositoryPublic(String orgName, String repoName) {
...
-	final String getRepoCommand = """
-		curl -L \\
-		  -H "Accept: application/vnd.github+json" \\
-		  -H "Authorization: Bearer ${token}" \\
-		  -H "X-GitHub-Api-Version: 2022-11-28" \\
-		  https://api.github.com/repos/${ownerName}/${repositoryName} \\
-	"""
+	final String getRepoCommand = new StringBuilder().append('curl -L -H "Accept: application/vnd.github+json"')
+		.append(""" -H "Authorization: Bearer ${token}" """)
+		.append("""-H "X-GitHub-Api-Version: 2022-11-28"  https://api.github.com/repos/${ownerName}/${repositoryName}""")
```
Jaguar0625, line 18: generally, i find old version more readable due to multiple args per line in new version. consider either multiple appends or something like (excuse the poor formatting):

```groovy
[
		'curl -L',
		  '-H "Accept: application/vnd.github+json"'
		  "-H "Authorization: Bearer ${token}"
		  '-H "X-GitHub-Api-Version: 2022-11-28" '
		 "https://api.github.com/repos/${ownerName}/${repositoryName}"
].join(' ')
```

jenkins/shared-library/vars/githubHelper.groovy
```diff
@@ -68,15 +55,11 @@ Object createPullRequest(
...
-			-X POST \\
-			-H "Accept: application/vnd.github+json" \\
-			-H "Authorization: Bearer ${token}" \\
-			-H "X-GitHub-Api-Version: 2022-11-28" \\
-			https://api.github.com/repos/${ownerName}/${repositoryName}/pulls/${pullRequestNumber}/requested_reviewers \\
-			-d '{"reviewers": ["${reviewers.join('", "')}"]}'
-	"""
+	final String reviewersCommand = new StringBuilder().append('curl -L -X POST -H "Accept: application/vnd.github+json" ')
+		.append("""-H "Authorization: Bearer ${token}" """)
+		.append('-H "X-GitHub-Api-Version: 2022-11-28" ')
```
Jaguar0625, line 60: i'd also consider refactoring this since these headers seem to be same for all github requests (only variable is method GET/POST, which can be argument)

jenkins/shared-library/vars/githubHelper.groovy
Wayonb: There should  be an exit code but still need to verify

jenkins/shared-library/vars/githubHelper.groovy
```diff
@@ -12,15 +12,21 @@ boolean isGitHubRepositoryPublic(String orgName, String repoName) {
...
 
-Object getRepositoryInfo(String token, String ownerName, String repositoryName) {
-	final String getRepoCommand = """
-		curl -L \\
-		  -H "Accept: application/vnd.github+json" \\
-		  -H "Authorization: Bearer ${token}" \\
-		  -H "X-GitHub-Api-Version: 2022-11-28" \\
-		  https://api.github.com/repos/${ownerName}/${repositoryName} \\
-	"""
+String getCurlCommand(String token, String url, String data = null, Boolean post = false) {
```
Jaguar0625, line 15: maybe call it build instead of get ?

Jaguar0625 (approved):
👍

PR #1077, opened 2024-05-28
[jenkins] fix: GitHub commands only requires the token

Wayonb:
problem: private repo failing to clone repo due to the GitHub token does not have permissions
                this happens when there are a lot of Jenkins jobs that reuse worker nodes but not
                in the same org(esp during the nightly builds)
solution: limit the GitHub token to the org since the GitHub token is org-specific.
               also, only configure Git for push/PR.

### DIFF
--- a/jenkins/shared-library/vars/githubHelper.groovy
+++ b/jenkins/shared-library/vars/githubHelper.groovy
@@ -2,7 +2,7 @@ import groovy.json.JsonOutput
 
 boolean isGitHubRepositoryPublic(String orgName, String repoName) {
 	try {
-		executeGitAuthenticatedCommand {
+		withGitHubToken {
 			final Object repo = getRepositoryInfo("${GITHUB_TOKEN}", orgName, repoName)
 			return repo.name == repoName && repo.visibility == 'public'
 		}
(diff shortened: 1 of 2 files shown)


jenkins/shared-library/vars/githubHelper.groovy
```diff
@@ -103,11 +103,19 @@ void configureGitHub() {
...
 			passwordVariable: 'GITHUB_TOKEN')]) {
-		final String replaceUrl = "https://${GITHUB_USER}:${GITHUB_TOKEN}@github.com/.insteadOf https://github.com/"
+		closure()
+	}
+}
+
+void executeGitAuthenticatedCommand(Closure command) {
+	withGitHubToken {
+		final String ownerName = helper.resolveOrganizationName()
+		final String replaceUrl = 'https://$GITHUB_USER:$GITHUB_TOKEN@github.com/' +
```
Jaguar0625, line 117: did you intentionally change from ${GITHUB_USER} to $GITHUB_USER ?

Wayonb: Yeah, it's to ensure that Jenkins does not try to inject the values before they are executed by Bash.

Jaguar0625 (approved):
👍

PR #1080, opened 2024-05-31
[sdk/javascript] fix: missed updating version in the package-lock.json

Wayonb:
problem: missed updating the package version in the lock file
solution: update the package version in the lock file


Jaguar0625 (approved):
👍

PR #1097, opened 2024-06-10
[client/rest]: restructure project to use ES6 instead of CJS modules

Jaguar0625:
problem: REST is using CJS modules
    solution: migrate REST to ES6 modules

first commit is most of rote changes
second commit fixes breakages
third commit cleans up exports


Wayonb (approved):
👍🏾

PR #1110, opened 2024-06-17
    [client/rest]: replace custom serialization with SDK serialization

Jaguar0625:
NOTE: cannot merge directly into dev until SDK is updated with all changes 

    [client/rest]: replace custom serialization with SDK serialization
    
     problem: there is bespoke serialization logic for all transactions
    solution: replace serialization logic with JS sdk

    [sdk/javascript]: add toJson to models
    
     problem: there is no easy way to get a JSON-safe representation of a model
    solution: add toJson to generated model classes


Wayonb (approved):
Looks good👍🏾

Once the JS sdk is checked-in we can have an internal build so the rest tests will pass.

PR #1112, opened 2024-06-18
[client/rest]: reuse utils that are implemented in the js sdk

Jaguar0625:
problem: rest has own implementations of a handful of common utilities
    solution: use sdk implementations where available


AnthonyLaw (approved):

Wayonb (approved):
👍🏾

PR #1113, opened 2024-06-19
[client/rest]: replace custom model types with sdk types as appropriate

Jaguar0625:
[client/rest]: convert aliasType to sdk-like NamespaceAliasType enum
    
     problem: namespace/aliasType is an enumeration using old conventions
    solution: replace it with an sdk-like enum (NamespaceAliasType)

    [client/rest]: replace encodeNamespace with sdk implementation
    
     problem: rest has custom implementation of Address.fromNamespaceId
    solution: use Address.fromNamespaceId from sdk

    [client/rest]: remove restriction.js
    
     problem: restriction.js is never used
    solution: remove restriction.js

    [client/rest]: use TransactionType from sdk instead of maintaining separate list
    
     problem: rest has a separate list of entity/transaction types
    solution: use TransactionType enum from sdk instead


Wayonb (approved):
👍🏾

PR #1114, opened 2024-06-19
[client/rest]: convert bootstrap tests to return Promise

Jaguar0625:
problem: async bootstrap tests are explicitly calling done
solution: return Promise instead


Wayonb (approved):
👍🏾

PR #1115, opened 2024-06-21
[sdk/python]: add to_json to models

Jaguar0625:
problem: there is no easy way to get a JSON-safe representation of a model
solution: add to_json to generated model classes


Wayonb (approved):
👍🏾

PR #1119, opened 2024-06-26
[client/rest]: add routeUtils.sendJson

Jaguar0625:
problem: there is no way to easily send a JSON object that bypasses the formatting subsystem
solution: add routeUtils.sendJson


Wayonb (approved):
👍🏾

PR #1120, opened 2024-06-26
[client/rest]: add route-only plugins

Jaguar0625:
problem: architecture assumes a single set of sdk (transaction) plugins and route plugins
solution: add separate 'routeExtensions' configuration
          in order to allow route plugins to be a superset of sdk plugins
          add placeholder for route-only plugin (rosetta)

### DIFF
--- a/client/rest/src/plugins/routeSystem.js
+++ b/client/rest/src/plugins/routeSystem.js
@@ -34,6 +34,7 @@ import { NetworkLocator } from 'symbol-sdk';
 import { Network } from 'symbol-sdk/symbol';
 
 const plugins = {
+	// transactions
 	accountLink: empty,
 	aggregate,
 	lockHash,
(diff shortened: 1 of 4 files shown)


client/rest/src/plugins/routeSystem.js
```diff
@@ -44,7 +45,10 @@ const plugins = {
 	namespace,
 	receipts,
 	restrictions,
-	transfer: empty
+	transfer: empty,
+
+	// other
+	rosetta: empty
```
Wayonb, line 51: Isn't the DB undefined in the empty plugin? Rosetta routes will need access to the db.

Wayonb (changes requested):

client/rest/src/plugins/routeSystem.js
Jaguar0625: yes, this will be changed in `rest/dev`

Wayonb (approved):
👍🏾

PR #1121, opened 2024-06-27
[rest/client]: move simple senders out of route utils and support statusCode override

Jaguar0625:
problem: simple send helpers don't use formatters, so don't need to specify type
          HTTP error codes can only be set via Restify errors
solution: move simple send helpers to new file
          allow statusCode to be specified


Wayonb (approved):
👍🏾

PR #1123, opened 2024-06-30
[client/rest]: replace unused addressToPublicKey with lookupPublicKeys

Jaguar0625:
problem: addressToPublicKey only supports a single address and is not used
solution: replace it with lookupPublicKeys and support multiple addresses


Wayonb (approved):
👍🏾

PR #1132, opened 2024-07-01
[client/rest]: move MockServer out of routeTestUtils into own file

Jaguar0625:
problem: MockServer is subobject of routeTestUtils
solution: move to own file and export as default


Wayonb (approved):
👍🏾

PR #1133, opened 2024-07-01
[client/rest]: add construction/preprocess and construction/metadata

Jaguar0625:
problem: add additional rosetta routes and refactor to allow async routes
solution: add two more construction routes


client/rest/src/plugins/rosetta/constructionRoutes.js
```diff
@@ -31,6 +35,8 @@ export default {
 		const networkName = services.config.network.name;
 		const network = NetworkLocator.findByName(Network.NETWORKS, networkName);
 
+		const restUrl = `${services.config.rest.protocol}://localhost:${services.config.rest.port}`;
+
```
Wayonb, line 39: certs(https) are not created for localhost.  I will need to see how this works with a reverse proxy.

Wayonb (approved):
👍🏾

Wayonb (approved):
👍🏾

PR #1134, opened 2024-07-02
[sdk]: add extract signing payload to facades

Jaguar0625:
[sdk/python]: add extract_signing_payload to facades
    
     problem: there is no easy way to get signed data payload, which is needed to allow signing by hardware keys
    solution: add extract_signing_payload to facades

    [sdk/javascript]: add extractSigningPayload to facades
    
     problem: there is no easy way to get signed data payload, which is needed to allow signing by hardware keys
    solution: add extractSigningPayload to facades

### DIFF
--- a/sdk/javascript/src/facade/NemFacade.js
+++ b/sdk/javascript/src/facade/NemFacade.js
@@ -219,14 +219,23 @@ export class NemFacade {
 	}
 
 	/**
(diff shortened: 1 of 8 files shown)


Wayonb (approved):
👍🏾

AnthonyLaw (approved):
👍🏼

PR #1137, opened 2024-07-07
[client/catapult] : msvc warnings

AndreaLanfranchi:
This PR amends some warnings from MSVC toolchain as well removing some warning suppression code which is actually not needed anymore.


Jaguar0625:
what version of msvc is reporting these errors?

@Wayonb - are there any issues with any msvc CI builds?

AndreaLanfranchi:
> what version of msvc is reporting these errors?

Actually none. For example [this one](https://github.com/symbol/symbol/pull/1137/commits/f1bc24b2286239d4ce98810ba1d9956935f09c74) is clearly a leftover from some previous implementation. (There's no more shadowing)

The purpose of this PR is to **remove** silencers that are actually not needed.

Wayonb:
> @Wayonb - are there any issues with any msvc CI builds?

The Windows builds are currently passing.
Are these related to the Debug build? 🤔

AndreaLanfranchi:
> The Windows builds are currently passing. Are these related to the Debug build? 🤔

Of course they are ... but the point of this PR is to **remove** silencers that are actually not needed.
Leaving them in place could lead to the unwanted situation when, maybe due to future modifications, the warning effectively triggers but we wouldn't know as it's already silenced.

Wayonb (approved):
👍🏾

AndreaLanfranchi:
Thank you for the approval.
Either of you must merge though as I don't have write permissions.

PR #1144, opened 2024-07-12
[client/rest]: add CatapultProxy for caching responses

Jaguar0625:
problem: data routes can benefit from response caching
solution: promote fetch helper from constructionRoutes to CatapultProxy
          add caching support for repetitive data requests


Wayonb (approved):
👍🏾

PR #1152, opened 2024-07-16
[client/rest]: extend and refactor transaction to operating parsing

Jaguar0625:
problem: both construction and data routes need to parse transactions into operations
solution: refactor and extend logic in /parse


client/rest/src/plugins/rosetta/OperationParser.js
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	/**
+	 * Parses a transaction into operations.
+	 * @param {number} startOperationId First operation id to use.
+	 * @param {object} transaction Transaction to process (REST object model is assumed).
+	 * @returns {Array<Operation>} Transaction operations.
+	 * @private
+	 */
+	async parseTransactionInternal(startOperationId, transaction) {
```
Wayonb, line 166: Do you plan to add more transactions like mosaic supply change and revocation?

Wayonb (approved):
Looks good. 👍🏾

PR #1154, opened 2024-07-17
[client/rest]: add nemesis block to CatapultProxy cache

Jaguar0625:
problem: nemesis block is queried frequently
solution: cache it


Wayonb (approved):
👍🏾

PR #1155, opened 2024-07-18
[client/rest]: fixes to CatapultProxy and OperationParser and addition of more utilities for data routes

Jaguar0625:
[client/rest]: add additional utilities needed by data routes
    
     problem: common logic in data routes should be refactored
    solution: refactor and add new utilities

    [client/rest]: fixes to OperationParser
    
     problem: JSON used in tests is slightly different from REST JSON
    solution: use REST JSON in tests and update implementation

    [client/rest]: fixes to CatapultProxy
    
     problem: pagination routes return object with data array, not array directly
    solution: update fetchAll to extract data array
    
     problem: bigint to hex string formatting drops leading zeros
    solution: always return 16 character strings

### DIFF
--- a/client/rest/src/plugins/rosetta/CatapultProxy.js
+++ b/client/rest/src/plugins/rosetta/CatapultProxy.js
@@ -21,7 +21,7 @@
 
 import { RosettaErrorFactory } from './rosettaUtils.js';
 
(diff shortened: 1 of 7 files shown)


Wayonb (approved):
👍🏾

PR #1156, opened 2024-07-18
[client/rest]: fix ReceiptSource comparison in CatapultProxy

Jaguar0625:
problem: secondaryId should only be checked when primaryId is equal
solution: only check secondaryId when primaryId is equal


Wayonb (approved):
👍🏾

PR #1157, opened 2024-07-19
[client/rest]: more rosetta integration changes 

Jaguar0625:
[client/rest]: filter out zero transfer operations
    
     problem: zero transfer operations do not change state
    solution: do not create zero transfer operations

    [client/rest]: calculate transaction location in OperationParser and pass down to lookupCurrency
    
     problem: transaction location is needed for proper mosaic alias resolution
    solution: calculate it in OperationParser

### DIFF
--- a/client/rest/src/plugins/rosetta/OperationParser.js
+++ b/client/rest/src/plugins/rosetta/OperationParser.js
@@ -164,6 +164,15 @@ export class OperationParser {
 		};
 
 		const operations = [];
+		const appendOperation = operation => {
+			// filter out zero transfers
+			if ('transfer' === operation.type && '0' === operation.amount.value)
+				return;
+
+			operation.operation_identifier.index = operations.length;
+			operations.push(operation);
+		};
+
 		if (transaction.transactions) {
 			const operationGroups = Array(transaction.transactions.length);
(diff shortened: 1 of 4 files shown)


client/rest/src/plugins/rosetta/OperationParser.js
```diff
@@ -161,15 +181,17 @@ export class OperationParser {
 				// due to async nature, operations from previous transaction might not be added yet
 				// so simply using operations.length would yield wrong result
 				// instead, renumber the operations below
-				const subOperations = await this.parseTransactionInternal(0, subTransaction.transaction);
+				const subOperations = await this.parseTransactionInternal(
+					subTransaction.transaction,
+					subTransaction.meta && undefined !== subTransaction.meta.index
```
Wayonb, line 186: Just wondering when you will not be present. 🤔

Wayonb (approved):
👍🏾

client/rest/src/plugins/rosetta/OperationParser.js
Jaguar0625: when JSON is derived from sdk model

Wayonb (approved):
👍🏾

PR #1158, opened 2024-07-19
[client/rest]: add address resolution support to rosetta routes

Jaguar0625:
problem: account balances cannot be calculated correctly without resolutions
solution: resolve addresses


Wayonb (approved):
👍🏾

PR #1159, opened 2024-07-21
[client/rest]: automatically map fee from SDK JSON model to REST JSON model

Jaguar0625:
problem: SDK JSON model 'fee' should be mapped to REST JSON model 'maxFee'
solution: add logic to convertTransactionSdkJsonToRestJson


Wayonb (approved):
👍🏾

PR #1161, opened 2024-07-21
[client/rest]: refactor rosetta test utilities

Jaguar0625:
problem: data route tests can reuse construction route tests utilities
solution: extract common routines into rosettaTestUtils


Wayonb (approved):
👍🏾

PR #1170, opened 2024-07-22
[client/rest]: skip parsing of receipts without amounts

Jaguar0625:
problem: OperationParser assumes all receipts have amount property
solution: check if amount property exists


Wayonb (approved):
👍🏾

PR #1171, opened 2024-07-22
[client/rest]: workaround race condition in CatapultProxy

Jaguar0625:
problem: when multiple cache fill requests are made in parallel, properties will be parsed multiple times
solution: only parse first cache object returned

### DIFF
--- a/client/rest/src/plugins/rosetta/CatapultProxy.js
+++ b/client/rest/src/plugins/rosetta/CatapultProxy.js
@@ -114,6 +114,9 @@ export default class CatapultProxy {
 				this.fetch('blocks/1')
 			]);
 
+			if (this.cache)
+				return this.readCacheProperty(propertyName);
+
 			this.cache = {
 				nodeInfo: results[0],
 				networkProperties: results[1],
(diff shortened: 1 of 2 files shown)


client/rest/src/plugins/rosetta/CatapultProxy.js
```diff
@@ -114,6 +114,9 @@ export default class CatapultProxy {
 				this.fetch('blocks/1')
 			]);
 
+			if (this.cache)
+				return this.readCacheProperty(propertyName);
```
Wayonb, line 118: Is there still a race between the assignment of the cache object and ``currencyMosaicId`` getting formatted correctly? 🤔

Wayonb (approved):
👍🏾

client/rest/src/plugins/rosetta/CatapultProxy.js
Jaguar0625: should not be - after the await, the code should execute on main thread and there shouldn't be parallel executions afaikt

PR #1172, opened 2024-07-22
[client/rest]: rosetta integration fixes after addition of data routes 

Jaguar0625:
[rest/client]: don't return currency metadata from /account/balance
    
     problem: mesh cli does deep compare of metadata and expects it to be unset
    solution: clear metadata from route response

    [rest/client]: operations returned by construction routes should not have operation status
    
     problem: all operations have success operation status
    solution: only return operations with operation status from data routes


Wayonb (approved):
👍🏾

PR #1173, opened 2024-07-23
[client/rest]: add support to adjust account coins with mempool tx

Wayonb:
problem: user request to include mempool balance was not implemented
solution: implement mempool balance support for account

### DIFF
--- a/client/rest/src/plugins/rosetta/accountRoutes.js
+++ b/client/rest/src/plugins/rosetta/accountRoutes.js
@@ -70,7 +70,7 @@ export default {
 			return accountOperations;
 		};
 
-		const getAccountAmounts = async typedRequest => {
+		const getAccountAmounts = async (typedRequest, includeMempool) => {
 			if (typedRequest.block_identifier)
 				throw RosettaErrorFactory.NOT_SUPPORTED_ERROR;
 
@@ -81,7 +81,7 @@ export default {
 			const startChainHeight = await getChainHeight();
 
 			const promises = [services.proxy.fetch(`accounts/${address}`, json => json.account.mosaics)];
-			if (typedRequest.include_mempool)
+			if (includeMempool)
 				promises.push(services.proxy.fetchAll('transactions/unconfirmed?embedded=true', PAGE_SIZE));
 
 			const [mosaics, unconfirmedTransactions] = await Promise.all(promises);
@@ -92,8 +92,8 @@ export default {
 				throw RosettaErrorFactory.SYNC_DURING_OPERATION;
 
 			let amounts = await mapMosaicsToRosettaAmounts(mosaics);
-			if (typedRequest.include_mempool && unconfirmedTransactions.length) {
-				const combineAmountsValue = (lhs, rhs) => (BigInt(lhs.value) + BigInt(rhs.value)).toString();
+			if (includeMempool && unconfirmedTransactions.length) {
+				const addAmountsValue = (lhs, rhs) => (BigInt(lhs.value) + BigInt(rhs.value)).toString();
 
 				// combine the operations by each currency
 				const currencyAmountMap = new Map();
@@ -101,14 +101,14 @@ export default {
 				userOperations.forEach(userOperation => {
 					const currentSumAmount = currencyAmountMap.get(userOperation.amount.currency.symbol)
 						|| new Amount('0', userOperation.amount.currency);
-					currentSumAmount.value = combineAmountsValue(currentSumAmount, userOperation.amount);
+					currentSumAmount.value = addAmountsValue(currentSumAmount, userOperation.amount);
 					currencyAmountMap.set(userOperation.amount.currency.symbol, currentSumAmount);
 				});
 
 				// update the current mosaics
 				amounts.forEach(amount => {
 					if (currencyAmountMap.has(amount.currency.symbol)) {
-						amount.value = combineAmountsValue(amount, currencyAmountMap.get(amount.currency.symbol));
+						amount.value = addAmountsValue(amount, currencyAmountMap.get(amount.currency.symbol));
 						currencyAmountMap.delete(amount.currency.symbol);
 					}
 				});
@@ -126,13 +126,14 @@ export default {
 		};
 
 		server.post('/account/balance', rosettaPostRouteWithNetwork(networkName, AccountBalanceRequest, async typedRequest => {
-			const { blockIdentifier, amounts } = await getAccountAmounts(typedRequest);
+			const includeMempool = false;
+			const { blockIdentifier, amounts } = await getAccountAmounts(typedRequest, includeMempool);
 
 			return new AccountBalanceResponse(blockIdentifier, amounts);
 		}));
 
 		server.post('/account/coins', rosettaPostRouteWithNetwork(networkName, AccountCoinsRequest, async typedRequest => {
-			const { blockIdentifier, amounts } = await getAccountAmounts(typedRequest);
+			const { blockIdentifier, amounts } = await getAccountAmounts(typedRequest, typedRequest.include_mempool);
 
 			const mapRosettaAmountToCoin = amount => new Coin(new CoinIdentifier(amount.currency.metadata.id), amount);
 			const coins = await Promise.all(amounts.map(mapRosettaAmountToCoin));
--- a/client/rest/src/plugins/rosetta/accountRoutes.js
+++ b/client/rest/src/plugins/rosetta/accountRoutes.js
@@ -70,7 +70,7 @@ export default {
 			return accountOperations;
 		};
 
-		const getAccountAmounts = async (typedRequest, includeMempool) => {
+		const getAccountAmounts = async typedRequest => {
 			if (typedRequest.block_identifier)
 				throw RosettaErrorFactory.NOT_SUPPORTED_ERROR;
 
@@ -81,7 +81,7 @@ export default {
 			const startChainHeight = await getChainHeight();
 
 			const promises = [services.proxy.fetch(`accounts/${address}`, json => json.account.mosaics)];
-			if (includeMempool)
+			if (typedRequest.include_mempool)
 				promises.push(services.proxy.fetchAll('transactions/unconfirmed?embedded=true', PAGE_SIZE));
 
 			const [mosaics, unconfirmedTransactions] = await Promise.all(promises);
@@ -92,7 +92,7 @@ export default {
 				throw RosettaErrorFactory.SYNC_DURING_OPERATION;
 
 			let amounts = await mapMosaicsToRosettaAmounts(mosaics);
-			if (includeMempool && unconfirmedTransactions.length) {
+			if (typedRequest.include_mempool && unconfirmedTransactions.length) {
 				const addAmountsValue = (lhs, rhs) => (BigInt(lhs.value) + BigInt(rhs.value)).toString();
 
 				// combine the operations by each currency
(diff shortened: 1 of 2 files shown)


client/rest/test/plugins/rosetta/accountRoutes_spec.js
```diff
@@ -199,10 +238,217 @@ describe('account routes', () => {
...
 		});
 
-		addAccountSuccessTests('/account/coins', {
-			createValidRequest,
-			createRosettaAmount: createRosettaCoin,
-			Response: AccountCoinsResponse
+		const incomingSignerPublicKey = '527068DA90B142D98D27FF9BA2103A54230E3C8FAC8529E804123D986CACDCC9';
+		const incomingRecipientAddress = 'TDI2ZPA7U72GHU2ZDP4C4J6T5YMFSLWEW4OZQKI';
+		const signerPublicKey = 'ED7FE5166BDC65D065667630B96362B3E57AFCA2B557B57E02022631C8C8F1A6';
+		const recipientAddress = 'TARZARAKDFNYFVFANAIAHCYUADHHZWT2WP2I7GI';
```
Jaguar0625, line 244: imo naming is a bit awkward as (incomingSignerPublicKey, recipientAddress) and (incomingRecipientAddress, signerPublicKey) are paired.

might be better to use concepts like TargetPublicKey/Address and OtherPublicKey/Address where target matches accountid send to route

client/rest/src/plugins/rosetta/accountRoutes.js
```diff
@@ -52,43 +60,80 @@ export default {
 
 			const startChainHeight = await getChainHeight();
 
-			const mosaics = await services.proxy.fetch(`accounts/${address}`, json => json.account.mosaics);
+			const pageSize = 100;
+			const results = await Promise.all([
+				services.proxy.fetch(`accounts/${address}`, json => json.account.mosaics),
+				...(typedRequest.include_mempool ? [services.proxy.fetchAll('transactions/unconfirmed?embedded=true', pageSize)] : [])
```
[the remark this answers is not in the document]
Wayonb: yup, good catch 👍🏾

client/rest/src/plugins/rosetta/accountRoutes.js
```diff
@@ -28,21 +29,49 @@ import BlockIdentifier from './openApi/model/BlockIdentifier.js';
...
 		const networkName = services.config.network.name;
 		const network = NetworkLocator.findByName(Network.NETWORKS, networkName);
 		const lookupCurrency = createLookupCurrencyFunction(services.proxy);
 		const getChainHeight = () => services.proxy.fetch('chain/info', json => json.height);
 		const getBlockIdentifier = height => services.proxy.fetch(`blocks/${height}`)
 			.then(blockInfo => new BlockIdentifier(Number(blockInfo.block.height), blockInfo.meta.hash));
+		const parser = new OperationParser(network, {
+			includeFeeOperation: true,
+			lookupCurrency,
+			resolveAddress: (unresolvedAddress, transactionLocation) =>
```
Jaguar0625, line 50: for consistency would rename unresolvedAddress => address or change other places where OperationParser is created. could even add function to rosettaUtils for creating OperationParser for data routes

client/rest/src/plugins/rosetta/accountRoutes.js
```diff
@@ -52,43 +81,61 @@ export default {
...
+
+			const [mosaics, unconfirmedTransactions] = await Promise.all(promises);
 
 			const endChainHeight = await getChainHeight();
 
 			if (startChainHeight !== endChainHeight)
 				throw RosettaErrorFactory.SYNC_DURING_OPERATION;
 
+			let amounts = await mapMosaicsToRosettaAmounts(mosaics);
+			if (typedRequest.include_mempool && unconfirmedTransactions?.length) {
```
Jaguar0625, line 96: i don't think you need ? because you should always have unconfirmedTransactions when include_mempool is true

client/rest/src/plugins/rosetta/accountRoutes.js
```diff
@@ -52,43 +81,61 @@ export default {
...
+				const combineAmountsValue = (lhs, rhs) => (BigInt(lhs) + BigInt(rhs)).toString();
+
+				// combine the operations by each currency
+				const currencyAmountMap = new Map();
+				const userOperations = await getAccountOperations(unconfirmedTransactions, address);
+				userOperations.forEach(userOperation => {
+					const storedValue = currencyAmountMap.get(userOperation.amount.currency.symbol)?.value || 0;
```
Jaguar0625, line 103: alternatively, might be clearer to use temp object with value:
```js
const amount = currencyAmountMap.get(userOperation.amount.currency.symbol) || { value: 0 };
```

client/rest/src/plugins/rosetta/accountRoutes.js
```diff
@@ -52,43 +81,61 @@ export default {
...
+					userOperation.amount.value = combineAmountsValue(userOperation.amount.value, storedValue);
```
Jaguar0625, line 104: did you intend to update the userOperation value too? original version was only updating stored value

client/rest/test/plugins/rosetta/accountRoutes_spec.js
```diff
@@ -79,6 +81,41 @@ describe('account routes', () => {
...
+		stubNamespaceResolution('new.coin', '2345678901ABBCCD');
+	};
+
+	const createTransactionJson = (mosaics, signerPublicKey, recipientAddress) => {
+		const facade = new SymbolFacade('testnet');
+		const transaction = facade.transactionFactory.create({
+			type: 'transfer_transaction_v1',
+			recipientAddress,
+			signerPublicKey,
+			mosaics: [...mosaics],
```
Jaguar0625, line 107: just `mosaics` unless there is some reason you want to make (shallow) copy

client/rest/test/plugins/rosetta/accountRoutes_spec.js
```diff
@@ -188,6 +225,11 @@ describe('account routes', () => {
 	// region coins
 
 	describe('coins', () => {
+		const OTHER_SIGNER_PUBLIC_KEY = '527068DA90B142D98D27FF9BA2103A54230E3C8FAC8529E804123D986CACDCC9';
+		const OTHER_RECIPIENT_ADDRESS = 'TDI2ZPA7U72GHU2ZDP4C4J6T5YMFSLWEW4OZQKI';
+		const TARGET_SIGNER_PUBLIC_KEY = 'ED7FE5166BDC65D065667630B96362B3E57AFCA2B557B57E02022631C8C8F1A6';
+		const TARGET_RECIPIENT_ADDRESS = 'TARZARAKDFNYFVFANAIAHCYUADHHZWT2WP2I7GI';
```
Jaguar0625, line 231: do you intend to have OTHER* and TARGET* ids to not point to same account?

Jaguar0625: i mean this reads like it is sending to self (even though it is not)
```
const transaction = createTransactionJson(mosaicsToReceive, OTHER_SIGNER_PUBLIC_KEY, OTHER_RECIPIENT_ADDRESS);
```

client/rest/test/plugins/rosetta/accountRoutes_spec.js
```diff
@@ -199,10 +241,212 @@ describe('account routes', () => {
...
+			const transactionJson = createUnconfirmedTransactionsResponse([transaction]);
+
+			// Act + Assert:
+			await assertIncludeMempoolAccountTest(transactionJson, expectedCoins);
+		});
+
+		it('succeeds when multiple coins sent in the mempool (include fee)', async () => {
+			// Arrange:
+			const mosaicsToSend = [
+				{ mosaicId: generateMosaicAliasId('cat.dog'), amount: 100 },
```
Jaguar0625, line 394: would change rest of amounts to != 100 too

client/rest/test/plugins/rosetta/accountRoutes_spec.js
```diff
@@ -199,10 +241,212 @@ describe('account routes', () => {
...
+			const expectedCoins = [
+				createRosettaCoin('123', 'foo.bar', 3, '1122334455667788'),
+				createRosettaCoin('262', 'cat.dog', 4, '0F61E4A360897965'),
+				createRosettaCoin('112', 'symbol.xym', 6, '1ABBCCDDAABBCCDD')
+			];
+			const transaction = createTransactionJson(mosaicsToSend, TARGET_SIGNER_PUBLIC_KEY, TARGET_RECIPIENT_ADDRESS);
+			const transactionJson = createUnconfirmedTransactionsResponse([transaction]);
+
+			// Act + Assert:
+			await assertIncludeMempoolAccountTest(transactionJson, expectedCoins, false);
```
Jaguar0625, line 449: if possible try to add test that parses aggregate

client/rest/test/plugins/rosetta/accountRoutes_spec.js
```diff
@@ -199,10 +238,217 @@ describe('account routes', () => {
...
+
+			// Act + Assert:
+			await assertRosettaSuccessBasic('/account/coins', request, expectedResponse);
+		};
+
+		const createUnconfirmedTransactionsResponse = transactionJsons => ({
+			data: transactionJsons
+		});
+
+		it('succeeds when all fetches succeed', async () => {
```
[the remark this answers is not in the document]
Wayonb: Yes, the other tests still works.

client/rest/test/plugins/rosetta/accountRoutes_spec.js
Wayonb: Ok. will update hopefully it will be better 😅

client/rest/test/plugins/rosetta/accountRoutes_spec.js
Wayonb: ok.  I will take a look.  I guess there should be one also for the block routes

client/rest/test/plugins/rosetta/accountRoutes_spec.js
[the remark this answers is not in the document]
Jaguar0625: did you miss this one?

client/rest/src/plugins/rosetta/accountRoutes.js
```diff
@@ -52,43 +80,62 @@ export default {
...
+			const [mosaics, unconfirmedTransactions] = await Promise.all(promises);
 
 			const endChainHeight = await getChainHeight();
 
 			if (startChainHeight !== endChainHeight)
 				throw RosettaErrorFactory.SYNC_DURING_OPERATION;
 
+			let amounts = await mapMosaicsToRosettaAmounts(mosaics);
+			if (typedRequest.include_mempool && unconfirmedTransactions.length) {
+				const combineAmountsValue = (lhs, rhs) => (BigInt(lhs.value) + BigInt(rhs.value)).toString();
```
Jaguar0625, line 96: maybe instead of "combine", consider "sum" or "add"

client/rest/test/plugins/rosetta/accountRoutes_spec.js
```diff
@@ -121,6 +144,80 @@ describe('account routes', () => {
...
+				transaction: convertTransactionSdkJsonToRestJson(aggregateTransaction.toJson()),
+				meta: {
+					hash: TRANSACTION_HASH,
+					height: '0',
+					index: 1
+				}
+			};
+		};
+
+		const assertIncludeMempoolAccountTest = async (transactions, expectedCoins, includeMempool = true) => {
```
Jaguar0625, line 187: i think there was miscommunication again?

these should be under `describe('/account/coins'` since they're only testing that route (not both). currently, these are being added twice - once under `describe('/account/coins` and again under `describe('/account/balance`. 

only the other (two?) routes that are generic and applicable to both routes should be added in addAccountSuccessTests. even if they have the same `it` names, they'll be testing the appropriate (different) routes.

Wayonb: Updated to what you said.  After I debug this realized that the include_mempool gets striped out when we create the request object(`` Request.constructFromObject(req.body);``).  So its not an issue and I don't need to add the tests for ``/account/balance``.

Jaguar0625 (approved):
👍

PR #1175, opened 2024-07-23
[client/rest]: update rosetta to support id being set in currency metadata

Jaguar0625:
problem: resolved mosaic id should be specified in currency metadata
solution: update construction/parse to resolve 'currencyMosaicId' and set metadata
          restore currency metadata in account/balance


Wayonb (approved):
👍🏾

PR #1176, opened 2024-07-25
[client/rest]: fix aggregate transaction handling

Jaguar0625:
problem: rosetta requires all signatures to be made in one shot, which isn't supported by Symbol natively
 solution: add "fee payer" account as main transaction account so transaction hash can be calculated deterministically and passed to cosigners


Wayonb (approved):
👍🏾

PR #1178, opened 2024-07-28
[client/rest]: refactor most of process hosting into server/process

Jaguar0625:
problem: rosetta is bundled within catapult REST directly
solution: move rosetta implementation into new mini REST

### DIFF
--- a/client/rest/resources/rest.rosetta.json
+++ b/client/rest/resources/rest.rosetta.json
@@ -1,0 +1,48 @@
+{
+  "network": {
+    "name": "testnet",
+    "description": "catapult public test network"
+  },
+
+  "port": 4000,
+  "protocol": "HTTPS",
+  "sslKeyPath": "",
+  "sslCertificatePath": "",
+  "crossDomain": {
+    "allowedHosts": ["*"],
+    "allowedMethods": ["GET", "POST", "PUT", "OPTIONS"]
+  },
+
+  "restEndpoint": "http://localhost:3000",
+  "routeExtensions": [
+    "rosetta"
+  ],
+
+  "throttling": {
+    "burst": 20,
+    "rate": 5
+  },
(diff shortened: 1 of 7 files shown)


client/rest/resources/rest.rosetta.json
```diff
@@ -0,0 +1,48 @@
...
+  "port": 4000,
+  "protocol": "HTTPS",
+  "sslKeyPath": "",
+  "sslCertificatePath": "",
+  "crossDomain": {
+    "allowedHosts": ["*"],
+    "allowedMethods": ["GET", "POST", "PUT", "OPTIONS"]
+  },
+
+  "restEndpoint": "http://localhost:3000",
```
Wayonb, line 16: I am not sure if it's possible to call the HTTP endpoint from HTTPS.

client/rest/src/index.js
```diff
@@ -122,10 +59,7 @@ const registerRoutes = (server, db, services) => {
 	const servicesView = {
 		config: {
 			network: services.config.network,
-			rest: {
-				protocol: services.config.protocol,
-				port: services.config.port
-			},
+			restEndpoint: `${services.config.protocol}://localhost:${services.config.port}`,
```
Wayonb, line 62: Should this be restEndpoint in the config?

Wayonb (approved):

client/rest/src/index.js
Jaguar0625: this is only when rosetta is hosted within main REST, so, this is just endpoint of REST (which is shared with rosetta REST)

client/rest/resources/rest.rosetta.json
Jaguar0625: in that case, we'd need to update it as part of deployment

PR #1188, opened 2024-08-02
[client/rest]: add support for NEM /construction/derive

Jaguar0625:
[client/rest]: add support for NEM /construction/derive
    
     problem: start adding NEM construction routes
    solution: add NEM /construction/derive
              refactor shared Symbol and NEM test utilities

    [client/rest]: promote shared NEM and Symbol rosetta utils
    
     problem: all rosetta utils are under /symbol but some can be used by nem
    solution: move shared utils up a level


Wayonb (approved):
👍🏾

PR #1197, opened 2024-08-05
[client/rest]: properly handle fee calculation when block fee multiplier is zero

Jaguar0625:
problem: weak comparison is treating zero as falsy
solution: check feeMultiplier against undefined


Wayonb (approved):
👍🏾

PR #1199, opened 2024-08-06
[client/rest]: add NEM createLookupCurrencyFunction

Jaguar0625:
problem: OperationParser needs to be initialized with a lookupCurrency function
solution: add a default implementation
          cache mosaic properties


Wayonb (approved):
👍🏾

PR #1220, opened 2024-08-12
[sdk/javascript]: export bytesToInt via utils

Jaguar0625:
problem: bytesToBigInt is exported but not bytesToInt
solution: export both


Wayonb (approved):
👍🏾

PR #1221, opened 2024-08-12
[client/rest]: first group of NEM rosetta related fixes

Jaguar0625:
[client/rest]: rename modification cosignatoryPublicKey to cosignatoryAccount
    
     problem: NEM REST uses name 'cosignatoryAccount' instead of SDK name 'cosignatoryPublicKey'
    solution: add mapping in convertTransactionSdkJsonToRestJson

    [client/rest]: fix version handling when processing NEM transaction
    
     problem: NEM version is composed of network-type and version
    solution: mask version part when comparing

    [client/rest]: add nem rosetta to known plugins list
    
     problem: nem rosetta is not in list
    solution: add it


Wayonb (approved):
Looks good 👍🏾 but some other tests need to be updated

Wayonb (approved):
👍🏾

PR #1222, opened 2024-08-13
[client/rest]: NEM rosetta /block needs to create rosetta transaction for harvest fee

Jaguar0625:
problem: harvest fees are not extracted by rosetta
solution: add block transaction consisting of harvest fee


Wayonb (approved):
👍🏾

PR #1227, opened 2024-08-16
[sdk/javascript]: fix TS descriptor generator

Jaguar0625:
problem: TS descriptor generator does not work
solution: fix it
          add dryrun to build script


Wayonb (approved):
👍🏾

PR #1228, opened 2024-08-16
[catbuffer/schemas]: NEM multisig is not correct

Jaguar0625:
[catbuffer/schemas]: add NonVerifiableCosignatureV1
    
     problem: cosignature transactions cannot currently be signed
    solution: add non-verifiable version for signing

    [catbuffer/schemas]: NEM cosignature multisig_transaction_hash field should be other_transaction_hash
    
     problem: field contains hash of other - not multisig - transaction
    solution: rename it


Wayonb (approved):
👍🏾

PR #1229, opened 2024-08-16
[client/rest]: fix cosigning of NEM transactions in rosetta construction 

Jaguar0625:
[client/rest]: fix multisig transaction processing
    
     problem: NEM multisig transaction processing is wrong
    solution: fix multisig transaction construction

    [client/rest]: fix multisig (outer) fee calculation
    
     problem: multisig is using inner transaction fee
    solution: use constant fee


Wayonb (approved):
Looks good 👍🏾

There is still one missing fix.  For multisig conversion to work, no cosigner is used so in the preprocess you will need to request all the public keys for ``addressAdditions``

PR #1230, opened 2024-08-16
[client/rest]: request public keys for multisig address additions

Jaguar0625:
problem: NEM requires public keys of addresses being added to multisig
solution: request public keys of addressAdditions in /preprocess


Wayonb (approved):
👍🏾

PR #1231, opened 2024-08-16
[client/rest]: fix /construction/submit for multisig transactions

Jaguar0625:
problem: NEM doesn't have concept of complete multisig and strips embedded multisig signatures
solution: submit multisig transaction and cosignatures separately


Wayonb (approved):
👍🏾

PR #1239, opened 2024-08-19
[client/rest]: update NEM rosetta to use new /mosaic/definition/last api

Jaguar0625:
problem: rosetta implementation is unable to get information for expired mosaics
solution: use last (active or expired) mosaic definition for processing

### DIFF
--- a/client/rest/src/plugins/rosetta/nem/NemProxy.js
+++ b/client/rest/src/plugins/rosetta/nem/NemProxy.js
@@ -104,15 +104,7 @@ export default class NemProxy {
 		if (this.mosaicPropertiesMap.has(fullyQualifiedName))
 			return this.mosaicPropertiesMap.get(fullyQualifiedName);
 
-		const mosaicDefinitions = await this.fetchAll(
-			`namespace/mosaic/definition/page?namespace=${mosaicId.namespaceId}`,
-			100,
-			jsonObject => jsonObject.mosaic
-		);
-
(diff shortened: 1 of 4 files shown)


client/rest/src/plugins/rosetta/nem/NemProxy.js
```diff
@@ -104,15 +104,7 @@ export default class NemProxy {
...
 
-		const mosaicDefinitions = await this.fetchAll(
-			`namespace/mosaic/definition/page?namespace=${mosaicId.namespaceId}`,
-			100,
-			jsonObject => jsonObject.mosaic
-		);
-
-		const mosaicDefinition = mosaicDefinitions.find(definition => definition.id.name === mosaicId.name);
-		if (!mosaicDefinition)
-			throw RosettaErrorFactory.INTERNAL_SERVER_ERROR;
```
Wayonb, line 115: Should we still not fail, or should the mosaic def always be there?

Wayonb (approved):
👍🏾

client/rest/src/plugins/rosetta/nem/NemProxy.js
Jaguar0625: in case of missing, NIS will return 404, which will be mapped by `global.fetch` to `result.ok === false`. 
thus, `NemProxy.fetch` will throw `RosettaErrorFactory.CONNECTION_ERROR`

PR #1242, opened 2024-08-20
[jenkins]: localDepsLocal boost build is failing with clang

Jaguar0625:
problem: clang is defaulting to older c++ standard
solution: specify c++ standard version to b2 explicitly


Wayonb (approved):
👍🏾

PR #1243, opened 2024-08-22
Rest/rosetta/fixes

Jaguar0625:
[rest/client]: truncate percentile levy amounts
    
     problem: when calculating percentile levy amount, it must be full number
    solution: truncate it

    [rest/client]: map SDK JSON 'action' to REST JSON 'supplyType'
    
     problem: REST JSON and SDK JSON use different property names
    solution: add mapping to 'convertTransactionSdkJsonToRestJson'

    [client/rest]: add missing await to lookupCurrency call
    
     problem: call to lookupCurrency is missing await
    solution: add await


Wayonb (approved):
👍🏾

PR #1253, opened 2024-08-26
[client/rest]: truncate mosaic transfer amount calculated from bag

Jaguar0625:
problem: mosaic transfer bags can be fractional, causing calculations to result in floating point values
solution: truncate calculated mosaic transfer amount derived from mosaic bags


Wayonb (approved):
👍🏾

PR #1254, opened 2024-08-26
[client/rest]: refactor creation of OperationParser

Jaguar0625:
problem: OperationParser is constructed same way in all online routes
solution: refactor creation in helper function (per blockchain)


Wayonb (approved):
👍🏾

PR #1256, opened 2024-08-27
[client/rest]: update NEM rosetta to handle mosaic (re)definitions

Jaguar0625:
[client/rest]: update NEM rosetta to handle mosaic (re)definitions
    
     problem: NEM mosaic (re)definition behavior is dependent on current state
    solution: integrate 'mosaic/definition/supply' into mosaic definition creation
              transaction processing

    [client/rest]: in NEM rosetta use ':' as token separator to match NEM
    
     problem: NEM uses ':' as mosaic name separator but rosetta implementation is using '.'
    solution: use ':' as separator for consistency


Wayonb (approved):
👍🏾

Wayonb (approved):
👍🏾

PR #1257, opened 2024-08-30
[client/rest]: query new NIS expired mosaics endpoint when processing blocks

Jaguar0625:
problem: rosetta needs to be informed about expiring mosaics and zeroing of balances
solution: integrate new local/mosaics/expired NIS API


Wayonb (approved):
👍🏾

PR #1265, opened 2024-09-03
[client/rest]: NEM rosetta should only credit LAST unique mosaic definition in block

Jaguar0625:
problem: NEM allows multiple definitions of same mosaic in block, but only last one is relevant
solution: add postprocessing in parseBlock to debit earlier duplicate mosaic definitions

### DIFF
--- a/client/rest/src/plugins/rosetta/nem/OperationParser.js
+++ b/client/rest/src/plugins/rosetta/nem/OperationParser.js
@@ -21,6 +21,7 @@
 
 import {
 	areMosaicDefinitionsEqual,
+	areMosaicIdsEqual,
 	createLookupCurrencyFunction,
 	getBlockchainDescriptor,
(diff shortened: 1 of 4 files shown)


client/rest/src/plugins/rosetta/nem/OperationParser.js
```diff
@@ -495,6 +496,30 @@ export class OperationParser {
 			});
 		}
 
+		// if the same mosaic is defined multiple times in a single block, only the last definition is relevant
+		// fix up balances by removing supplies created by earlier definitions
+		if (block.block.txes) {
```
Wayonb, line 501: ``block.block.txes`` -> ``block.txes``

same below.

Wayonb (approved):
👍🏾

Wayonb (approved):
👍🏾

PR #1266, opened 2024-09-04
[client/rest]: use 64-bit arithmetic when calculating transfer amounts

Jaguar0625:
problem: some intermediate values can exceed Number.MAX_SAFE_INTEGER, leading to loss of precision
solution: use BitInt math and convert result to Number


Wayonb (approved):
👍🏾

PR #1267, opened 2024-09-04
[client/rest]: update NEM rosetta to handle restored mosaics

Jaguar0625:
problem: NIS API has been modified to return restored mosaics and include expiredMosaicType flag in output
solution: check flag and credit or debit as appropriate


Wayonb (approved):
👍🏾

PR #1268, opened 2024-09-06
[client/rest]: add handling for expirationHeight

Jaguar0625:
problem: NIS /local/mosaic/definition/supply API was extended to include expirationHeight
solution: handle expirationHeight in NEM rosetta operation parser


Wayonb (commented):
👍🏾

PR #1269, opened 2024-09-07
[client/rest]: consider network state when undoing duplicate mosaic definitions in block

Jaguar0625:
problem: parseBlock is undoing duplicate mosaic definitions assuming they are new mosaic definitions
solution: consider network state when determining amount to undo


Wayonb (approved):
👍🏾

PR #1276, opened 2024-09-09
[client/rest]: add mosaic redefinition fork handling to rosetta implementation

Jaguar0625:
problem: prior to the mosaic redefinition fork, mosaic description changes are considered significant
solution: change comparison of mosaic definitions prior to fork

### DIFF
--- a/client/rest/src/plugins/rosetta/nem/OperationParser.js
+++ b/client/rest/src/plugins/rosetta/nem/OperationParser.js
@@ -24,6 +24,7 @@ import {
 	areMosaicIdsEqual,
 	createLookupCurrencyFunction,
 	getBlockchainDescriptor,
+	isMosaicDefinitionDescriptionSignificant,
 	mosaicIdToString
 } from './rosettaUtils.js';
 import AccountIdentifier from '../openApi/model/AccountIdentifier.js';
@@ -427,7 +428,14 @@ export class OperationParser {
 			);
 
 			let creditAmount = supplyToAtomicUnits(initialSupply, divisibility);
-			if (existingMosaicDefinitionTuple) {
+
+			// prior to the mosaic redefinition fork, mosaic description changes are significant
+			// in addition, prior to this fork, NIS reports the zeroing of all mosaic balances over the NIS mosaics/expired endpoint
+			// as a result, no special handling is needed for existing mosaics since their existing balances will be zeroed
+			// by data sent over this endpoint
+			const isDescriptionSignificant = isMosaicDefinitionDescriptionSignificant(this.network, transactionLocation);
+
+			if (!isDescriptionSignificant && existingMosaicDefinitionTuple) {
 				const existingMosaicDefinition = existingMosaicDefinitionTuple.mosaicDefinition;
 
 				const isExistingPurged = transactionLocation && transactionLocation.height > existingMosaicDefinitionTuple.expirationHeight;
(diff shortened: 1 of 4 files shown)


client/rest/src/plugins/rosetta/nem/OperationParser.js
```diff
@@ -432,7 +433,8 @@ export class OperationParser {
 
 				const isExistingPurged = transactionLocation && transactionLocation.height > existingMosaicDefinitionTuple.expirationHeight;
 				if (!isExistingPurged) {
-					if (areMosaicDefinitionsEqual(mosaicDefinition, existingMosaicDefinition)) {
+					const isDescriptionSignificant = isMosaicDefinitionDescriptionSignificant(this.network, transactionLocation);
```
Wayonb, line 436: If ``isDescriptionSignificant`` is true, then if the owner is the same, we want to do nothing.
From what I noticed, before the fork, the expired endpoint was triggered when the mosaic was redefined.

client/rest/src/plugins/rosetta/nem/OperationParser.js
Wayonb, line 1: In the ``parseBlock``

1. At line 493, the height should be height - 1. In some cases, before the fork, a mosaic redefinition will trigger the expired endpoint for the old definition of the mosaic. So, the new mosaic definition and the expired old definition will be the same height.

```
currency: (await this.options.lookupCurrency(expiredMosaic.mosaicId, { height })).currency,
```

2. at line 515, also check for multi-sig transaction(otherTrans) for mosaic def.
```
const transaction = block.block.transactions[i];
```

Wayonb (changes requested):

client/rest/src/plugins/rosetta/nem/rosettaUtils.js
```diff
@@ -131,3 +132,20 @@ export const areMosaicDefinitionsEqual = (lhs, rhs, isDescriptionSignificant) =>
...
+ * @param {Network} network Blockchain network.
+ * @param {object} transactionLocation Location of transaction.
+ * @returns {boolean} true if mosaic definition description differences are significant.
+ */
+export const isMosaicDefinitionDescriptionSignificant = (network, transactionLocation) => {
+	const forkHeight = (() => {
+		if ('mainnet' === network.name)
+			return 1_110_000;
+
+		return 'testnet' === network.name ? 871_500 : 1;
```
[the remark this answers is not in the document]
Jaguar0625: would change it later as this is consistent with code in NIS (which means we might need NIS change for testnet) 😅 
see: https://github.com/NemProject/nem/blob/dev/nis/src/main/java/org/nem/nis/BlockMarkerConstants.java#L8

client/rest/src/plugins/rosetta/nem/OperationParser.js
Jaguar0625: so, you're saying we should always treat description as insignificant for purposes of rosetta?

client/rest/src/plugins/rosetta/nem/OperationParser.js
Jaguar0625: were these only changes needed to get Symbol rosetta to work or were there others as well?

Wayonb: yea, that's all I think

Wayonb (approved):
👍🏾

PR #1285, opened 2024-09-17
[jenkins] fix: create jenkins c++ versioned caching

Wayonb:
problem: ABI for compilers is forward compatibility
solution: move compiler caching into a versioned folder


AnthonyLaw (approved):
👍🏼

PR #1295, opened 2024-09-25
[jenkins]: upgrade to clang 19

Wayonb:
problem: not building with the latest version of Clang
solution: upgrade to Clang 19

### DIFF
--- a/client/catapult/tests/catapult/io/PodIoUtilsTests.cpp
+++ b/client/catapult/tests/catapult/io/PodIoUtilsTests.cpp
@@ -143,7 +143,7 @@ namespace catapult { namespace io {
 			mocks::MockMemoryStream stream(buffer);
 
 			// Act:
-			WriteTraits::template Write(stream, source);
+			WriteTraits::template Write<mocks::MockMemoryStream, T>(stream, source);
 			stream.seek(0);
 			return TReadTraits::template Read<T>(stream);
 		}
--- a/jenkins/catapult/runDockerTestsInnerTest.py
+++ b/jenkins/catapult/runDockerTestsInnerTest.py
@@ -1,7 +1,7 @@
 import argparse
 import sys
-from pathlib import Path
 from os import sep
+from pathlib import Path
 
 from configuration import load_compiler_configuration
 from environment import EnvironmentManager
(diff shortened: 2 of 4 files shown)


jenkins/catapult/runDockerTestsInnerTest.py
```diff
@@ -168,14 +169,18 @@ def main():
 		environment_manager.set_env_var('LD_LIBRARY_PATH', f'{output_path}/lib:{output_path}/deps')
 	logs_path = Path(args.out_dir) / 'logs'
 
+	# There seems to be a bug in gtest where specifying the `--gtest_output=` parameter causes
+	# llvm-symbolizer not to resolves its libraries.
+	# setting the test_output using environment variable works.
+	environment_manager.set_env_var('GTEST_OUTPUT', f'xml:{logs_path}{sep}')
```
Jaguar0625, line 175: shouldn't this be `base_output_filepath` like previous?

client/catapult/tests/catapult/io/PodIoUtilsTests.cpp
```diff
@@ -143,7 +143,7 @@ namespace catapult { namespace io {
 			mocks::MockMemoryStream stream(buffer);
 
 			// Act:
-			WriteTraits::template Write(stream, source);
+			WriteTraits::template Write<mocks::MockMemoryStream, T>(stream, source);
```
Jaguar0625, line 146: what was warning/error?

jenkins/catapult/runDockerTestsInnerTest.py
Wayonb: That's also an option but would need to set the environment variable for each test run.

If ``GTEST_OUTPUT`` is set to the output directory then it will automatically add the test filename.

I can move it back into the loop if you want.

Jaguar0625: it's fine, just add comment as to behavior [it will automatically add the test filename]

client/catapult/tests/catapult/io/PodIoUtilsTests.cpp
Wayonb: https://clang.llvm.org/docs/DiagnosticsReference.html#wmissing-template-arg-list-after-template-kw

Jaguar0625 (approved):
👍

PR #1296, opened 2024-09-27
[jenkins] fix: fail to decode line on Windows

Wayonb:
problem: building catapult deps fails on Windows due to an invalid UTF-8 character
solution: add exception handling

### DIFF
--- a/jenkins/catapult/process.py
+++ b/jenkins/catapult/process.py
@@ -18,7 +18,11 @@ class ProcessManager:
 		with Popen(command_line, stdout=PIPE, stderr=STDOUT) as process:
 			process_lines = []
 			for line_bin in iter(process.stdout.readline, b''):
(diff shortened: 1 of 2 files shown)


Jaguar0625:
what is source of `invalid UTF-8 character` ?

Wayonb:
> what is source of `invalid UTF-8 character` ?

It happens when build the mongocxx

Jaguar0625:
is it possible to conditionally enable this new handling only for mongocxx by adding another paramter?

Wayonb:
> is it possible to conditionally enable this new handling only for mongocxx by adding another parameter?

The best I can do is enable exception handling only when building the catapult dependencies. This is one command.

Jaguar0625 (approved):
👍

PR #1297, opened 2024-09-27
[jenkins] fix: mongo-cxx-driver libraries name change in r3.10

Wayonb:
problem: starting with mongocxx r3.10.0, MSVC toolchain on Windows generates different library names
solution: disable this feature by setting ENABLE_ABI_TAG_IN_LIBRARY_FILENAMES=OFF


jenkins/catapult/installDepsLocal.py
```diff
@@ -96,6 +96,10 @@ def build_git_dependency(self, organization, project):
 			# https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/
 			cmake_options += ['-DCMAKE_CXX_FLAGS="/Zc:__cplusplus"', f'-DCMAKE_PREFIX_PATH={self.target_directory / organization}']
 
+			version = self.versions[f'{organization}_{project}']
+			if 'r3.10.0' <= version:
```
Jaguar0625, line 100: this is just lexical compare, so it's not *completely* right, but probably good enough

Jaguar0625 (approved):

PR #1298, opened 2024-09-27
[jenkins] fix: rocksdb fails with gcc 12.3

Wayonb:
problem: rocksdb fails to build with GCC 12.3 due to an array-bounds error
solution: disable the warning for GCC 12


Jaguar0625 (approved):
👍

PR #1318, opened 2024-10-09
[client/rest]: block websocket handler needs to map block SDK JSON to REST JSON

Jaguar0625:
problem: block server message handler is propagating SDK JSON, which is slightly different from REST JSON
          proof fields are dropped from output because they are contained in subobject in SDK JSON but not REST JSON
solution: map block SDK JSON to REST JSON before propagating


Wayonb (approved):
👍🏾

PR #1325, opened 2024-10-15
[jenkins] fix: conan fails to set home path on Windows

Wayonb:
problem: latest version of Conan fails to set home path on Windows
solution: Conan now requires absolute path in the format of ``c:\conan``


Jaguar0625 (approved):
👍

PR #1336, opened 2024-10-21
[client/rest]: Initial the light rest project

AnthonyLaw:
Initial the light rest project


client/rest/resources/rest.light.json
```diff
@@ -0,0 +1,62 @@
...
+    "protocol": "HTTPS",
+    "sslKeyPath": "",
+    "sslCertificatePath": "",
+    "crossDomain": {
+        "allowedHosts": ["*"],
+        "allowedMethods": ["GET", "POST", "PUT", "OPTIONS"]
+    },
+
+    "routeExtensions": [
+        "light"
```
Wayonb, line 17: Should light be under extensions vs routeExtensions? 🤔

client/rest/src/index.light.js
```diff
@@ -0,0 +1,92 @@
...
+import messageFormattingRules from './server/messageFormattingRules.js';
+import runProcess from './server/process.js';
+import sshpk from 'sshpk';
+import winston from 'winston';
+import fs from 'fs';
+
+const createServer = config => {
+	const modelSystem = catapult.plugins.catapultModelSystem.configure([], {
+		json: dbFormattingRules,
+		ws: messageFormattingRules
```
Wayonb, line 37: Do you need ws rules?

client/rest/src/index.light.js
```diff
@@ -0,0 +1,92 @@
...
+		},
+		connections: services.connectionService
+	};
+
+	// 2. configure extension routes
+	routeSystem.configure(
+		services.config.routeExtensions,
+		server,
+		db,
+		servicesView
```
Wayonb, line 59: Can this be on one line?
``routeSystem.configure(services.config.routeExtensions, server, db, servicesView);``

client/rest/src/plugins/light/light.js
```diff
@@ -0,0 +1,41 @@
...
+
+/** @module plugins/light */
+import nodeRoutes from './nodeRoutes.js';
+
+/**
+ * Creates a light plugin.
+ * @type {module:plugins/CatapultRestPlugin}
+ */
+export default {
+	createDb: db => db,
```
Wayonb, line 30: should be able to do ``createDb: () => {}`` since no DB.

client/rest/src/plugins/light/light.js
```diff
@@ -0,0 +1,41 @@
...
+
+	registerTransactionStates: () => {},
+
+	registerMessageChannels: () => {},
+
+	registerRoutes: (server, db, services) => {
+		[
+			nodeRoutes
```
Wayonb, line 38: Can you load the light node routes into the standard REST node? I guess it will fail since the endpoints are duplicates 🤔

Wayonb (changes requested):
is ``node/peers`` endpoint needed for static server to work

client/rest/resources/rest.light.json
AnthonyLaw: hmm, I think it make sense.
as `light` is a REST route, instead of a new plugins

client/rest/src/plugins/light/light.js
AnthonyLaw: We can use different prefixes to identify the `nodeRoutes` or `lightNodeRoutes`.
But why do we need to do that as we need to separate the REST service right?

client/rest/src/index.light.js
AnthonyLaw: We only need ws rules.

AnthonyLaw:
> is node/peers endpoint needed for static server to work

do you mean `node/peers ` is needed for light rest ?

client/rest/resources/rest.light.json
Wayonb: Ok, should this route always be loaded instead of an extension?

Wayonb:
> do you mean `node/peers ` is needed for light rest ?

no, I meant for the statistics service.  If we can move nodewatch that's even better.

client/rest/resources/rest.light.json
AnthonyLaw: > Ok, should this route always be loaded instead of an extension?
yes, we no need anything extension.

Wayonb (approved):
👍🏾

PR #1341, opened 2024-10-30
[sdk/javascript] fix: wasm-opt failing with rustc 1.82.0

Wayonb:
problem: rustc 1.82.0 seems to break wasm-opt
solution: downgrade to rustc 1.81.0 for now.


Jaguar0625 (approved):
👍

PR #1361, opened 2024-11-12
[jenkins] fix: pip upgrade fails with Python 3.13.0

Wayonb:
problem: Windows docker image for Python and JavaScript is failing with Python 3.13.0
solution: Use the default pip installs with Python 3.13.0


Jaguar0625 (approved):
👍

PR #1364, opened 2024-11-15
[client/rest]: Added light rest readme

AnthonyLaw:
added readme docs

### DIFF
--- a/client/rest/README_LIGHT.md
+++ b/client/rest/README_LIGHT.md
@@ -1,0 +1,22 @@
+# Light-rest
+
+## Introduction
+
+This is a lightweight version of the REST server. It provides a limited set of endpoints compared to the full API node. The main differences between the light rest and the full API node are:
+
+- The light rest does not use MongoDB or broker services.
+- it communicates with the peer node using a socket connection.
+
+## Purpose
+
+The purpose of the light rest is to reduce server resource usage, making it a cost-effective. it also offers users the ability to delegate harvesting to the peer node.
+
+## Supported endpoints
+
+- chain/info
+- node/info
+- node/peers
+- node/unlockedaccount
+- node/server
+
+review the [REST documentation](https://docs.symbol.dev/api.html) for more information.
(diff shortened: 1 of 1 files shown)


client/rest/README_LIGHT.md
```diff
@@ -0,0 +1,22 @@
+# Light-rest
+
+## Introduction
+
+This is a lightweight version of the REST server. It provides a limited set of endpoints compared to the full API node. The main differences between the light rest and the full API node are:
```
Wayonb, line 5: try to rephrase...maybe something like this:

```
The light version of the REST server provides limited endpoints compared to the full API node. The main differences between the two are:
```

client/rest/README_LIGHT.md
```diff
@@ -0,0 +1,22 @@
...
+
+- The light rest does not use MongoDB or broker services.
+- it communicates with the peer node using a socket connection.
```
Wayonb, line 8: it -> It

Maybe change to:
```
Light REST only communicates with the server using a socket connection.
```

client/rest/README_LIGHT.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Purpose
+
+The purpose of the light rest is to reduce server resource usage, making it a cost-effective. it also offers users the ability to delegate harvesting to the peer node.
+
```
Wayonb, line 13: Does Wallet still work? 🤔

Wayonb (changes requested):

client/rest/README_LIGHT.md
AnthonyLaw: technically it should work, but I found out our statistic service didn't pick up `nodePublicKey`, something I need to fix and test it again in desktop wallet.

Wayonb: I think the functionality available in the wallet will be limited if database and websocket is not available. 🤔
I will that for your testing.

Wayonb (approved):
👍🏾

PR #1371, opened 2024-11-20
[jenkins] fix: add Python 3.13 to build

Wayonb:
problem: automation is not using the latest Python release
solution: update the automation to use Python 3.13 for Ubuntu latest release


Jaguar0625 (approved):
👍

PR #1391, opened 2024-12-03
[sdk/javascript] fix: disable polyfill in webpack

Wayonb:
problem: warning: Can't resolve 'vm' in webpack for polyfill
solution: disable polyfill for the webpack


Jaguar0625 (approved):
👍

PR #1405, opened 2024-12-09
[monorepo] fix: remove the mocha-jenkins-reporter tool

Wayonb:
problem: mocha-jenkins-reporter tool seems to not be used
solution: remove mocha-jenkins-reporter from package.json


Jaguar0625 (approved):
👍

PR #1416, opened 2024-12-18
[jenkins] fix: latest Transifex CLI requires a token

Wayonb:
problem: The latest Transifex CLI uses a token instead of a username/password
solution: update the publish to get an API token


Jaguar0625 (approved):
👍

PR #1447, opened 2025-01-29
[jenkins] fix: boost download URI update

Wayonb:
problem: Boost update their download URI
solution: update the Boost URI for the image build

https://github.com/boostorg/boost/issues/843#issuecomment-2567034014


Jaguar0625 (approved):
:+1:

PR #1457, opened 2025-02-06
[jenkins] fix: Jenkins no jobs run

Wayonb:
problem: in some instances, Jenkin(controller) thinks no job needs to be run
         this seems to be the case, especially on rebase.
solution: run all jobs when none is found.


AnthonyLaw (approved):
👍🏼

Jaguar0625 (approved):
👍

PR #1459, opened 2025-02-07
[jenkins] fix: add Gradle to the Java docker image

Wayonb:
problem: there is no gradle tool in Java image
solution: add Gradle to the Java docker image


Jaguar0625 (approved):
👍

PR #1460, opened 2025-02-07
[jenkins] fix: add support for Jacoco in Gradle

Wayonb:
problem: no Jacoco support for Gradle
solution: add Jacoco support for Gradle

### DIFF
--- a/jenkins/shared-library/vars/codeCoverage.groovy
+++ b/jenkins/shared-library/vars/codeCoverage.groovy
@@ -50,6 +50,10 @@ void verifyCodeCoverageResult(String tool, Integer minimumCodeCoverage) {
 			logger.logInfo("Minimum code coverage set in pom is ${readJacocoCoverageLimit()}")
 			runScript('mvn jacoco:check@jacoco-check')
 		},
+		'jacoco-gradle': { Integer target ->
+			runScript('gradle jacocoLogTestCoverage')
+			runScript('gradle jacocoTestCoverageVerification')
+		},
 		'golang': { Integer target ->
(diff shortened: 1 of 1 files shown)


jenkins/shared-library/vars/codeCoverage.groovy
```diff
@@ -50,6 +50,10 @@ void verifyCodeCoverageResult(String tool, Integer minimumCodeCoverage) {
 			logger.logInfo("Minimum code coverage set in pom is ${readJacocoCoverageLimit()}")
 			runScript('mvn jacoco:check@jacoco-check')
 		},
+		'jacoco-gradle': { Integer target ->
+			runScript('gradle jacocoLogTestCoverage')
+			runScript('gradle jacocoTestCoverageVerification')
+		},
```
AnthonyLaw, line 56: Does it need the target variable?
because I didn't see it being used.

Wayonb: yea, the variable is not used.  I guess I can rename it to __ as a placeholder.

Wayonb: I can't use "_" in the parameter name, or else the linter fails 😅 
https://codenarc.org/codenarc-rules-naming.html#parametername-rule

Which do you prefer
1. rename var from ``target`` to  ``notUsed``
2. leave it as is
3. update the default linting rules.

AnthonyLaw: let's go for option 2.

AnthonyLaw (approved):
👍🏼

PR #1467, opened 2025-02-11
[sdk/java] feat: initial Java SDK

Wayonb:
1. add Jenkins builds setup
2. add utils files for conversion, array helper, etc.


sdk/java/eclipse-formatter.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
Jaguar0625, line 1: i am not sure how much of the gradle configuration is per-project vs global, but the global configs should be moved into /linters and symlinked (like is done for python, javascript, etc)

sdk/java/src/main/java/org/symbol/sdk/utils/ArrayHelpers.java
```diff
@@ -0,0 +1,303 @@
...
+			final T element = factory.apply(buffer);
+
+			if (0 >= element.getSize())
+				throw new IllegalStateException("element size has invalid size");
+
+			if (null != accessor && null != previousElement && 0 <= deepCompare(accessor.apply(previousElement), accessor.apply(element)))
+				throw new IllegalStateException("elements in array are not sorted");
+
+			elements.add(element);
+			previousElement = element;
```
Jaguar0625, line 62: where is shift happening? in factory.apply?

sdk/java/src/main/java/org/symbol/sdk/utils/BufferView.java
```diff
@@ -0,0 +1,102 @@
...
+public class BufferView {
+	ByteBuffer bufferViewImpl;
+
+	BufferView(final byte[] buffer) {
+		this.bufferViewImpl = ByteBuffer.wrap(buffer).order(ByteOrder.LITTLE_ENDIAN);
+	}
+
+	BufferView(final ByteBuffer buffer) {
+		this.bufferViewImpl = buffer.slice();
+	}
```
Jaguar0625, line 18: should these be private ?

sdk/java/src/main/java/org/symbol/sdk/utils/Writer.java
```diff
@@ -0,0 +1,19 @@
...
+	 * Writes array into buffer.
+	 *
+	 * @param buffer Data to write.
+	 */
+	void write(final byte[] buffer);
+
+	/**
+	 * Get the underlining storage.
+	 */
+	byte[] getStorage();
```
Jaguar0625, line 18: why do you need this on interface?

sdk/java/src/main/java/org/symbol/sdk/utils/Serializer.java
```diff
@@ -0,0 +1,21 @@
+package org.symbol.sdk.utils;
+
+/**
+ * This interface serialize a catbuffer object.
+ */
+public interface Serializer {
```
Jaguar0625, line 6: serializable ?

.github/codeowners.md
```diff
@@ -6,7 +6,8 @@ catbuffer/schemas/		@Jaguar0625 @gimre-xymcity
 client/catapult/		@Jaguar0625 @gimre-xymcity
```
Jaguar0625, line 6: we might want (separate) PR where we alphabetize projects in all these files

Jaguar0625 (commented):
went through setup files and source

if there is anything specific in tests you want me to review, tag me on it as i assume they're mostly straightforward ports

sdk/java/src/test/java/org/symbol/sdk/utils/BufferWriterTest.java
```diff
@@ -0,0 +1,90 @@
...
+	@Test
+	public void canConstructWithEmptyBuffer() {
+		// Arrange:
+		final Writer writer = new BufferWriter(0);
+
+		// Act + Assert:
+		final Exception exception = assertThrows(Exception.class, () -> writer.write(new byte[]{
+				42
+		}));
+		// .to.throw('offset is out of bounds')
```
Jaguar0625, line 21: remove ?

sdk/java/src/test/java/org/symbol/sdk/utils/ArrayHelpersTest.java
```diff
@@ -0,0 +1,679 @@
...
+		}), equalTo(-1));
+		assertThat(ArrayHelpers.deepCompare(new Integer[]{
+				1, 12, 3, 4
+		}, new Integer[]{
+				1, 12, 3, 4
+		}), equalTo(0));
+		assertThat(ArrayHelpers.deepCompare(new Integer[]{
+				1, 12, 3, 4, 16
+		}, new Integer[]{
+				1, 15, 3
```
Jaguar0625, line 142: 1, 12, 3, 4 ?

sdk/java/src/test/java/org/symbol/sdk/utils/ArrayHelpersTest.java
```diff
@@ -0,0 +1,679 @@
...
+		final List<Serializer> elements = ArrayHelpers.readArray(context.subView, context::factory, element -> new Integer[]{
+				((MockElement) element).getTag()
+		});
+
+		// Assert:
+		assertThat(elements, equalTo(readArrayExpectedElement));
+	}
+
+	@Test
+	public void canReadArrayWhenUsingAccessorAndElementAreUnOrdered() {
```
Jaguar0625, line 343: cannot? (multiple)

sdk/java/src/test/java/org/symbol/sdk/utils/ArrayHelpersTest.java
```diff
@@ -0,0 +1,679 @@
...
+				}));
+		assertThat(thrown.getMessage(), equalTo("elements in array are not sorted"));
+	}
+
+	// endregion
+
+	// region ReadArrayCount
+
+	@Test
+	public void readArrayCountCanThrowsWhenAnyElementHasZeroSize() {
```
Jaguar0625, line 360: throw~~s~~

sdk/java/src/test/java/org/symbol/sdk/utils/ArrayHelpersTest.java
```diff
@@ -0,0 +1,679 @@
...
+		final IllegalArgumentException thrown = assertThrows(IllegalArgumentException.class,
+				() -> ArrayHelpers.readVariableSizeElements(context.subView, context::factory, 4, true));
+		assertThat(thrown.getMessage(), equalTo("newPosition > limit: (64 > 63)"));
+	}
+
+	// endregion
+
+	// region writers
+
+	// region writeArray
```
Jaguar0625, line 521: try to avoid nested regions as some tools don't support that

sdk/java/src/test/java/org/symbol/sdk/utils/Base32Test.java
```diff
@@ -0,0 +1,184 @@
...
+		{
+			put("68BA9E8D1AA4502E1F73DA19784B5D7DA16CA1E4AF895FAC12", "NC5J5DI2URIC4H3T3IMXQS25PWQWZIPEV6EV7LAS");
+			put("684C2605E5B366BB94BC30755EC9F50D74E80FC9283D20E283", "NBGCMBPFWNTLXFF4GB2V5SPVBV2OQD6JFA6SBYUD");
+			put("68D7B09A14BEA7CE060E71C0FA9AC9B4226DE167013DE10B3D", "NDL3BGQUX2T44BQOOHAPVGWJWQRG3YLHAE66CCZ5");
+			put("686C44C024F1089669F53C45AC6D62CC17A0D9CBA67A6205E6", "NBWEJQBE6EEJM2PVHRC2Y3LCZQL2BWOLUZ5GEBPG");
+			put("98A0FE84BBFC5EEE7CADC2B12F790DAA4A7A9505096E674FAB", "TCQP5BF37RPO47FNYKYS66INVJFHVFIFBFXGOT5L");
+		}
+	};
+
+	// encode
```
Jaguar0625, line 23: use // region // endregion

sdk/java/src/test/java/org/symbol/sdk/utils/BufferViewTest.java
```diff
@@ -0,0 +1,261 @@
...
+		});
+	}
+
+	@Test
+	public void canShiftToEnd() {
+		assertShift(5, new byte[0]);
+	}
+
+	@Test
+	public void canDoMultipleShift() {
```
Jaguar0625, line 119: shifts

sdk/java/src/test/java/org/symbol/sdk/utils/CharacterMapBuilderTest.java
```diff
@@ -0,0 +1,113 @@
...
+				put('e', (byte) 1);
+				put('f', (byte) 2);
+			}
+		};
+
+		// Act:
+		final Map<Character, Byte> charMap = builder.build();
+
+		// Assert:
+		assertThat("Character map should be empty", charMap, is(expectedMap));
```
Jaguar0625, line 40: message is wrong

Jaguar0625: isn't is identity ?

Jaguar0625: (multiple)

sdk/java/src/test/java/org/symbol/sdk/utils/ConverterTest.java
```diff
@@ -0,0 +1,210 @@
...
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
+import org.junit.Test;
+
+public class ConverterTest {
+
+	// toByte tests
```
Jaguar0625, line 15: use // region // endregion

Jaguar0625 (commented):
reviewed tests too

sdk/java/eclipse-formatter.xml
Wayonb: yea, I was thinking this but forgot to do it 😅

sdk/java/src/main/java/org/symbol/sdk/utils/ArrayHelpers.java
Wayonb: Yeah, this is one thing that's different from the Javascript SDK. The factory uses the ByteBuffer, which automatically shifts when reading a buffer. In some cases, the ByteBuffer will throw first instead of our checks when checking buffer limits.
I can change it if you want to have the swifts explicitly, but this is a little more efficient.

sdk/java/src/main/java/org/symbol/sdk/utils/Writer.java
Wayonb: Let me check.  I think I added it for tests, but let me check

sdk/java/src/main/java/org/symbol/sdk/utils/ArrayHelpers.java
Jaguar0625: it's ok, but maybe add a comment

sdk/java/eclipse-formatter.xml
Jaguar0625: is any of the gradle stuff common and can also be moved? (can do this in separate PR)

sdk/java/src/test/java/org/symbol/sdk/utils/ArrayHelpersTest.java
```diff
@@ -0,0 +1,676 @@
...
+		final IllegalStateException thrown = assertThrows(IllegalStateException.class,
+				() -> ArrayHelpers.readArrayCount(context.subView, context::factory, 3, element -> new Integer[]{
+						-((MockElement) element).getTag()
+				}));
+		assertThat(thrown.getMessage(), equalTo("elements in array are not sorted"));
+	}
+
+	// endregion
+
+	// read variable size
```
Jaguar0625, line 435: region

sdk/java/src/test/java/org/symbol/sdk/utils/CharacterMapBuilderTest.java
Jaguar0625: is `is` checking equals or identiy (same object)?

sdk/java/src/test/java/org/symbol/sdk/utils/BufferViewTest.java
```diff
@@ -0,0 +1,261 @@
...
+	@Test
+	public void cannotShrinkNegativeSubview() {
+		// Arrange:
+		final TestContext context = new TestContext();
+
+		// Act + Assert:
+		final IllegalArgumentException exception = assertThrows(IllegalArgumentException.class, () -> context.view.window(-1));
+		assertThat(exception.getMessage(), equalTo("size cannot be negative"));
+	}
+}
```
Jaguar0625, line 261: missing endregion

Jaguar0625 (commented):
👍

sdk/java/eclipse-formatter.xml
Wayonb: The gradle files are version-specific. If we plan to keep the same version for each project then I think we can move them to somewhere common. 🤔

Jaguar0625: i'd imagine we'd want to use the same gradle version monorepo-wide, or is there some reason we wouldn't?

Wayonb: Yeah, it makes sense.  Where would we want to put them?  Jenkins folder?

sdk/java/src/test/java/org/symbol/sdk/utils/CharacterMapBuilderTest.java
Wayonb: It is more for readability, but I will update them all to use ``equalTo``
```
Sugar

Hamcrest strives to make your tests as readable as possible. For example, the is matcher is a wrapper that doesn’t add any extra behavior to the underlying matcher. The following assertions are all equivalent:

assertThat(theBiscuit, equalTo(myBiscuit)); 
assertThat(theBiscuit, is(equalTo(myBiscuit))); 
assertThat(theBiscuit, is(myBiscuit));

The last form is allowed since is(T value) is overloaded to return is(equalTo(value)).
```

sdk/java/src/main/java/org/symbol/sdk/utils/Writer.java
Wayonb: This will also be needed for the model class, which serializes as in the [JS SDK](https://github.com/symbol/symbol/blob/445b82ede955b0185597e3f8f890529d13b791d7/sdk/javascript/src/nem/models.js#L786), since the buffer is allocated in the Writer class.

sdk/java/src/test/java/org/symbol/sdk/utils/CharacterMapBuilderTest.java
Jaguar0625: ok, in that case, just use either equalTo or is consistently everywhere

Jaguar0625 (approved):
good job 👍

PR #1479, opened 2025-02-20
[client/rest] fix: upgrade to zeromq 6.x

Wayonb:
problem: REST needs changes to work with Zeromq.js
solution: update REST to work with Zeromq


Jaguar0625 (approved):
👍

PR #1486, opened 2025-02-24
[sdk/javascript] fix: upgrade rustc to supports edition2024 feature

Wayonb:
problem: the latest version of hex-literal requires edition2024 feature
solution: upgrade to rustc 1.85, which supports edition2024 feature


Jaguar0625 (approved):
👍

PR #1489, opened 2025-02-28
[client/rest] fix: some issues reported by docker scout

Wayonb:
problem: Docker Scout reports issues with REST image such as no default Non-Root User
solution: upgrade to Ubuntu 24.04, which has a non-root user


Jaguar0625 (approved):
👍

PR #1507, opened 2025-03-12
[catbuffer/schemas]: fix NEM block previous_block_hash schema

Wayonb:
problem: previous_block_hash schema missing the object size
solution: update to type to SizePrefixedHash256

### DIFF
--- a/catbuffer/schemas/nem/block.cats
+++ b/catbuffer/schemas/nem/block.cats
@@ -15,6 +15,9 @@ struct Block
 
 	inline EntityBody
 
+	# previous block hash outer size
+	previous_block_hash_outer_size = make_reserved(uint32, 36)
+
 	# previous block hash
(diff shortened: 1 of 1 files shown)


catbuffer/schemas/nem/block.cats
```diff
@@ -16,7 +16,7 @@ struct Block
 	inline EntityBody
 
 	# previous block hash
-	previous_block_hash = Hash256
+	previous_block_hash = inline SizePrefixedHash256
```
Jaguar0625, line 19: isn't there an extra size somewhere? 🤔

Wayonb: yeah, you are right.  I forgot to add it. 😅

Jaguar0625 (approved):
👍

PR #1508, opened 2025-03-13
[jenkins] fix: add docker image for docs

Wayonb:
problem: no docker image for docs
solution: add Dockerfile for mkdocs project

### DIFF
--- a/jenkins/docker/ubuntu/docs.Dockerfile
+++ b/jenkins/docker/ubuntu/docs.Dockerfile
@@ -42,12 +42,6 @@ RUN curl -Os https://www.doxygen.nl/files/doxygen-${DOXYGEN_VERSION}.src.tar.gz 
 	&& cd ../.. \
 	&& rm -rf doxygen-${DOXYGEN_VERSION} doxygen-${DOXYGEN_VERSION}.src.tar.gz
 
-# codecov uploader
-RUN ARCH=$([ "$(uname -m)" = "x86_64" ] && echo "linux" || echo "aarch64") \
-	&& curl -Os "https://uploader.codecov.io/latest/${ARCH}/codecov" \
-	&& chmod +x codecov \
-	&& mv codecov /usr/local/bin
-
 # add ubuntu user (used by jenkins)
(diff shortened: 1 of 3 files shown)


segfaultxavi (approved):
I don't understand most of these changes, but they do seem to install the deps I need, so I approve.

jenkins/docker/ubuntu/docs.Dockerfile
```diff
@@ -0,0 +1,71 @@
...
+
+# install python
+RUN apt-get install -y python3-pip python3-venv
+
+# install shellcheck
+RUN apt-get install -y shellcheck
+
+# nodejs
+ENV NODE_OPTIONS="--dns-result-order=ipv4first"
+ARG NODEJS_VERSION=22
```
Jaguar0625, line 19: should we use same nodejs version as javascript dockerfile?

jenkins/docker/ubuntu/docs.Dockerfile
```diff
@@ -0,0 +1,71 @@
...
+	&& mkdir build \
+	&& cd build \
+	&& cmake -G "Unix Makefiles" .. \
+	&& cmake -Dbuild_wizard=YES .. \
+	&& make \
+	&& make install \
+	&& cd ../.. \
+	&& rm -rf doxygen-${DOXYGEN_VERSION} doxygen-${DOXYGEN_VERSION}.src.tar.gz
+
+# codecov uploader
```
Jaguar0625, line 45: do you need codecov?

jenkins/docker/ubuntu/docs.Dockerfile
Wayonb: I will leave the same version, but they will all get bumped to next month.

jenkins/docker/ubuntu/docs.Dockerfile
Wayonb: Yeah, it probably can be removed, but I wasn't sure.

Jaguar0625 (approved):
👍

PR #1514, opened 2025-03-20
[client/rest] fix: catapultdb initialized with incorrect network

Wayonb:
problem: catapultDB created an incorrect network, which caused the address from publicKey to be incorrect
                 this caused search by publicKey to create an incorrect address, which would fail.
solution: create the network correctly


Jaguar0625 (approved):
👍

PR #1529, opened 2025-03-31
[jenkins] fix: install node-canvas dependencies in Javascript image

OlegMakarenko:
problem: Javascript image does not support node-canvas module
solution: install node-canvas dependencies during Javascript image creation


Wayonb (approved):
👍🏾

PR #1533, opened 2025-04-04
[jenkins] fix: add twine to Python venv

Wayonb:
problem: twine ships with Ubuntu is older version(v5.0) and does not support newer metadata
solution: install newer version of twine(v6.1) into Python venv


Jaguar0625 (approved):
👍

PR #1534, opened 2025-04-04
[jenkins] fix: expose the python internal repository

Wayonb:
problem: projects using the custom publish script need Python internal repo
solution: expose the Python internal repository using CI


Jaguar0625 (approved):
👍

PR #1550, opened 2025-04-15
[jenkins] fix: cpp ci images fails due to permissions

Wayonb:
problem: cpp ci image fails to build due to permissions
solution: change to root user before installing apps


AnthonyLaw (approved):
👍🏼

PR #1576, opened 2025-05-09
[jenkins] fix: libzmq build fails with CMake 4

Wayonb:
problem: libzmq CMake version requirement is not compatible with CMake 4
solution: disable the CMake requirement by adding -DCMAKE_POLICY_VERSION_MINIMUM=3.5


AnthonyLaw (approved):
👍🏼

Jaguar0625 (approved):
👍

PR #1581, opened 2025-05-15
[catapult/client] : Align Vcpkg build to Conan build

AndreaLanfranchi:
This addresses Windows builds using Vcpkg instead of Conan

- Bumps dependencies to same versions as in conanfile.py
- Corrects from `Ninja` generator to 'Visual Studio 17 2022 Win64'


client/catapult/CMakeSettings.json
```diff
@@ -31,4 +31,4 @@
       "cmakeToolchain": "${env.VCPKG_ROOT}\\scripts\\buildsystems\\vcpkg.cmake"
     }
   ]
-}
+}
```
Wayonb: missing empty blank line

Wayonb (approved):
👍🏾

client/catapult/CMakeSettings.json
AndreaLanfranchi: Yes thank you for noticing.
Unfortunately it gets removed every time you use VS GUI to update the Cmake Settings.
I wouldn't bother much on that although, I agree, is disturbing.

PR #1597, opened 2025-06-03
[sdk/javascript] fix: pin rustc to 1.86.0

Wayonb:
problem: JS SDK fails to build with Rustc 1.87.0
solution: pin the Rustc version to 1.86.0


Jaguar0625 (approved):
👍

PR #1599, opened 2025-06-06
[client/rest] fix: websocket transaction missing maxFee

Wayonb:
problem: SDK serial transaction with fee property vs maxFee used in REST.
         This caused the property to drop for the websocket.
solution: fix up the transaction by mapping fee -> maxFee


Jaguar0625 (approved):
👍

AnthonyLaw (approved):
👍🏼

PR #1643, opened 2025-07-24
[client/catapult] fix : gracefully deal with CLI options parsing

AndreaLanfranchi:
Calling any tool with wrongly formatted or unexpected arguments causes the parser to throw an exception which remains unhandled eventually causing the program to abort. (In Windows in debug mode also a dialog may appear)
Example 
```
catapult.tools.addressgen.exe -help [enter]

[2025-07-24 11:47:06.189031] [0x00001c0c] [info]
thread: Tool Main
unhandled exception while running a tool!
Throw location unknown (consider using BOOST_THROW_EXCEPTION)
Dynamic exception type: struct boost::wrapexcept<class boost::program_options::unknown_option>
std::exception::what: unrecognised option '-help'
```

This PR fixes the behavior and gracefully terminates the program when needed.

```
catapult.tools.addressgen.exe -help [enter]

Address Generator Tool
Copyright (c) Jaguar0625, gimre, BloodyRookie, Tech Bureau, Corp.
catapult version: 1.0.3.8 01d1bd4eb [dev]

Error parsing command line options: unrecognised option '-help'

Try using [-h|--help] option.
```


client/catapult/tools/tools/ToolMain.cpp
```diff
@@ -184,7 +184,15 @@ namespace catapult { namespace tools {
 
 		// 2. parse all options
 		ParsedOptions options;
-		ParseOptions(options, tool, argc, argv);
+		try {
+			ParseOptions(options, tool, argc, argv);
+		} catch (const std::exception& e) {
+			std::cout << "Error parsing command line options: " << e.what() << "\nTry using [-h|--help] option." << std::endl;
+			return -1;
+		} catch (...) {
```
Jaguar0625: not sure if we should catch ... here vs letting it flow upstream. any parsing exceptions should be derived from std::exception.

client/catapult/tools/tools/ToolMain.cpp
```diff
@@ -184,7 +184,15 @@ namespace catapult { namespace tools {
 
 		// 2. parse all options
 		ParsedOptions options;
-		ParseOptions(options, tool, argc, argv);
+		try {
+			ParseOptions(options, tool, argc, argv);
+		} catch (const std::exception& e) {
+			std::cout << "Error parsing command line options: " << e.what() << "\nTry using [-h|--help] option." << std::endl;
```
Jaguar0625, line 190: generally, we've been using std::endl when outputting into stream instead of '\n' literals

AndreaLanfranchi: std::endl is a bit more than simple new line: it also forces a flush of the stream.
On the other hand `\n` clearly depicts a new line only.

client/catapult/tools/tools/ToolMain.cpp
AndreaLanfranchi: Actually there's no upstream ... erm ... well a very limited one like 
```
int main(int argc, const char** argv) {
	catapult::tools::addressgen::AddressGeneratorTool tool;
	return catapult::tools::ToolMain(argc, argv, tool);
}
```
where there's no exception handling. (apparently all tools replicate the same launch structure)

AndreaLanfranchi: Actually the catch _any_ is due to me being lazy investigating all possible exception types thrown by boost options (yes they _should_ all inherit from std::exception but ... you never know).

Jaguar0625: it looks like everything derives from std::logic_error https://www.boost.org/doc/libs/latest/boost/program_options/errors.hpp, but first catch is fine (std::exception). if anything is thrown that is NOT std::exception, it's unlikely to be a command line parsing issue. so, i'd drop second (...) catch.

client/catapult/tools/tools/ToolMain.cpp
Jaguar0625: that's true, no strong opinion on my part

client/catapult/tools/tools/ToolMain.cpp
```diff
@@ -184,7 +184,15 @@ namespace catapult { namespace tools {
...
+			return -1;
+		} catch (...) {
+			std::cout << UNHANDLED_EXCEPTION_MESSAGE("parsing command line options") << "\nTry using [-h|--help] option." << std::endl;
+			return -1;
```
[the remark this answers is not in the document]
AndreaLanfranchi: Won't do: the scope is limited and a serious impl would require a refactor

client/catapult/tools/tools/ToolMain.cpp
AndreaLanfranchi: Just removed the catch any block.

I also changed the return code for "help request" to negative to differentiate successful tool execution (which conventionally should return a value ge 0) from failures (negative values). Before this, in fact, on [verify tool](https://github.com/symbol/symbol/blob/dev/client/catapult/tools/verify/main.cpp#L358-L359) one possible return value is 1 (same as request help).

Jaguar0625 (approved):
👍

PR #1659, opened 2025-08-07
[sdk/javascript] fix: add TypeScript 5.9 support

Wayonb:
problem: ArrayBuffer changed where it is no longer a supertype to.
                 This includes subtypes of UInt8Array, such as Buffer from Node.js
https://devblogs.microsoft.com/typescript/announcing-typescript-5-9/#lib.d.ts-changes
solution: specify a more specific underlying buffer type instead of using the default


Jaguar0625 (approved):
👍

PR #1727, opened 2025-10-10
[jenkins] fix: deps and macOS build failing

Wayonb:
problem: OpenSSL and ZeroMQ are failing to build on macOS with CMake 4.
solution: ZeroMQ set the CMAKE_POLICY_VERSION_MINIMUM=3.5
          MacOS build set the CMAKE_OSX_SYSROOT so as not to use the system include
          OpenSSL Configure scripts now autodetect OS correctly


Jaguar0625 (approved):
👍 

in commit message, please add blank line after title (first) line or `git log --oneline` will process whole commit message as single line

PR #1737, opened 2025-10-13
[mkdocs] feat: add monitoring status tutorial

zero4862:
Adds tutorial for monitoring transaction status based on [Querying Unconfirmed Transactions]( https://github.com/symbol/symbol/blob/main/docs/index.md#tutorial-querying-unconfirmed-transactions) code snippet.


mkdocs/overrides/devbook/transactions/monitoring-status.py
```diff
@@ -0,0 +1,86 @@
+import json
+import os
+import time
+import urllib.request
+
+# Configuration
+NODE_URL = os.environ.get(
+	"NODE_URL", "http://ngl-dual-101.testnet.symboldev.network:3000"
```
segfaultxavi: This node is not responding anymore. Can you use the same node as the Transfer tutorial?
https://001-sai-dual.symboltest.net:3001
I think we are maintaining this one, so it has more guarantees to keep operating.

segfaultxavi: Also, can you print the URL, just like the Transfer tutorial does?

mkdocs/overrides/devbook/transactions/monitoring-status.py
```diff
@@ -0,0 +1,86 @@
...
+)
+
+# Transaction hash to monitor
+# In a real application, you would get this hash after announcing a transaction
```
segfaultxavi: When you look at the generated docs, you can see that this line (and line 82) exceed the code box width and require scrolling to read the whole content of the line.
I want to avoid that, so please make these lines shorter.

In practice, this means that the max line length for source code files is closer to 78 or 77 chars.

mkdocs/overrides/devbook/transactions/monitoring-status.py
```diff
@@ -0,0 +1,86 @@
...
+				if status_group == "confirmed":
+					print(f"\nTransaction confirmed!")
+					return True
+
+				# Check if the transaction failed
+				if status_group == "failed":
+					print(f"\nTransaction failed with code: {status_code}")
+					raise RuntimeError(f"Transaction failed: {status_code}")
+
+		except urllib.error.HTTPError as e:
```
segfaultxavi: Can you also catch `urllib.error.URLError`, which is what you get when the node is offline (or user mistyped the node URL), and is therefore a common enough error?

mkdocs/pages/en/devbook/transactions/monitoring-status.md
```diff
@@ -0,0 +1,192 @@
...
+
+    === "Python"
+        ```python
+        --8<-- 'devbook/transactions/getting-transaction-hash.py'
+        ```
+
+    === "JavaScript"
+        ```javascript
+        --8<-- 'devbook/transactions/getting-transaction-hash.mjs'
+        ```
```
segfaultxavi: I think you only need the last sentence ("See the Transfer tutorial..."). Adding the info box with a snippet here is a bit too much, and it means more code that we need to maintain. You can remove the box entirely (and the source files).

mkdocs/pages/en/devbook/transactions/monitoring-status.md
```diff
@@ -0,0 +1,192 @@
...
+### Setting Up the Transaction Hash
+
+{{ tutorial.code_snippet(['py:13:16', 'js:9:10']) }}
+
+To monitor a transaction, you need its hash, which is generated after signing.
+The hash uniquely identifies the transaction on the Symbol network.
+
+For this tutorial, we use a sample transaction hash to demonstrate the monitoring.
+In a real application, you would obtain this hash immediately after signing a transaction and use it to track its
+status:
```
segfaultxavi:
```suggestion
status.
```

mkdocs/pages/en/devbook/transactions/monitoring-status.md
```diff
@@ -0,0 +1,192 @@
...
+        ```python
+        --8<-- 'devbook/transactions/getting-transaction-hash.py'
+        ```
+
+    === "JavaScript"
+        ```javascript
+        --8<-- 'devbook/transactions/getting-transaction-hash.mjs'
+        ```
+
+    See the [Transfer](transfer.md) tutorial for a complete example of creating, signing, and announcing transactions.
```
segfaultxavi:
```suggestion
The [Transfer](./transfer.md) tutorial shows how to obtain a transaction's hash after signing it.
```

mkdocs/pages/en/devbook/transactions/monitoring-status.md
```diff
@@ -0,0 +1,192 @@
...
+## Code Explanation
+
+### Setting Up the Transaction Hash
+
+{{ tutorial.code_snippet(['py:13:16', 'js:9:10']) }}
+
+To monitor a transaction, you need its hash, which is generated after signing.
+The hash uniquely identifies the transaction on the Symbol network.
+
+For this tutorial, we use a sample transaction hash to demonstrate the monitoring.
```
segfaultxavi: We never talk about ourselves in the docs, so no "we", "us" or "our". You can use passive voice if needed, but in this case it's simpler to just say:

```suggestion
This tutorial uses a sample transaction hash to demonstrate the monitoring.
```

mkdocs/pages/en/devbook/transactions/monitoring-status.md
```diff
@@ -0,0 +1,192 @@
...
+# Monitoring Transaction Status
+
+After announcing a <transaction:> to the Symbol network, it remains unconfirmed until it is included in a <block:>.
+
+Monitoring status changes is essential for building responsive applications that can react to transaction
+confirmation or failure.
+
+This tutorial shows how to monitor a transaction's status as it moves from unconfirmed to confirmed.
+
+!!! important "Confirmed transactions can still be reversed"
```
segfaultxavi: `important` is not a valid admonition type in the Material theme, so it renders as a `note` instead:
https://squidfunk.github.io/mkdocs-material/reference/admonitions/#supported-types
Maybe we can use `info` here? I feel like a `warning` would be too scary right in the introduction.

mkdocs/pages/en/devbook/transactions/monitoring-status.md
```diff
@@ -0,0 +1,192 @@
...
+    A confirmed transaction has been included in a block but is not yet irreversible.
+	The final state is <finalization:>, which occurs only after the block is finalized by the network.
+	Until then, <rollback:|rollbacks> are still possible.
```
segfaultxavi:
```suggestion
	Until then, <rollbacks:> are still possible.
```

The glossary plugin is smart enough to detect English plurals, so you can simplify this link.

mkdocs/pages/en/devbook/transactions/monitoring-status.md
```diff
@@ -0,0 +1,192 @@
...
+If you have not done so already, start with the [Hello World](../start/hello-world.md) tutorial to make sure your
+development environment is set up correctly.
+
+## Full Code
+
+{% import 'tutorial.jinja2' as tutorial with context %}
+
+{{ tutorial.code_full('devbook/transactions/monitoring-status', ['py', 'js']) }}
+
+!!! note "Polling vs WebSockets"
```
segfaultxavi: This is a good comment. Let's put it before the code.
And make it normal text instead of an info box, to avoid having too many boxes :D

mkdocs/pages/en/devbook/transactions/monitoring-status.md
```diff
@@ -0,0 +1,192 @@
...
+    This tutorial uses polling to check transaction status.
```
segfaultxavi:
```suggestion
    This tutorial uses polling to check the transaction status.
```

mkdocs/pages/en/devbook/transactions/monitoring-status.md
```diff
@@ -0,0 +1,192 @@
...
+    Polling is used here for illustration purposes, but it is not the recommended approach for production applications.
+
+    A production-grade application should use WebSockets to receive status change notifications directly from the <node:>.
+    This provides a more responsive solution without the overhead of repeated API calls.
```
segfaultxavi: "Production-grade" is a bit ambiguous... the fact is that websockets are more responsive and efficient, so let's just say that. Something like:

```suggestion
    WebSockets provide a more responsive solution without the overhead of repeated API calls.
```

mkdocs/pages/en/devbook/transactions/monitoring-status.md
```diff
@@ -0,0 +1,192 @@
...
+
+{{ tutorial.code_snippet(['py:13:16', 'js:9:10']) }}
+
+To monitor a transaction, you need its hash, which is generated after signing.
+The hash uniquely identifies the transaction on the Symbol network.
+
+For this tutorial, we use a sample transaction hash to demonstrate the monitoring.
+In a real application, you would obtain this hash immediately after signing a transaction and use it to track its
+status:
+
```
segfaultxavi: You also need to mention that the tutorial allows you to provide your own hash via the `TRANSACTION_HASH` environment variable.
It will be obvious to most people, sure, but it looks like we forgot about it if we don't mention it.

mkdocs/pages/en/devbook/transactions/monitoring-status.md
```diff
@@ -0,0 +1,192 @@
...
+        ```
+
+    === "JavaScript"
+        ```javascript
+        --8<-- 'devbook/transactions/getting-transaction-hash.mjs'
+        ```
+
+    See the [Transfer](transfer.md) tutorial for a complete example of creating, signing, and announcing transactions.
+
+### The Monitoring Function
```
segfaultxavi: There is no need for a section about the function. The code gets repeated too much then, and the actual useful snippets get buried in a level-4 header that does not show in the TOC.

Move the explanation of the loop to the section about "Querying the status", promote all subsections to level-3 headers, and remove this section.

mkdocs/pages/en/devbook/transactions/monitoring-status.md
```diff
@@ -0,0 +1,192 @@
...
+between attempts).
+This loop structure ensures that monitoring will eventually stop, even if the transaction never confirms.
+
+Let's break down how this function works:
+
+#### Querying the Status Endpoint
+
+{{ tutorial.code_snippet(['py:35:56', 'js:28:57']) }}
+
+On each attempt, the function queries the <get:/transactionStatus/{hash}> endpoint, which returns real-time information
```
segfaultxavi:
```suggestion
On each attempt, the function queries the <get:/transactionStatus/{hash}> endpoint, which returns information
```

mkdocs/pages/en/devbook/transactions/monitoring-status.md
```diff
@@ -0,0 +1,192 @@
...
+about the transaction's current state.
+
+The response includes:
+
+* **Group:** The transaction's current status group. Possible values:
+    * `unconfirmed`: The transaction is in the unconfirmed pool waiting to be included in a block.
+    * `confirmed`: The transaction has been included in a block.
+    * `failed`: The transaction failed validation and was rejected.
+    * `partial`: For <bonded aggregate transaction:|bonded aggregate transactions> waiting
+		for <cosignature:|cosignatures>.
```
segfaultxavi: I believe it looks nicer as a table. For tables I disregard the 120-column limit.
```suggestion

    | Group         | Meaning                                                                          |
    | ------------- | -------------------------------------------------------------------------------- |
    | `unconfirmed` | The transaction is in the <unconfirmed pool:> waiting to be included in a block. |
    | `confirmed`   | The transaction has been included in a block.                                    |
    | `failed`      | The transaction failed validation and has been rejected.                         |
    | `partial`     | For <bonded aggregate transactions:> waiting for <cosignatures:>.                |

```

mkdocs/pages/en/devbook/transactions/monitoring-status.md
```diff
@@ -0,0 +1,192 @@
...
+!!! note "Polling vs WebSockets"
+    This tutorial uses polling to check transaction status.
+    Polling is used here for illustration purposes, but it is not the recommended approach for production applications.
+
+    A production-grade application should use WebSockets to receive status change notifications directly from the <node:>.
+    This provides a more responsive solution without the overhead of repeated API calls.
+
+## Code Explanation
+
+### Setting Up the Transaction Hash
```
segfaultxavi: This is extremely picky, but when possible, I like that the entries in the TOC fit in one line. For example:

```suggestion
### Finding the Transaction Hash
```

mkdocs/pages/en/devbook/transactions/monitoring-status.md
```diff
@@ -0,0 +1,192 @@
...
+{{ tutorial.code_snippet(['py:77:79', 'js:84:89']) }}
+
+Between polling attempts, the function waits for a configurable delay (default: 2 seconds).
+This prevents overwhelming the <node:> with requests and allows time for network processing.
+
+#### Handling Timeouts
+
+{{ tutorial.code_snippet(['py:81:82', 'js:92:97']) }}
+
+If the transaction doesn't confirm after all attempts, the function raises a `RuntimeError`.
```
segfaultxavi:
```suggestion
If the transaction is not confirmed after the specified attempts, the function raises a `RuntimeError` explaining the problem.
```

mkdocs/pages/en/devbook/transactions/monitoring-status.md
```diff
@@ -0,0 +1,192 @@
...
+To see the transaction from the network's perspective, visit the
+[Symbol Testnet Explorer](https://testnet.symbol.fyi/) and search for the transaction hash.
+
+## Conclusion
+
+This tutorial showed how to:
+
+| Step                                                                           | Related documentation                   |
+| ------------------------------------------------------------------------------ | --------------------------------------- |
+| [Set up the transaction hash](#setting-up-the-transaction-hash)                | <dy:SymbolFacade.signTransaction><br><dy:SymbolTransactionFactory.attachSignature> |
```
segfaultxavi: The tutorial no longer uses these methods, better remove this row.

mkdocs/pages/en/devbook/transactions/monitoring-status.md
```diff
@@ -0,0 +1,192 @@
...
+| [Query the status endpoint](#querying-the-status-endpoint)                     | <get:/transactionStatus/{hash}> |
+| [Check for confirmation and failure](#checking-for-confirmation)               | [TransactionStatusEnum](../reference/rest/symbol.md#model-TransactionStatusEnum) |
+| [Query block height](#output)                                                  | <get:/transactions/confirmed/{transactionId}> |
```
segfaultxavi: Yeah, this has been hinted, but not really explained. Better remove this row too.

segfaultxavi (changes requested):
Some comments apply to both source files.

Please remove all remaining tabs from the markdown file.

mkdocs/overrides/devbook/transactions/monitoring-status.py
zero4862: Good catch. Updated it to catch the generic Exception to handle any error.

mkdocs/pages/en/devbook/transactions/monitoring-status.md
zero4862: Applied, updated [transfer.md](https://github.com/symbol/symbol/pull/1737/commits/ed9e362eba358daf9608c7e090f09de66eece299#diff-afc8aada36f2a6c0450ae4fb5796b65ee88c700b9c4e82bdf842cb3dfad86e1e) admonition too.

mkdocs/pages/en/devbook/transactions/monitoring-status.md
zero4862: Applied.

thought: what do you think about using bare infinitive instead of gerund for section titles? titles would be shorter, more in line with the conclusion section, and create a stronger sense of action.

mkdocs/pages/en/devbook/transactions/monitoring-status.md
```diff
@@ -0,0 +1,178 @@
...
+{{ tutorial.code_snippet(['py:77:79', 'js:82:87']) }}
+
+Between polling attempts, the function waits for a configurable delay (default: 2 seconds).
+This prevents overwhelming the <node:> with requests and allows time for network processing.
+
+### Handling Timeouts
+
+{{ tutorial.code_snippet(['py:81:82', 'js:90:95']) }}
+
+If the transaction is not confirmed after the specified attempts, the function raises a `RuntimeError` explaining the problem.
```
segfaultxavi:
```suggestion
If the transaction is not confirmed after the specified number of attempts, the function raises a `RuntimeError` explaining the problem.
```

And I know this was _my_ suggestion 😅

mkdocs/pages/en/devbook/transactions/monitoring-status.md
```diff
@@ -0,0 +1,178 @@
...
+| ------------------------------------------------------------------------------ | --------------------------------------- |
+| [Find the transaction hash](#finding-the-transaction-hash)                     |                                         |
+| [Query the status endpoint](#querying-the-status-endpoint)                     | <get:/transactionStatus/{hash}>         |
+| [Check for confirmation](#checking-for-confirmation)                           | [TransactionStatusEnum](../reference/rest/symbol.md#model-TransactionStatusEnum) |
+| [Check for failure](#checking-for-failure)                                     | [TransactionStatusEnum](../reference/rest/symbol.md#model-TransactionStatusEnum) |
+| [Handle unknown status](#handling-unknown-status)                              |                                         |
+| [Wait between attempts](#waiting-between-attempts)                             |                                         |
+| [Handle timeouts](#handling-timeouts)                                          |                                         |
```
segfaultxavi:
```suggestion
| Step                                                       | Related documentation                                                            |
| ---------------------------------------------------------- | -------------------------------------------------------------------------------- |
| [Query the status endpoint](#querying-the-status-endpoint) | <get:/transactionStatus/{hash}>                                                  |
| [Check for confirmation](#checking-for-confirmation)       | [TransactionStatusEnum](../reference/rest/symbol.md#model-TransactionStatusEnum) |
| [Check for failure](#checking-for-failure)                 | [TransactionStatusEnum](../reference/rest/symbol.md#model-TransactionStatusEnum) |
```

I want these tables to be like a reference sheet for experienced developers. So if they want to know how to check the status, they come here, copy the endpoint and that's it.
So no need to list all the tutorial's sections without any further information. That's the TOC's job :D

mkdocs/overrides/devbook/transactions/monitoring-status.py
```diff
@@ -0,0 +1,86 @@
...
+				)
+			else:
+				raise
+
+		# Wait before next attempt (except on last attempt)
+		if attempt < max_attempts:
+			time.sleep(wait_seconds)
+
+	print(f"\nTransaction not confirmed after {max_attempts} attempts")
+	raise RuntimeError(f"Transaction {transaction_hash} not confirmed in time")
```
segfaultxavi: This is now the only line that's too long, the JS file is fine.

mkdocs/pages/en/devbook/transactions/monitoring-status.md
segfaultxavi: Hmmmm... I've tried it, and it sounds a bit too bossy for my taste. It reads like we are giving commands to the reader.
I prefer a more humble approach, where we just describe what we are doing, and the reader decides if he does the same or not.
So we stick with gerunds!

segfaultxavi (changes requested):
Almost done!

mkdocs/pages/en/devbook/transactions/monitoring-status.md
zero4862: Got it 👌

segfaultxavi (approved):

PR #1738, opened 2025-10-13
[mkdocs] feat: send messages tutorial

zero4862:
Adds a tutorial for sending transfer transactions (simple and encrypted), based on the code snippet from [Working with encrypted messages](https://github.com/symbol/symbol/blob/main/docs/index.md#tutorial-working-with-encrypted-messages).


mkdocs/pages/en/devbook/transactions/transfer.md
```diff
@@ -139,7 +141,7 @@ transaction's size in bytes, which is only known once the descriptor has been co
 !!! info "Including a message in the transaction"
 
     Transactions can optionally include a free-form message.
-    A later tutorial explains how to do this.
+    [This tutorial](messages.md) explains how to do this.
```
segfaultxavi:
```suggestion
    The [Sending Messages with Transfer Transactions](./messages.md) tutorial explains how to do this.
```

Always try to provide some preview of where the link goes, so avoid [this], [here] and similar.

mkdocs/overrides/devbook/transactions/messages.mjs
```diff
@@ -0,0 +1,166 @@
...
+	MessageEncoder
+} from 'symbol-sdk/symbol';
+
+// Configuration
+const NODE_URL = process.env.NODE_URL ||
+	'https://001-sai-dual.symboltest.net:3001';
+
+// Set up sender and recipient accounts
+const facade = new SymbolFacade('testnet');
+const senderPrivateKey = process.env.PRIVATE_KEY ||
```
segfaultxavi: For symmetry, maybe call the env var `SENDER_PRIVATE_KEY`.
And the variable should probably be called `senderPrivateKeyString`, to mimic the recipient's.

mkdocs/overrides/devbook/transactions/messages.mjs
```diff
@@ -0,0 +1,164 @@
...
+console.log('Plain message:',
+	new TextDecoder().decode(plainMessage));
+
+// Build transfer transaction with plain message
+let transaction = facade.transactionFactory.create({
+	type: 'transfer_transaction_v1',
+	signerPublicKey: senderKeyPair.publicKey.toString(),
+	deadline: timestamp.addHours(2).timestamp,
+	recipientAddress: recipientAddress.toString(),
+	mosaics: [{
```
segfaultxavi: Don't send any currency, let's not mix tutorials! :D

mkdocs/overrides/devbook/transactions/messages.mjs
```diff
@@ -0,0 +1,164 @@
...
+
+let announceResponse = await fetch(`${NODE_URL}/transactions`, {
+	method: 'PUT',
+	headers: { 'Content-Type': 'application/json' },
+	body: jsonPayload
+});
+console.log('Plain message transaction announced\n');
+
+// Wait a moment before sending the next transaction
+await new Promise(resolve => setTimeout(resolve, 2000));
```
segfaultxavi: Why do you need to wait? This will make readers think that the wait is necessary.

mkdocs/overrides/devbook/transactions/messages.mjs
```diff
@@ -0,0 +1,164 @@
...
+
+// ===== ENCRYPTED MESSAGE =====
+console.log('=== Sending Encrypted Message ===');
+
+// Fetch updated network time
```
segfaultxavi: No need to fetch the time again. Reuse the previous deadline.

mkdocs/overrides/devbook/transactions/messages.mjs
```diff
@@ -0,0 +1,164 @@
...
+console.log('Transaction hash:', transactionHash);
+
+announceResponse = await fetch(`${NODE_URL}/transactions`, {
+	method: 'PUT',
+	headers: { 'Content-Type': 'application/json' },
+	body: jsonPayload
+});
+console.log('Encrypted message transaction announced\n');
+
+// ===== DECRYPTING MESSAGE =====
```
segfaultxavi: I think it would be more illustrative if we simulated both sender and receiver, so the transactions are recovered from the network using their hashes and then the messages decoded.
Of course then the code needs both private keys and the text needs to explain why.

No need to track the confirmation status like the other tutorials, I think just polling `/transactions/confirmed` until the hash is found will be enough.

mkdocs/overrides/devbook/transactions/messages.mjs
```diff
@@ -0,0 +1,164 @@
...
+	console.log('Sender failed to decrypt message');
+}
+
+// The recipient would decrypt using their private key:
+// const recipientEncoder = new MessageEncoder(recipientKeyPair);
+// const result = recipientEncoder.tryDecode(
+//	senderKeyPair.publicKey, encryptedPayload
+// );
+
+console.log('\nBoth transactions have been sent successfully!');
```
segfaultxavi: Instead of this message, which might not be true because we haven't checked, you can say "Messages received and decoded successfully".

mkdocs/overrides/devbook/transactions/messages.mjs
```diff
@@ -0,0 +1,164 @@
...
+	method: 'PUT',
+	headers: { 'Content-Type': 'application/json' },
+	body: jsonPayload
+});
+console.log('Encrypted message transaction announced\n');
+
+// ===== DECRYPTING MESSAGE =====
+console.log('=== Decrypting Message ===');
+
+// Sender can decrypt using recipient's public key
```
segfaultxavi: Yeah, but this is weird use case. We want to show the code that's currently commented-out in line 158.

mkdocs/pages/en/devbook/transactions/messages.md
```diff
@@ -0,0 +1,151 @@
...
+can read them.
+
+This tutorial shows how to send both plain and encrypted messages, and demonstrates how to decrypt received messages.
+
+## Prerequisites
+
+If you have not done so already, start with the [Hello World](../start/hello-world.md) tutorial to make sure your
+development environment is set up correctly.
+
+Additionally, check the [Transfer transaction](transfer.md) tutorial to understand how fee
```
segfaultxavi:
```suggestion
Additionally, check the [Transfer transaction](./transfer.md) tutorial to understand how fee
```

It's not critical, but in the past I've used a plugin that automatically searched for files if they had not path.
I'm not using it now, but just in case I'll ever bring it back, always include the path in file names, even if it's just `./`

mkdocs/pages/en/devbook/transactions/messages.md
```diff
@@ -0,0 +1,151 @@
...
+
+The tutorial uses two accounts: a sender and a recipient.
+The code reads the sender's private key from the `PRIVATE_KEY` environment variable and the recipient's public key from
+the `RECIPIENT_PUBLIC_KEY` environment variable.
+Both variables default to test keys if not set.
+
+In a real application, you typically have the recipient's address rather than their public key.
+To encrypt messages, you need to obtain their public key by querying the network using the <get:/accounts/{accountId}>
+endpoint.
+The public key is only available for accounts that have previously sent transactions.
```
segfaultxavi: The reader and the docs might disagree on what "a real application" is :)
Let's stick to the facts. Also, this looks like a nice opportunity for an info box. How about:
```suggestion
!!! note
    A recipient's <address:> is sufficient to send a message, but encryption requires their <public key:>.
    When only the address is known, the public key can be retrieved from the network using the <get:/accounts/{accountId}> endpoint.
    An account's public key becomes available only after it has broadcast at least one transaction.
```

mkdocs/pages/en/devbook/transactions/messages.md
```diff
@@ -0,0 +1,151 @@
...
+
+### Sending a Plain Text Message
+
+{{ tutorial.code_snippet(['py:45:108', 'js:34:91']) }}
```
segfaultxavi: This snippet is way too long. We don't need to show all the code, just the parts relevant to the topic of the tutorial.

You can leave out the parts about fetching network time and fees, and in the text you can send readers to the Transfer tutorial.
Same thing with the part about announcing the tx, that's also explained in the Transfer tutorial.
So the snippet should only be lines 57 to 74 (in the JS file)

mkdocs/pages/en/devbook/transactions/messages.md
```diff
@@ -0,0 +1,151 @@
...
+
+To attach a plain text message to a transfer transaction, encode the message as bytes and include it in the `message`
+field.
```
segfaultxavi: I think this is a good place to talk about encodings, because we just told the readers to "encode" the message, but we didn't say what code they should use.

How about adding the paragraph from the Textbook:
> The Symbol protocol does not define a standard encoding for messages.
> It is up to the sender and recipient to agree on a format.
> By convention, most wallets and applications assume UTF-8 encoding.

mkdocs/pages/en/devbook/transactions/messages.md
```diff
@@ -0,0 +1,151 @@
...
+
+The transaction follows the same structure as a [basic transfer transaction](transfer.md), with the addition of the
+message field:
+
+* **Message format:** Plain text messages are encoded directly as UTF-8 bytes.
```
segfaultxavi: Now that we've talked about encodings, this line is no longer necessary (and it's not true, UTF-8 is a convention but anything else can be used).

mkdocs/pages/en/devbook/transactions/messages.md
```diff
@@ -0,0 +1,151 @@
...
+* **Maximum size:** Messages cannot exceed 1,024 bytes. The network will reject transactions with larger messages.
+
+    !!! tip "Handling larger data"
+
+        For applications requiring more than 1,024 bytes of data, consider storing the data off-chain and including a
+        hash or reference in the message field.
+        This approach maintains data integrity verification while keeping transaction sizes manageable.
```
segfaultxavi: This is a good addition!
Just two comments:
* I'd say "include a hash AND a reference", you need one to retrieve the message and the other to verify its integrity.
* I'd mention that you can also keep everything on-chain by using multiple transactions and an aggregate.

mkdocs/pages/en/devbook/transactions/messages.md
```diff
@@ -0,0 +1,151 @@
...
+
+* **Network visibility:** Plain text messages are publicly visible on the blockchain.
+    Anyone can read them by querying the transaction.
```
segfaultxavi: It's a nitpick, but encrypted messages are also publicly visible... just not decipherable. So I'd rephrase slightly:
```suggestion
* **Network visibility:** All messages are publicly visible on the blockchain, so unless they are encrypted,
    anyone can read them by querying the transaction.
```

mkdocs/pages/en/devbook/transactions/messages.md
```diff
@@ -0,0 +1,151 @@
...
+
+This example sends 1 XYM along with the message `"Hello, Symbol!"` to demonstrate that messages can be combined with
+mosaic transfers.
+
+!!! info "Messages without mosaics"
+
+    You can send a message without transferring any mosaics by providing an empty mosaics array.
```
segfaultxavi: Let's reverse this. We don't send any mosaic in the code, for simplicity, and just say here that you can combine mosaic transfers and messages if you want.

mkdocs/pages/en/devbook/transactions/messages.md
```diff
@@ -0,0 +1,151 @@
...
+
+### Sending an Encrypted Message
+
+{{ tutorial.code_snippet(['py:110:167', 'js:93:142']) }}
```
segfaultxavi: Again, too long. Skip the fetching of the network time (which I'd rather remove, as commented above) and the announcement of the tx.

mkdocs/pages/en/devbook/transactions/messages.md
```diff
@@ -0,0 +1,151 @@
...
+
+Encrypted messages provide confidentiality by encrypting the message content using a shared secret derived from the
```
segfaultxavi:
```suggestion
Encrypted messages provide confidentiality by protecting the message content using a shared secret derived from the
```

Minor tweak to avoid repetition. The paragraph is good!

mkdocs/pages/en/devbook/transactions/messages.md
```diff
@@ -0,0 +1,151 @@
...
+sender's private key and the recipient's public key.
+Both the sender and recipient can decrypt the message using their own private key and the other party's public key.
+
+The encryption process uses the <py:MessageEncoder> class:
```
segfaultxavi: Ugh, this should really be `<dy:MessageEncoder>` but I see the JS link was broken.
I pushed a commit to `new-docs` that fixes it, you can go back to using `dy:` everywhere :)

mkdocs/pages/en/devbook/transactions/messages.md
```diff
@@ -0,0 +1,151 @@
...
+
+    By convention, applications use a prefix byte to indicate message encoding:
+
+    * `0x00` - Unencrypted message
+    * `0x01` - Encrypted message (using Bouncy Castle's AES block cipher in CBC mode)
+
+    The SDK's `MessageEncoder` handles this convention, allowing both sender and recipient to decrypt messages using
+    each other's public keys.
+
+    See [Transfer Transactions](../../textbook/transfer_transactions.md#optional-message) in the textbook for details.
```
segfaultxavi: This is already the content of the info box in the Textbook, no need to repeat it.

How about something like:
```suggestion
Just as the Symbol protocol does not define a standard for message encoding, it also does not define one for encryption.
Whether a message is encrypted must be agreed upon in advance between the sender and the recipient.

However, some conventions have been adopted by applications over time.
For more details, see the note on [Optional Messages](../../textbook/transfer_transactions.md#optional-message) in the Textbook.
```

mkdocs/pages/en/devbook/transactions/messages.md
```diff
@@ -0,0 +1,151 @@
...
+
+### Decrypting a Received Message
+
+{{ tutorial.code_snippet(['py:169:188', 'js:144:164']) }}
+
+To decrypt a message, create a <py:MessageEncoder> with your key pair and call <py:MessageEncoder.try_decode> method
+with the other party's public key and the encrypted payload.
+
+In this example, the sender decrypts using their own key pair and the recipient's public key to verify the encryption.
```
segfaultxavi: You'll have to adapt this to match the code.

mkdocs/pages/en/devbook/transactions/messages.md
```diff
@@ -0,0 +1,151 @@
...
+    The SDK's `MessageEncoder` handles this convention, allowing both sender and recipient to decrypt messages using
+    each other's public keys.
+
+    See [Transfer Transactions](../../textbook/transfer_transactions.md#optional-message) in the textbook for details.
+
+### Decrypting a Received Message
+
+{{ tutorial.code_snippet(['py:169:188', 'js:144:164']) }}
+
+To decrypt a message, create a <py:MessageEncoder> with your key pair and call <py:MessageEncoder.try_decode> method
```
segfaultxavi:
```suggestion
To decrypt a message, create a <py:MessageEncoder> with one key pair (either the sender's or the recipient's) and call the <py:MessageEncoder.try_decode> method
```

To make it explicit that decoding can be used in two different ways.

mkdocs/pages/en/devbook/transactions/messages.md
```diff
@@ -0,0 +1,151 @@
...
+with the other party's public key and the encrypted payload.
+
+In this example, the sender decrypts using their own key pair and the recipient's public key to verify the encryption.
+The recipient would decrypt using their own key pair and the sender's public key (available in the transaction's
+`signerPublicKey` field).
+
+The method returns a tuple `(is_decoded, message)` indicating whether decryption was successful.
+
+If decryption succeeds, the message contains the original plaintext bytes, which you can decode as UTF-8.
```
segfaultxavi:
```suggestion
The method returns a tuple `(is_decoded, message)` indicating whether decryption was successful, and, if so, contains the original plaintext bytes, which still need to be decoded.
```

mkdocs/pages/en/devbook/transactions/messages.md
```diff
@@ -0,0 +1,151 @@
...
+
+```text
+--8<-- 'devbook/transactions/messages.log'
+```
+
+You can view the transactions on the [Symbol Testnet Explorer](https://testnet.symbol.fyi/) by searching for the
+transaction hashes printed in the output.
+
+When you view an encrypted message transaction in the explorer, the message field will appear as hexadecimal data
+beginning with `01`.
```
segfaultxavi: This does not work for symbol.fyi, you just get a random jumble, not a hex string.

mkdocs/pages/en/devbook/transactions/messages.md
```diff
@@ -0,0 +1,151 @@
...
+The explorer cannot decrypt the message because it doesn't have access to the private keys.
+
+## Conclusion
+
+This tutorial showed how to:
+
+| Step                                                           | Related documentation                     |
+| -------------------------------------------------------------- | ----------------------------------------- |
+| [Set up accounts](#setting-up-accounts)                        |                                           |
```
segfaultxavi: This part was not related to the topic, not worth mentioning here.

I think I would only list:
- Convert text into UTF-8 bytes using TextEncoder
- Encrypt and Decrypt a message using MessageEncoder
- How to include a message in a Transfer Transactions using SymbolTransactionFactory

segfaultxavi (changes requested):

mkdocs/overrides/devbook/transactions/messages.py
```diff
@@ -0,0 +1,232 @@
...
+
+# Fetch recommended fees
+fee_path = "/network/fees/transaction"
+print(f"Fetching recommended fees from {fee_path}")
+with urllib.request.urlopen(f"{NODE_URL}{fee_path}") as response:
+	response_json = json.loads(response.read().decode())
+	median_mult = response_json["medianFeeMultiplier"]
+	minimum_mult = response_json["minFeeMultiplier"]
+	fee_mult = max(median_mult, minimum_mult)
+	print(f"  Fee multiplier: {fee_mult}")
```
segfaultxavi, line 91: Since this bit is common for plain and encrypted messages, I would take it out of the Plain section and put it right after printing the addresses.

mkdocs/pages/en/devbook/transactions/messages.md
```diff
@@ -0,0 +1,159 @@
...
+## Conclusion
+
+This tutorial showed how to:
+
+| Step                                                           | Related documentation                       |
+| -------------------------------------------------------------- | ------------------------------------------- |
+| [Convert text into UTF-8 bytes](#sending-a-plain-text-message) | `TextEncoder`                               |
+| [Encrypt and decrypt a message](#sending-an-encrypted-message) | <dy:MessageEncoder>                         |
+| [Include a message in a Transfer Transaction](#sending-a-plain-text-message) | <dy:SymbolTransactionFactory> |
+
```
segfaultxavi: Extra blank line.

mkdocs/pages/en/devbook/transactions/messages.md
```diff
@@ -0,0 +1,159 @@
...
+### Sending a Plain Text Message
+
+{{ tutorial.code_snippet(['py:99:113', 'js:87:100']) }}
+
+You can also combine mosaic transfers with messages by including both the `mosaics` and `message` fields in the
+ transaction descriptor.
+
+See the [Creating a Transfer Transaction](./transfer.md) tutorial for details on fetching network time, calculating
+ fees, and announcing transactions.
```
segfaultxavi: This is a good comment, but I would move it to the intro of the Code Explanation section, in case anybody is following the code and is wondering what this is, before reaching this section.

You can say something like "the part about fetching network time and fees has been explained in the Transfer tutorial and is skipped here for brevity's sake."

mkdocs/pages/en/devbook/transactions/messages.md
```diff
@@ -0,0 +1,159 @@
...
+2. The message is encoded using the recipient's public key and the message bytes with <dy:MessageEncoder.encode>.
+3. The encrypted payload is attached to the transaction's `message` field.
+
+The transaction is then signed and announced following the same process as in
+[Creating a Transfer Transaction](./transfer.md).
+
+!!! note "Message encryption is a convention"
+
+    The Symbol protocol does not define a standard for message encryption.
+    Sender and recipient must agree in advance on whether messages are encrypted.
```
segfaultxavi:
```suggestion
    Sender and recipient must agree in advance on whether messages are encrypted and the cipher used.
```

mkdocs/pages/en/devbook/transactions/messages.md
segfaultxavi: I think this comment still applies. It's weird to see that the method returns `(is_decoded, message)` but the text only says that it returns whether decryption was successful. Yes, `message` is explained in the next paragraph, but if people try to understand paragraph by paragraph, this is confusing.

Also, I don't want to say "which you can decode as UTF-8", because we already said that's a convention, maybe some app wants to use a different encoding.

For these reasons, I still prefer my suggestion.

mkdocs/overrides/devbook/transactions/messages.py
```diff
@@ -0,0 +1,232 @@
...
+from symbolchain.sc import Amount
+
+# Configuration
+NODE_URL = os.environ.get(
+	"NODE_URL", "https://001-sai-dual.symboltest.net:3001"
+)
+
+
+# Helper function to poll for confirmed transaction
+def wait_for_confirmation(hash_value, label):
```
segfaultxavi: The function now returns the transaction data, so `wait_for_confirmation` is not descriptive enough.
How about `retrieve_transaction` or similar?

mkdocs/overrides/devbook/transactions/messages.py
```diff
@@ -0,0 +1,232 @@
...
+		attempts += 1
+		time.sleep(2)
+
+	if not confirmed:
+		raise Exception(
+			f"{label} not confirmed after {max_attempts} attempts"
+		)
+
+
+print(f"Using node {NODE_URL}")
```
segfaultxavi: This makes more sense right after defining the variable.

mkdocs/overrides/devbook/transactions/messages.py
```diff
@@ -0,0 +1,232 @@
...
+)
+recipient_address = facade.network.public_key_to_address(
+	recipient_key_pair.public_key
+)
+
+print(f"Sender address: {sender_address}")
+print(f"Recipient address: {recipient_address}\n")
+
+# ===== PLAIN TEXT MESSAGE =====
+print("=== Sending Plain Text Message ===")
```
segfaultxavi: Just an idea: why don't we use these headers to make it extra clear the direction of the messages?
So, for instance: `==> Sending Plain Text Message` and `<== Receiving Plain Text Message`.

mkdocs/overrides/devbook/transactions/messages.py
```diff
@@ -0,0 +1,232 @@
...
+(is_decoded, decrypted_message) = recipient_message_encoder.try_decode(
+	sender_public_key_from_tx, received_encrypted_message
+)
+
+if is_decoded:
+	message_text = decrypted_message.decode("utf-8")
+	print(f"Recipient decrypted message: {message_text}")
+else:
+	print(f"Recipient failed to decrypt message")
+
```
segfaultxavi: Extra blank line.

mkdocs/pages/en/devbook/transactions/messages.md
```diff
@@ -0,0 +1,159 @@
...
+        you to keep everything on the blockchain.
+    * **Off-chain storage:** Store the data off-chain and include a hash and a reference in the message field.
+        The hash verifies data integrity while the reference enables retrieval.
+
+### Receiving a Plain Text Message
+
+{{ tutorial.code_snippet(['py:138:151', 'js:120:130']) }}
+
+After announcing the transaction, the `wait_for_confirmation` helper function polls the
+`/transactions/confirmed/{hash}` endpoint until the transaction is confirmed.
```
segfaultxavi:
```suggestion
<get:/transactions/confirmed/{transactionId}> endpoint until the transaction is confirmed.
```

mkdocs/pages/en/devbook/transactions/messages.md
```diff
@@ -0,0 +1,159 @@
...
+    * **On-chain storage:** Split the data across multiple transactions within an <aggregate transaction:>, allowing
+        you to keep everything on the blockchain.
+    * **Off-chain storage:** Store the data off-chain and include a hash and a reference in the message field.
+        The hash verifies data integrity while the reference enables retrieval.
+
+### Receiving a Plain Text Message
+
+{{ tutorial.code_snippet(['py:138:151', 'js:120:130']) }}
+
+After announcing the transaction, the `wait_for_confirmation` helper function polls the
```
segfaultxavi: If you rename this function remember to search for the old name everywhere :)

mkdocs/pages/en/devbook/transactions/messages.md
```diff
@@ -0,0 +1,159 @@
...
+
+The explorer cannot decrypt encrypted messages because it doesn't have access to the private keys.
+
+## Conclusion
+
+This tutorial showed how to:
+
+| Step                                                           | Related documentation                       |
+| -------------------------------------------------------------- | ------------------------------------------- |
+| [Convert text into UTF-8 bytes](#sending-a-plain-text-message) | `TextEncoder`                               |
```
segfaultxavi: Maybe point out that `TextEncoder` (in JS) and the string `encode` and `decode` methods (in Python) are system methods and not part of the Symbol SDK.

segfaultxavi (changes requested):

PR #1739, opened 2025-10-17
[mkdocs] feat: add create account from private key tutorial

zero4862:
A tutorial for creating accounts based on this [code snippet](https://github.com/symbol/symbol/blob/main/docs/index.md#tutorial-creating-an-account).


mkdocs/overrides/devbook/accounts/create-from-private-key.log
```diff
@@ -0,0 +1,9 @@
+Creating a random account...
+    Address: TCWYXKVYBMO4NBCUF3AXKJMXCGVSYQOS7ZG2TLI
+    Public key: D04AB232742BB4AB3A1368BD4615E4E6D0224AB71A016BAF8520A332C9778737
+    Private key: 1111111111111111111111111111111111111111111111111111111111111111
```
segfaultxavi: This does not look like a randomly-generated key :)

mkdocs/overrides/devbook/accounts/create-from-private-key.py
```diff
@@ -0,0 +1,46 @@
+import os
+from symbolchain.CryptoTypes import PrivateKey
+from symbolchain.facade.SymbolFacade import SymbolFacade
+
+# Initialize the facade for the testnet network
+facade = SymbolFacade('testnet')
+
+# ===== RANDOM ACCOUNT =====
```
segfaultxavi: The two parts of the code are almost identical. I know it's for didactic purposes, but this case is so obvious the code is repeated that it will look weird.

How about we merge both parts and create the key like this:
```py
private_key_string = os.environ.get('PRIVATE_KEY', PrivateKey.random())
```
And then we explain the text that if you provide a key via the env var it will be used, and otherwise a random one will be created.
The rest of the explanation then remains the same. And this becomes a very short tutorial :D

segfaultxavi (changes requested):
Let's nail down the code first. I didn't see anything that needs changing in the text, but will review again after you adapt to the new code.

mkdocs/overrides/devbook/accounts/create-from-private-key.log
zero4862: Looks safer to use a placeholder (like 1111...1111) in the docs instead of real or random keys. In any case, I changed this example due to https://github.com/symbol/symbol/pull/1739#discussion_r2444111882, so the suggestion no longer applies.

mkdocs/overrides/devbook/accounts/create-from-private-key.py
zero4862: Interesting. I’ve adapted the tutorial as suggested. The code is a bit more complex since `PrivateKey.random()` returns a `PrivateKey object`, not a string, but in overall it looks shorter:

```
private_key_string = os.environ.get('PRIVATE_KEY')
if private_key_string:
	print("Loading account from environment variable...")
	private_key = PrivateKey(private_key_string)
else:
	print("Generating random account...")
	private_key = PrivateKey.random()
```

segfaultxavi (approved):
Excellent. Nothing to remark.

I like that you pointed the Hello World page to this tutorial, which is simpler than the Transfer one.

The JS link to `KeyPair.publicKey` does not work, but that's my fault and I'll fix it afterwards.

PR #1749, opened 2025-10-20
[jenkins]: add a new docker image just for Python and PostgreSQL

Wayonb:
problem: current PostgreSQL image has both Java and Python.
         Python needs to be upgraded but this will break Java
solution: split the image so Java and Python used seperate images


Jaguar0625 (approved):
👍

AnthonyLaw (approved):

PR #1772, opened 2025-11-07
[mkdocs] docs: add complete aggregate tutorial

zero4862:
Adds tutorial on how to send complete aggregate transactions.


mkdocs/overrides/devbook/transactions/complete-aggregate.py
```diff
@@ -0,0 +1,174 @@
...
+import time
+import urllib.request
+
+from symbolchain.CryptoTypes import PrivateKey
+from symbolchain.facade.SymbolFacade import SymbolFacade
+from symbolchain.sc import Amount
+from symbolchain.symbol.IdGenerator import generate_mosaic_alias_id
+from symbolchain.symbol.Network import NetworkTimestamp
+
+NODE_URL = 'https://001-sai-dual.symboltest.net:3001'
```
segfaultxavi: I'm thinking that we should use `reference.symboltest.net` in all tutorials.
I just recently learnt of this node.

mkdocs/overrides/devbook/transactions/complete-aggregate.py
```diff
@@ -0,0 +1,174 @@
...
+print(f'Using node {NODE_URL}')
+
+# Account A (initiates the aggregate and sends XYM)
```
segfaultxavi:
```suggestion
# Account A (initiates the aggregate tx and sends XYM to Account B)
```

mkdocs/overrides/devbook/transactions/complete-aggregate.py
```diff
@@ -0,0 +1,174 @@
...
+ACCOUNT_A_PRIVATE_KEY = os.getenv(
+	'ACCOUNT_A_PRIVATE_KEY',
+	'0000000000000000000000000000000000000000000000000000000000000000')
+account_a_key_pair = SymbolFacade.KeyPair(
+	PrivateKey(ACCOUNT_A_PRIVATE_KEY))
+
+# Account B (sends custom mosaic back to Account A)
```
segfaultxavi:
```suggestion
# Account B (sends custom mosaic to Account A)
```

mkdocs/overrides/devbook/transactions/complete-aggregate.py
```diff
@@ -0,0 +1,174 @@
...
+	fee_path = '/network/fees/transaction'
+	print(f'Fetching recommended fees from {fee_path}')
+	with urllib.request.urlopen(f'{NODE_URL}{fee_path}') as response:
+		response_json = json.loads(response.read().decode())
+		median_mult = response_json['medianFeeMultiplier']
+		minimum_mult = response_json['minFeeMultiplier']
+		fee_mult = max(median_mult, minimum_mult)
+		print(f'  Fee multiplier: {fee_mult}')
+
+	# Account A sends 10 XYM to Account B
```
segfaultxavi:
```suggestion
	# Embedded tx 1: Account A transfers 10 XYM to Account B
```

mkdocs/overrides/devbook/transactions/complete-aggregate.py
```diff
@@ -0,0 +1,174 @@
...
+		'type': 'transfer_transaction_v1',
+		'signer_public_key': account_a_key_pair.public_key,
+		'recipient_address': account_b_address,
+		'mosaics': [{
+			'mosaic_id': generate_mosaic_alias_id('symbol.xym'),
+			'amount': 10_000_000  # 10 XYM (divisibility = 6)
+		}]
+	})
+
+	# Account B sends 1 custom mosaic to Account A
```
segfaultxavi:
```suggestion
	# Embedded tx 2: Account B transferss 1 custom mosaic to Account A
```

mkdocs/overrides/devbook/transactions/complete-aggregate.py
```diff
@@ -0,0 +1,174 @@
...
+
+	# --- ACCOUNT B (Cosignatory) ---
+	print('[Account B] Received payload')
+	received_transaction = facade.transaction_factory.deserialize(
+		bytes.fromhex(json.loads(shared_payload)['payload']))
+
+	print('[Account B] Cosigning...')
+	cosignature_b = facade.cosign_transaction(
+		account_b_key_pair, received_transaction)
+	print(f'[Account B] Cosignature created: {cosignature_b}')
```
segfaultxavi:
```suggestion
	cosignature_formatted = json.dumps(cosignature_b.to_json(), indent=2)
	print(f'[Account B] Cosignature created:\n{cosignature_formatted}')
```
For consistency, so we always output JSON.

mkdocs/overrides/devbook/transactions/complete-aggregate.py
```diff
@@ -0,0 +1,174 @@
...
+		'signer_public_key': account_a_key_pair.public_key,
+		'deadline': timestamp.add_hours(2).timestamp,
+		'transactions_hash': facade.hash_embedded_transactions(
+			embedded_transactions),
+		'transactions': embedded_transactions
+	})
+	# Reserve space for one cosignature (104 bytes each)
+	# and calculate fee for the final transaction size
+	transaction.fee = Amount(fee_mult * (transaction.size + 104))
+	print('Built aggregate transaction:')
```
segfaultxavi:
```suggestion
	print('Built aggregate transaction without signatures:')
```

mkdocs/overrides/devbook/transactions/complete-aggregate.py
```diff
@@ -0,0 +1,174 @@
...
+		json.loads(transaction_payload), indent=2)
+	print(f'[Account A] Payload ready to share:\n{payload_formatted}')
+
+	# --- OFF-CHAIN COORDINATION ---
+	# Account A sends the payload to Account B
+	shared_payload = transaction_payload
+	print('[Sharing] Payload sent to Account B...')
+
+	# --- ACCOUNT B (Cosignatory) ---
+	print('[Account B] Received payload')
```
segfaultxavi:
```suggestion
	print('[Account A] ==> Payload sent to Account B (offchain)')

	# --- ACCOUNT B (Cosignatory) ---
```

I would use the ==> arrows as we did for the Messages tutorial.
And I would remove the second print, since it might be a bit confusing because we are not actually sending or receiving anything.

mkdocs/overrides/devbook/transactions/complete-aggregate.py
```diff
@@ -0,0 +1,174 @@
...
+		account_b_key_pair, received_transaction)
+	print(f'[Account B] Cosignature created: {cosignature_b}')
+
+	# --- OFF-CHAIN COORDINATION ---
+	# Account B sends the cosignature back to Account A
+	shared_cosignature = cosignature_b
+	print('[Sharing] Cosignature sent back to Account A...')
+
+	# --- ACCOUNT A (Initiator) ---
+	print('[Account A] Received cosignature')
```
segfaultxavi:
```suggestion
	print('[Account B] <== Cosignature sent back to Account A (offchain)')

	# --- ACCOUNT A (Initiator) ---
```

Again, use arrows and remove second print.

mkdocs/pages/en/devbook/transactions/complete-aggregate.md
```diff
@@ -0,0 +1,221 @@
...
+        }
+        subgraph clusterT1 {
+            label = "Embedded Transfer 1";
+            tooltip = "Embedded Transfer 1";
+            style = dashed;
+            A2 [label="Account A" tooltip="Account A"];
+            B2 [label="Account B" tooltip="Account B"];
+            A2 -> B2 [label="10 XYM"];
+        }
+
```
segfaultxavi: Put clusterT1 before clusterT2 so they appear in their natural order.
Remove empty line.

mkdocs/pages/en/devbook/transactions/complete-aggregate.md
```diff
@@ -0,0 +1,221 @@
...
+    }
+}
+```
+
+Both parties coordinate off-chain to collect signatures before announcement, ensuring the swap executes as a
+single atomic transaction.
+
+
```
segfaultxavi, line 43:
```suggestion

```
No multiple blank lines. Are you using the markdownlint extension?

mkdocs/pages/en/devbook/transactions/complete-aggregate.md
```diff
@@ -0,0 +1,221 @@
...
+!!! tip "When to use complete vs bonded aggregates"
+
+    Complete aggregates work well in the following cases:
+
+    - A single account batches multiple transactions.
+    - Multiple parties coordinate off-chain to collect <signatures:> before publishing the transaction.
+
+    If off-chain coordination is impractical, use <Bonded Aggregate Transactions:> instead,
+    which allow <cosignatures:> to be added on-chain.
```
segfaultxavi: Good idea, this is a useful info box.
Just a couple of things:
* "Bonded" appears for the first time in the title, so we cannot provide a link to explain what it is. I would remove it from the title.
* You're mixing what agg txes are good for (batching txes) with what complete agg txes are good for (coordinating offchain). This is a confusing topic, so we need to be careful.
* This is the ideal place to provide a concrete example that helps understand the tutorial.

So I suggest something like this. Feel free to adapt.

```suggestion
!!! tip "Two types of aggregate transactions"

    <Aggregate transactions:> group multiple <transactions:> in single operation, and require <signatures:> from all involved accounts.

    A _complete aggregate transaction_ collects all signatures before being announced,
    and therefore works best when parties can easily communicate off-chain and share the same signing tools.
    For example, inside the same company.

    If off-chain coordination is impractical, use <Bonded Aggregate Transactions:> instead,
    which allow <cosignatures:> to be added on-chain.
```

mkdocs/pages/en/devbook/transactions/complete-aggregate.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Prerequisites
+
+Before you start, make sure to:
+
+- Set up your development environment.
```
segfaultxavi: Use asterisks for unordered lists.

mkdocs/pages/en/devbook/transactions/complete-aggregate.md
```diff
@@ -0,0 +1,221 @@
...
+  See [Setting Up a Development Environment](../start/setup.md).
+- Create an <account:> to initiate the aggregate transaction, either
```
segfaultxavi: You need to create two accounts for this tutorial.

mkdocs/pages/en/devbook/transactions/complete-aggregate.md
```diff
@@ -0,0 +1,221 @@
...
+* **Transaction-specific fields:** All fields specific to the transaction type must be provided.
+  For transfers, this includes the recipient address and the <mosaics:> to send.
+
+Note that embedded transactions do **not** include fee or deadline fields.
+These are inherited from the enclosing aggregate transaction.
+
+The example creates two <transfer transactions:> for the swap:
+
+* The first transfer sends 10 <XYM:> from Account A to Account B.
+* The second transfer sends 1 custom <mosaic:> from Account B to Account A.
```
segfaultxavi: No need to link to mosaic again so soon.

mkdocs/pages/en/devbook/transactions/complete-aggregate.md
```diff
@@ -0,0 +1,221 @@
...
+### Building the Aggregate Transaction
+
+{{ tutorial.code_snippet(['py:81:96', 'js:81:98']) }}
+
+Once the embedded transactions are prepared, create the complete aggregate transaction that wraps them:
+
+* **Type:** Use `aggregate_complete_transaction_v3`.
+
+* **Signer public key:** The account initiating the aggregate.
+  This account announces the transaction and pays the transaction fee.
```
segfaultxavi, line 152: There's an interesting detail worth mentioning here, maybe even in an info box.
If the accounts want to share the cost of the transaction, B can include a XYM transfer to A to help pay the fee.
This can be done as an additional transaction or including it in the list of mosaics if the B->A embedded tx.

At some point we'll want a full tutorial about this technique, because it also enables sending transactions for free by having another account pay the fees.

mkdocs/pages/en/devbook/transactions/complete-aggregate.md
```diff
@@ -0,0 +1,221 @@
...
+
+* **Deadline:** The maximum time the network should attempt to confirm the transaction.
+
+* **Transactions hash:** A hash computed from all embedded transactions.
+  This ensures the embedded transactions cannot be modified after signing.
+  Use <dy:SymbolFacade.hashEmbeddedTransactions> to compute this value.
+
+* **Transactions:** The array of embedded transactions to execute.
+
+The <fee:> is calculated based on the aggregate's total size, which includes all embedded transactions plus
```
segfaultxavi: Oops! We don't have a definition for `fee` yet!
I'll add one, and add links to it everywhere, but meanwhile, we better remove this link.

mkdocs/overrides/devbook/transactions/complete-aggregate.py
```diff
@@ -0,0 +1,174 @@
...
+		embedded_transaction_1, embedded_transaction_2]
+	transaction = facade.transaction_factory.create({
+		'type': 'aggregate_complete_transaction_v3',
+		'signer_public_key': account_a_key_pair.public_key,
+		'deadline': timestamp.add_hours(2).timestamp,
+		'transactions_hash': facade.hash_embedded_transactions(
+			embedded_transactions),
+		'transactions': embedded_transactions
+	})
+	# Reserve space for one cosignature (104 bytes each)
```
segfaultxavi:
```suggestion
	# Reserve space for one cosignature (104 bytes)
```

mkdocs/pages/en/devbook/transactions/complete-aggregate.md
```diff
@@ -0,0 +1,221 @@
...
+{{ tutorial.code_snippet(['py:81:96', 'js:81:98']) }}
+
+Once the embedded transactions are prepared, create the complete aggregate transaction that wraps them:
+
+* **Type:** Use `aggregate_complete_transaction_v3`.
+
+* **Signer public key:** The account initiating the aggregate.
+  This account announces the transaction and pays the transaction fee.
+
+* **Deadline:** The maximum time the network should attempt to confirm the transaction.
```
segfaultxavi:
```suggestion
* **Deadline:** The timestamp, in network time, after which the transaction expires and can no longer be confirmed.
```

And link `network time` to the place where it is defined in the Transfer Transaction tutorial.

mkdocs/pages/en/devbook/transactions/complete-aggregate.md
```diff
@@ -0,0 +1,221 @@
...
+Account A then sends this payload to Account B through an off-chain channel.
+
+2. **Account B (Cosignatory)** receives the payload and deserializes it using
+<dy:SymbolTransactionFactory.deserialize> to reconstruct the transaction object.
+Account B then cosigns using <dy:SymbolFacade.cosignTransaction>, which computes the transaction hash and
+signs it.
+The cosignature is sent back to Account A.
+
+3. **Account A** receives the cosignature and adds it to the transaction object's cosignatures array,
+then rebuilds the payload.
```
segfaultxavi:
```suggestion
With the aggregate transaction built, both accounts must sign it off-chain before it can be announced.
The snippet above separates the process by machine:

1. **Account A (Initiator)** signs the transaction using <dy:SymbolFacade.signTransaction>.
It then uses <dy:SymbolTransactionFactory.attachSignature> which normally produces a fully announce-ready payload.
In this case, however, the payload is still missing Account B’s cosignature.
Account A sends this intermediate payload to Account B through an off-chain channel.

2. **Account B (Cosignatory)** receives the payload and deserializes it using
<dy:SymbolTransactionFactory.deserialize> to reconstruct the transaction object.
Account B should verify that the embedded transactions match what it expects to sign.
It then cosigns using <dy:SymbolFacade.cosignTransaction>, which computes the transaction hash and
produces a cosignature object.
Only this cosignature is sent back to Account A.

3. **Account A** receives the Account B's cosignature, adds it to the transaction object's `cosignatures` array,
and rebuilds the payload for announcement.
```

I felt this section was a bit convoluted.

mkdocs/pages/en/devbook/transactions/complete-aggregate.md
```diff
@@ -0,0 +1,221 @@
...
+
+!!! note "Single-signer complete aggregates"
```
segfaultxavi: Again, good point! This is a useful bit of info.

We can complete it by explaining that if an account needs to sign multiple embedded transactions, only one signature is needed.
For example, Account A only needs to sign once, even though its signature is needed for the aggregate tx AND the first embedded tx.
And link to https://docs.symboltest.net/en/textbook/transactions/#embedded-transactions
And the title of the box needs to be changed then.

mkdocs/pages/en/devbook/transactions/complete-aggregate.md
```diff
@@ -0,0 +1,221 @@
...
+
+### Waiting for Confirmation
+
+{{ tutorial.code_snippet(['py:151:171', 'js:156:180']) }}
+
+After announcement, the transaction status is monitored using <get:/transactionStatus/{hash}>.
+
+The polling loop checks the status every second until the transaction is confirmed or fails.
+Once confirmed, the swap is complete and both transfers have executed.
+
```
segfaultxavi, line 231: Add that, for more information about announcing and awaiting confirmation the reader should check the Transfer Transaction tutorial.

mkdocs/pages/en/devbook/transactions/complete-aggregate.md
```diff
@@ -0,0 +1,221 @@
...
+!!! note "Single-signer complete aggregates"
+
+    When all embedded transactions share the same signer, such as batching multiple operations
+    from one account, cosignatures are **not** required.
+
+    In this case, the complete aggregate transaction can be announced immediately after signing, and the fee
+    calculation does not need to reserve space for cosignatures.
+
+### Announcing the Transaction
+
```
segfaultxavi, line 211: Before the snippet, add that: now that the transaction is ready to be announced, it follows the same process as regular, non-aggregate transactions, as shown in the Transfer Transaction tutorial.

mkdocs/pages/en/devbook/transactions/complete-aggregate.md
```diff
@@ -0,0 +1,221 @@
...
+The polling loop checks the status every second until the transaction is confirmed or fails.
+Once confirmed, the swap is complete and both transfers have executed.
+
+## Output
+
+The output shown below corresponds to a typical run of the program.
+
+```text
+--8<-- 'devbook/transactions/complete-aggregate.log'
```
segfaultxavi, line 237: This is quite a long output log. It will be more useful if you add line numbers and highlight some lines, and then comment on those lines.

```md
   ```text linenums="1" hl_lines="10 20 30"
```

For example, I think these lines are worth explaining:
* `"type"` indicates that this is an aggregate.
* `"transactions"` contains the two embedded txes.
* `"signature": "0000"` contains the signature for the aggregate.
* `"cosignatures": []` contains any other cosignatures. If you output the transaction once the signatures have been added instead of at the beginning, these fields will make more sense, and you can add "note how Account A's signature is only needed once".
* `Waiting for confirmation` contains the hash of the transaction, in case readers wants to check it in the explorer.

And any others you see fit.

mkdocs/pages/en/devbook/transactions/complete-aggregate.md
```diff
@@ -0,0 +1,221 @@
...
+You can view the transactions on the [Symbol Testnet Explorer](https://testnet.symbol.fyi/) by searching for the
+aggregate transaction hash announced.
+
+## Conclusion
+
+This tutorial showed how to:
+
+| Step                                                                   | Related documentation                                                               |
+| ---------------------------------------------------------------------- | ------------------------------------------------------------------------------------|
+| [Obtain deadline and fee information](#fetching-network-time-and-fees) | <get:/node/time><br/><get:/network/fees/transaction>                                |
```
segfaultxavi: But this was not the purpose of this tutorial, so I wouldn't list this.

mkdocs/pages/en/devbook/transactions/complete-aggregate.md
```diff
@@ -0,0 +1,221 @@
...
+| [Create embedded transactions](#creating-embedded-transactions)        | <dy:SymbolTransactionFactory.createEmbedded>                                        |
+| [Build the aggregate](#building-the-aggregate-transaction)             | <dy:SymbolTransactionFactory.create><br/><dy:SymbolFacade.hashEmbeddedTransactions> |
+| [Collect signatures off-chain](#collecting-signatures)                 | <dy:SymbolFacade.signTransaction><br/><dy:SymbolFacade.cosignTransaction>           |
+| [Announce the transaction](#announcing-the-transaction)                | <put:/transactions>                                                                 |
+| [Wait for confirmation](#waiting-for-confirmation)                     | <get:/transactionStatus/{hash}>                                                     |
```
segfaultxavi: Same with these two, I would remove them.
People reading this tutorial should already know how to do this. And if they don't, we've given them plenty opportunities to read the more basic tutorial.

segfaultxavi (changes requested):

segfaultxavi:
I sent the review too soon and missed one last bit.

We should talk a bit about the custom mosaic. We should explain that it was created for the purpose of this tutorial and the default Account B was seeded with it so the aggregate transaction succeeded.

mkdocs/pages/en/devbook/transactions/complete-aggregate.md
segfaultxavi: Weird. It doesn't happen in the public docs.
https://docs.symboltest.net/en/textbook/transactions/#aggregate-transactions

Different version of some tool?

mkdocs/overrides/devbook/transactions/complete-aggregate.py
zero4862: Let's replace them all in a separate PR after merging this PR and #1773, to avoid reviewing too many changes here.

mkdocs/pages/en/devbook/transactions/complete-aggregate.md
zero4862: Edited, but it renders in the opposite order:

image

It also happens in other pages:

image

zero4862: Ok, should be an issue on my side.

mkdocs/pages/en/devbook/transactions/complete-aggregate.md
zero4862: Edited part (title, explain what an aggregate is), but did not apply:  

- “share the same signing tools”:  Could be misleading because parties don't actually need to use the same tools. Each party signs with their own private key on their own machine using whatever SDK, wallet, or signing infrastructure they prefer.
- “for example, inside the same company”:  could be too prescriptive and potentially limiting. Could be done between comapnies / individuals if they can colaborate off-chain (e.g. via  a platform that makes that easy).  

Also, kept the use case “A single account batches multiple transactions”.

mkdocs/pages/en/devbook/transactions/complete-aggregate.md
```diff
@@ -0,0 +1,263 @@
...
+--8<-- 'devbook/transactions/complete-aggregate.log'
+```
+
+Key points in the output:
+
+* `"signature": "0000..."`: Shows all zeros initially because the transaction hasn't been signed yet.
+* `"type": 16705`: Indicates this is an `aggregate_complete_transaction_v3`.
+* `"transactions"`: Contains the two embedded transfers that will execute atomically.
+* `"cosignatures": []`: Initially empty. Account B's cosignature is added before announcement.
+  Note how Account A's signature is only needed once, even though it appears as signer in both the aggregate and the
```
zero4862:
```suggestion
  Note how Account A's signature is only needed once, even though it appears as signer in both the aggregate and the
```

Extra whitespace, will fix after the review.

mkdocs/pages/en/devbook/transactions/complete-aggregate.md
```diff
@@ -0,0 +1,263 @@
...
+    }
+}
+```
+
+Both parties coordinate off-chain to collect signatures before announcement, ensuring the swap executes as a
+single atomic transaction.
+
+!!! note "Two types of aggregate transactions"
+
+    <Aggregate transactions:> group multiple <transactions:> in single operation, and require <signatures:> from all
```
segfaultxavi:
```suggestion
    <Aggregate transactions:> group multiple <transactions:> in a single operation, and require <signatures:> from all
```

mkdocs/pages/en/devbook/transactions/complete-aggregate.md
```diff
@@ -0,0 +1,263 @@
...
+    involved accounts.
+
+    A _complete aggregate transaction_ collects all signatures before being announced.
+    This works well in two scenarios:
```
segfaultxavi:
```suggestion
    This procedure works well, for example, in the following scenarios:
```

I don't want to suggest that these are the _only_ scenarios where this works well.
Also, "naked this" are to be avoided.

mkdocs/pages/en/devbook/transactions/complete-aggregate.md
```diff
@@ -0,0 +1,263 @@
...
+
+    * **Multi-party coordination:** When parties can communicate off-chain to exchange transaction payloads and
+      cosignatures.
```
segfaultxavi: This is now the first usage of the term "cosignature", so the link to the definition should be moved here.

Also, I realize the definition of cosignature is a bit ambiguous: it should clarify that it applies to multiple signatures from different accounts. I'll fix that in the Textbook.

mkdocs/pages/en/devbook/transactions/complete-aggregate.md
```diff
@@ -0,0 +1,263 @@
...
+
+* Set up your development environment.
+  See [Setting Up a Development Environment](../start/setup.md).
+* Create an <account:> (Account A) to initiate the aggregate transaction, either
+  [from code](../accounts/create-from-private-key.md) or
+  [by using a wallet](../../userbook/wallet/create-account.md).
+* Create a second account (Account B) to participate in the swap.
+* Obtain <XYM:> for Account A to pay for the transaction fee and transfer amounts.
+  See [Getting Testnet Funds from the Faucet](../accounts/testnet-faucet.md).
+* Create a <mosaic:> owned by Account B for the swap.
```
segfaultxavi: This long list of prerequisites looks a bit scary. Maybe we can split it like this:

```suggestion
* Set up your development environment.
  See [Setting Up a Development Environment](../start/setup.md).
  
This tutorial requires two accounts with adequate funds to perform the transfers.
Although pre-funded accounts are provided for convenience, they are not maintained and may run out of funds.
If so, you will need to:

* Create an <account:> (Account A) to initiate the aggregate transaction, either
  [from code](../accounts/create-from-private-key.md) or
  [by using a wallet](../../userbook/wallet/create-account.md).
* Create a second account (Account B) to participate in the swap.
* Obtain <XYM:> for Account A to pay for the transaction fee and transfer amounts.
  See [Getting Testnet Funds from the Faucet](../accounts/testnet-faucet.md).
* Create a <mosaic:> owned by Account B for the swap.
```

mkdocs/pages/en/devbook/transactions/complete-aggregate.md
```diff
@@ -0,0 +1,263 @@
...
+
+    An account only signs once, even if it appears as the signer in multiple embedded transactions.
+    In this tutorial, Account A signs the aggregate transaction, which covers both the aggregate itself and the
+    first embedded transaction where Account A is the signer.
+
+    When all embedded transactions share the same signer (batching multiple operations from one account),
+    cosignatures are **not** required. The aggregate can be announced immediately after signing, and the fee
+    calculation does not need to reserve space for cosignatures.
+
+    See [Embedded Transactions](../../textbook/transactions.md#embedded-transactions) for more details.
```
segfaultxavi: Better use a definition `<embedded transactions:>`. It's shorter, path-independent, and provides a nice tooltip (although at this point in the tutorial it shouldn't be necessary).

mkdocs/pages/en/devbook/transactions/complete-aggregate.md
```diff
@@ -0,0 +1,263 @@
...
+
+## Output
+
+The output shown below corresponds to a typical run of the program.
+
+```text linenums="1" hl_lines="10 14 18 48 53 60 66"
+--8<-- 'devbook/transactions/complete-aggregate.log'
+```
+
+Key points in the output:
```
segfaultxavi, line 240: The idea behind adding line numbers to the output log was to reference them now :)

For example:
```md
* Line 10 (`"signature": "0000..."`): Shows all zeros...
```
Or:
```md
* Line 10, `"signature": "0000..."`
    Shows all zeros...
```

mkdocs/pages/en/devbook/transactions/complete-aggregate.md
```diff
@@ -0,0 +1,263 @@
...
+
+* `"signature": "0000..."`: Shows all zeros initially because the transaction hasn't been signed yet.
+* `"type": 16705`: Indicates this is an `aggregate_complete_transaction_v3`.
+* `"transactions"`: Contains the two embedded transfers that will execute atomically.
+* `"cosignatures": []`: Initially empty. Account B's cosignature is added before announcement.
+  Note how Account A's signature is only needed once, even though it appears as signer in both the aggregate and the 
+  first embedded transaction.
+* `"payload": "8010..."`: The transaction payload computed from the aggregate transaction and its embedded transactions.
```
segfaultxavi:
```suggestion
* `"payload": "68010..."`: The transaction payload computed from the aggregate transaction and its embedded transactions.
```

segfaultxavi (changes requested):
Almost there!

mkdocs/overrides/devbook/transactions/complete-aggregate.py
```diff
@@ -0,0 +1,173 @@
...
+import time
+import urllib.request
+
+from symbolchain.CryptoTypes import PrivateKey
+from symbolchain.facade.SymbolFacade import SymbolFacade
+from symbolchain.sc import Amount
+from symbolchain.symbol.IdGenerator import generate_mosaic_alias_id
+from symbolchain.symbol.Network import NetworkTimestamp
+
+NODE_URL = 'https://001-sai-dual.symboltest.net:3001'
```
segfaultxavi: Should fetch NODE_URL env var as in the other tutorials.

segfaultxavi (approved):

PR #1773, opened 2025-11-07
[mkdocs] docs: add bonded aggregate tutorial

zero4862:
Adds tutorial on how to send bonded aggregate transactions.

### DIFF
--- a/mkdocs/overrides/devbook/transactions/bonded-aggregate.py
+++ b/mkdocs/overrides/devbook/transactions/bonded-aggregate.py
@@ -205,21 +205,19 @@ try:
 
 	print(f'Found {len(partial_txs["data"])} partial transaction(s)')
 
-	# Find the transaction matching our expected hash
-	partial_tx_hash = None
-	for tx in partial_txs['data']:
-		if tx['meta']['hash'] == str(bonded_hash):
-			partial_tx_hash = tx['meta']['hash']
-			print(f'Found matching transaction: {partial_tx_hash}')
-			break
-
-	if not partial_tx_hash:
+	# Find the transaction matching the expected hash
+	found = any(
+		tx['meta']['hash'] == str(bonded_hash)
+		for tx in partial_txs['data']
+	)
+	if not found:
 		raise Exception(
 			f'Expected transaction {bonded_hash} not found in '
 			f'partial transactions')
+	print(f'Found matching transaction: {bonded_hash}')
 
 	# Fetch full transaction details using the hash
-	detail_path = f'/transactions/partial/{partial_tx_hash}'
+	detail_path = f'/transactions/partial/{bonded_hash}'
 	with urllib.request.urlopen(f'{NODE_URL}{detail_path}') as response:
 		partial_tx_json = json.loads(response.read().decode())
 
@@ -234,7 +232,7 @@ try:
 	cosignature_path = '/transactions/cosignature'
 	print('[Account B] Cosigning the bonded aggregate...')
 	cosignature = facade.cosign_transaction_hash(
-		account_b_key_pair, Hash256(partial_tx_hash), True)
+		account_b_key_pair, bonded_hash, True)
 	cosignature_payload = json.dumps({
 		'version': str(cosignature.version),
 		'signerPublicKey': str(cosignature.signer_public_key),
@@ -249,7 +247,7 @@ try:
 
 	# Wait for final confirmation
 	wait_for_status(
-		Hash256(partial_tx_hash), 'confirmed',
+		bonded_hash, 'confirmed',
 		'Bonded aggregate transaction'
 	)
 
--- a/mkdocs/overrides/devbook/transactions/bonded-aggregate.py
+++ b/mkdocs/overrides/devbook/transactions/bonded-aggregate.py
@@ -3,7 +3,7 @@ import os
 import time
 import urllib.request
 
-from symbolchain.CryptoTypes import PrivateKey
+from symbolchain.CryptoTypes import Hash256, PrivateKey
 from symbolchain.facade.SymbolFacade import SymbolFacade
 from symbolchain.sc import Amount
 from symbolchain.symbol.IdGenerator import generate_mosaic_alias_id
@@ -12,14 +12,65 @@ from symbolchain.symbol.Network import NetworkTimestamp
 NODE_URL = 'https://001-sai-dual.symboltest.net:3001'
 print(f'Using node {NODE_URL}')
 
-# Account A (initiates the aggregate and sends XYM)
+# Helper function to announce transaction
+def announce_transaction(payload, endpoint, label):
+	print(f'Announcing {label} to {endpoint}')
+	request = urllib.request.Request(
+		f'{NODE_URL}{endpoint}',
+		data=payload.encode(),
+		headers={'Content-Type': 'application/json'},
+		method='PUT'
+	)
+	with urllib.request.urlopen(request) as response:
+		print(f'  Response: {response.read().decode()}')
+
+# Helper function to wait for transaction status
+def wait_for_status(hash_value, expected_status, label):
+	print(
+		f'Waiting for {label} to reach {expected_status} status from '
+		f'/transactionStatus/{hash_value}'
+	)
+
+	status_path = f'/transactionStatus/{hash_value}'
+	for attempt in range(60):
+		time.sleep(1)
+		try:
+			with urllib.request.urlopen(
+				f'{NODE_URL}{status_path}'
+			) as response:
+				status = json.loads(response.read().decode())
+				if status['group'] == expected_status:
+					if expected_status == 'confirmed':
+						print(f'  Transaction status: {status["group"]}')
+						print(
+							f'{label} confirmation in ' +
+							f'{attempt + 1} seconds')
+					elif expected_status == 'partial':
+						print(
+							'  Transaction is partial, '
+							'ready for cosignatures'
+						)
+					return attempt + 1
+				if status['group'] == 'failed':
+					print(f'  Transaction status: {status["group"]}')
+					print(f'{label} failed: {status["code"]}')
+					return attempt + 1
+				if expected_status == 'confirmed':
+					print(f'  Transaction status: {status["group"]}')
+		except urllib.error.HTTPError as e:
+			print(f'  Transaction status: unknown | Cause: ({e.msg})')
+	else:
+		print(f'{label} took too long.')
+	return 60
+
+# Account A (initiates the aggregate tx and sends XYM to Account B)
 ACCOUNT_A_PRIVATE_KEY = os.getenv(
 	'ACCOUNT_A_PRIVATE_KEY',
 	'0000000000000000000000000000000000000000000000000000000000000000')
 account_a_key_pair = SymbolFacade.KeyPair(
 	PrivateKey(ACCOUNT_A_PRIVATE_KEY))
 
-# Account B (sends custom mosaic back to Account A)
+# Account B (sends custom mosaic to Account A)
 ACCOUNT_B_PRIVATE_KEY = os.getenv(
 	'ACCOUNT_B_PRIVATE_KEY',
 	'1111111111111111111111111111111111111111111111111111111111111111')
@@ -55,7 +106,7 @@ try:
 		fee_mult = max(median_mult, minimum_mult)
 		print(f'  Fee multiplier: {fee_mult}')
 
-	# Account A sends 10 XYM to Account B
+	# Embedded tx 1: Account A transfers 10 XYM to Account B
 	embedded_transaction_1 = facade.transaction_factory.create_embedded({
 		'type': 'transfer_transaction_v1',
 		'signer_public_key': account_a_key_pair.public_key,
@@ -66,7 +117,7 @@ try:
 		}]
 	})
 
-	# Account B sends 1 custom mosaic to Account A
+	# Embedded tx 2: Account B transfers 1 custom mosaic to Account A
 	custom_mosaic_id = 0x6D1314BE751B62C2
 	embedded_transaction_2 = facade.transaction_factory.create_embedded({
 		'type': 'transfer_transaction_v1',
@@ -78,7 +129,7 @@ try:
 		}]
 	})
 
-	# Build the aggregate bonded transaction
+	# Build the bonded aggregate transaction
 	embedded_transactions = [
 		embedded_transaction_1, embedded_transaction_2]
 	bonded_transaction = facade.transaction_factory.create({
@@ -89,20 +140,22 @@ try:
 			embedded_transactions),
 		'transactions': embedded_transactions
 	})
-	# Reserve space for one cosignature (104 bytes each)
+	# Reserve space for one cosignature (104 bytes)
 	# and calculate fee for the final transaction size
 	bonded_transaction.fee = Amount(
 		fee_mult * (bonded_transaction.size + 104))
-	print('Built aggregate bonded transaction:')
+	print('Built aggregate without signatures:')
 	print(json.dumps(bonded_transaction.to_json(), indent=2))
 
-	# Sign the bonded transaction
+	# --- ACCOUNT A (Initiator) ---
+	# Sign the bonded aggregate transaction
+	print('[Account A] Signing the bonded aggregate...')
 	bonded_signature = facade.sign_transaction(
 		account_a_key_pair, bonded_transaction)
 	bonded_json_payload = facade.transaction_factory.attach_signature(
 		bonded_transaction, bonded_signature)
 	bonded_hash = facade.hash_transaction(bonded_transaction)
-	print(f'Bonded transaction hash: {bonded_hash}')
+	print(f'Bonded aggregate transaction hash: {bonded_hash}')
 
 	# Create hash lock transaction
 	print('Creating hash lock transaction...')
@@ -120,6 +173,7 @@ try:
 	hash_lock.fee = Amount(fee_mult * hash_lock.size)
 
 	# Sign and announce hash lock
+	print('[Account A] Signing the hash lock...')
 	hash_lock_signature = facade.sign_transaction(
 		account_a_key_pair, hash_lock)
 	hash_lock_payload = facade.transaction_factory.attach_signature(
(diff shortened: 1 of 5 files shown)


mkdocs/overrides/devbook/transactions/bonded-aggregate.py
```diff
@@ -0,0 +1,231 @@
...
+		try:
+			with urllib.request.urlopen(
+				f'{NODE_URL}{status_path}'
+			) as response:
+				status = json.loads(response.read().decode())
+				if status['group'] == 'confirmed':
+					print(f'  Hash lock confirmed')
+					break
+		except urllib.error.HTTPError as e:
+			print(f'  Hash lock status: unknown | Cause: ({e.msg})')
```
segfaultxavi: Should probably rethrow?

mkdocs/overrides/devbook/transactions/bonded-aggregate.py
```diff
@@ -0,0 +1,231 @@
...
+	print(f'Waiting for hash lock confirmation...')
+	status_path = f'/transactionStatus/{hash_lock_hash}'
+	for attempt in range(60):
+		time.sleep(1)
+		try:
+			with urllib.request.urlopen(
+				f'{NODE_URL}{status_path}'
+			) as response:
+				status = json.loads(response.read().decode())
+				if status['group'] == 'confirmed':
```
segfaultxavi: We probably want to throw something like "Hash lock failed to confirm" if status is not 'confirmed'.

mkdocs/overrides/devbook/transactions/bonded-aggregate.py
```diff
@@ -0,0 +1,231 @@
...
+			) as response:
+				status = json.loads(response.read().decode())
+				if status['group'] == 'partial':
+					print(
+						'  Transaction is partial, '
+						'ready for cosignatures'
+					)
+					break
+		except urllib.error.HTTPError as e:
+			print(f'  Transaction status: unknown | Cause: ({e.msg})')
```
segfaultxavi: Same comments as for the previous confirmation loop regarding errors.

mkdocs/overrides/devbook/transactions/bonded-aggregate.py
```diff
@@ -0,0 +1,231 @@
...
+
+	# Submit Account B's cosignature
+	print('Submitting Account B\'s cosignature...')
+	cosignature_path = '/transactions/cosignature'
+	cosignature = facade.cosign_transaction(
+		account_b_key_pair, bonded_transaction, True)
```
segfaultxavi: If you use `bonded_transaction` directly, it means that Account B has had access to it off-chain.
This is OK for the Complete Aggregate tutorial, but I think this one should do everything on-chain.

You knew I was going to ask for this :D

So, at the very least, the transaction to cosign should be retrieved from the chain, using its hash (which would be provided off-chain).
But even better, Account B can monitor partial transactions waiting for its signature using /transactions/partial (GET).
This is an example, so there is no need to poll (we know the partial transaction will be there at this point),
but the text should explain that Account B can periodically check if there are partial transactions waiting for its signature.

mkdocs/overrides/devbook/transactions/bonded-aggregate.py
```diff
@@ -0,0 +1,231 @@
...
+	})
+
+	cosig_request = urllib.request.Request(
+		f'{NODE_URL}{cosignature_path}',
+		data=cosignature_payload.encode(),
+		headers={'Content-Type': 'application/json'},
+		method='PUT'
+	)
+	with urllib.request.urlopen(cosig_request) as response:
+		print(f'  Cosignature from Account B: {response.read().decode()}')
```
segfaultxavi: These prints should follow the same format as the previous ones:
```
Announcing cosignature to {cosignature_path}
  Response: {response}
```

mkdocs/overrides/devbook/transactions/bonded-aggregate.py
```diff
@@ -0,0 +1,231 @@
...
+		try:
+			with urllib.request.urlopen(
+				f'{NODE_URL}{status_path}'
+			) as response:
+				status = json.loads(response.read().decode())
+				print(f'  Transaction status: {status["group"]}')
+				if status['group'] == 'confirmed':
+					print(f'Transaction confirmed in {attempt} seconds')
+					break
+				if status['group'] == 'failed':
```
segfaultxavi: Yeah, this is the kind of extended error handling that's missing in the previous polls.
Maybe move the Submit and Wait for confirmation can be moved to a function, like you did for the Message tutorials.

mkdocs/pages/en/devbook/transactions/bonded-aggregate.md
```diff
@@ -0,0 +1,241 @@
...
+        tooltip = "Bonded Aggregate Transaction";
+        subgraph clusterT2 {
+            label = "Embedded Transfer 2";
+            tooltip = "Embedded Transfer 2";
+            style = dashed;
+            A1 [label="Account A" tooltip="Account A"];
+            B1 [label="Account B" tooltip="Account B"];
+            A1 -> B1 [label="1 Custom Mosaic" dir=back];
+        }
+        subgraph clusterT1 {
```
segfaultxavi, line 19: Move clusterT1 before clusterT2.

mkdocs/pages/en/devbook/transactions/bonded-aggregate.md
```diff
@@ -0,0 +1,241 @@
...
+            A2 [label="Account A" tooltip="Account A"];
+            B2 [label="Account B" tooltip="Account B"];
+            A2 -> B2 [label="10 XYM"];
+        }
+
+    }
+}
+```
+
+Unlike <complete aggregate transactions:>, bonded aggregates collect <cosignatures:> on-chain.
```
segfaultxavi: Like with the Complete Aggregate tutorial, I think explaining this difference in an info box would be nice.
And also provide a concrete example where a bonded agg works better, e.g., when cosignatures have to be provided by different companies using different tools and programming languages. In this case, the blockchain acts as the common interface.

mkdocs/pages/en/devbook/transactions/bonded-aggregate.md
```diff
@@ -0,0 +1,241 @@
...
+Each required party submits their cosignature directly to the network, without needing to coordinate or
+communicate with the other parties off-chain.
+
+## Prerequisites
+
+Before you start, make sure to:
+
+- Set up your development environment.
```
segfaultxavi: Asterisks instead of hyphens for unordered lists.

mkdocs/pages/en/devbook/transactions/bonded-aggregate.md
```diff
@@ -0,0 +1,241 @@
...
+  See [Setting Up a Development Environment](../start/setup.md).
+- Create an <account:> to initiate the aggregate transaction, either
+  [from code](../accounts/create-from-private-key.md) or
+  [by using a wallet](../../userbook/wallet/create-account.md).
+- Obtain <XYM:> to pay for the transaction fee, transfer amounts, and the <hash lock:> deposit.
```
segfaultxavi: Would be good to explain here that the tutorial uses default testnet accounts which are already funded, but these funds will eventually run out.
And also comment on the custom mosaic, which is also pre-funded.

segfaultxavi (changes requested):
I've requested some changes to the program which will result in changes to the text, so we'll continue the review when that is implemented.

Also, this tutorial is very similar to the Complete Aggregate one, so make sure relevant comments I made there are applied here too.

mkdocs/pages/en/devbook/transactions/bonded-aggregate.md
zero4862: > funds will eventually run out

Opted to not add this here since we are not doing so for any other tutorial.
Think it's best if they try to use their accounts.

mkdocs/pages/en/devbook/transactions/bonded-aggregate.md
```diff
@@ -0,0 +1,299 @@
...
+
+This hash is required for the next step: creating a hash lock transaction.
+
+### Creating the Hash Lock
+
+{{ tutorial.code_snippet(['py:160:186', 'js:167:196']) }}
+
+Before announcing a bonded aggregate, a hash lock transaction must be created and confirmed.
+The hash lock serves as a deposit to prevent spam and ensure network resources are not exhausted by unfinished
+<partial transactions:>.
```
segfaultxavi:
```suggestion
<bonded aggregate transaction:|partial transactions>.
```

mkdocs/pages/en/devbook/transactions/bonded-aggregate.md
```diff
@@ -0,0 +1,299 @@
...
+          B1 [label="Account B" tooltip="Account B"];
+          A1 -> B1 [label="1 Custom Mosaic" dir=back];
+        }
+    }
+}
+```
+
+!!! note "Two types of aggregate transactions"
+
+    <Aggregate transactions:> group multiple <transactions:> in single operation, and require <signatures:> from all
```
segfaultxavi:
```suggestion
    <Aggregate transactions:> group multiple <transactions:> in a single operation, and require <signatures:> from all
```

mkdocs/pages/en/devbook/transactions/bonded-aggregate.md
```diff
@@ -0,0 +1,299 @@
...
+    involved accounts.
+
+    A _bonded aggregate transaction_ collects signatures on-chain after being announced.
+    This works well when off-chain coordination is impractical:
```
segfaultxavi:
```suggestion
    This works well when off-chain coordination is impractical. For example:
```

mkdocs/overrides/devbook/transactions/bonded-aggregate.py
```diff
@@ -0,0 +1,261 @@
...
+		'mosaic': {
+			'mosaic_id': generate_mosaic_alias_id('symbol.xym'),
+			'amount': 10_000_000  # 10 XYM deposit
+		},
+		'duration': 100,  # Lock duration in blocks
+		'hash': bonded_hash
+	})
+	hash_lock.fee = Amount(fee_mult * hash_lock.size)
+
+	# Sign and announce hash lock
```
segfaultxavi:
```suggestion
	# Sign hash lock
```

mkdocs/overrides/devbook/transactions/bonded-aggregate.py
```diff
@@ -0,0 +1,261 @@
...
+import time
+import urllib.request
+
+from symbolchain.CryptoTypes import Hash256, PrivateKey
+from symbolchain.facade.SymbolFacade import SymbolFacade
+from symbolchain.sc import Amount
+from symbolchain.symbol.IdGenerator import generate_mosaic_alias_id
+from symbolchain.symbol.Network import NetworkTimestamp
+
+NODE_URL = 'https://001-sai-dual.symboltest.net:3001'
```
segfaultxavi: Should fetch NODE_URL env var as in the other tutorials.

mkdocs/overrides/devbook/transactions/bonded-aggregate.py
```diff
@@ -0,0 +1,261 @@
...
+						print(f'  Transaction status: {status["group"]}')
+						print(
+							f'{label} confirmation in ' +
+							f'{attempt + 1} seconds')
+					elif expected_status == 'partial':
+						print(
+							'  Transaction is partial, '
+							'ready for cosignatures'
+						)
+					return attempt + 1
```
segfaultxavi: Why do you return this if nobody is using it?

mkdocs/overrides/devbook/transactions/bonded-aggregate.py
```diff
@@ -0,0 +1,261 @@
...
+				if status['group'] == 'failed':
+					print(f'  Transaction status: {status["group"]}')
+					print(f'{label} failed: {status["code"]}')
+					return attempt + 1
```
segfaultxavi: This should throw an exception so the program halts. Same thing with `HTTPError` and timeouts.

Otherwise, we print the error and continue execution, which is confusing.

mkdocs/pages/en/devbook/transactions/bonded-aggregate.md
```diff
@@ -0,0 +1,299 @@
...
+
+* Set up your development environment.
+  See [Setting Up a Development Environment](../start/setup.md).
+* Create an <account:> (Account A) to initiate the aggregate transaction, either
+  [from code](../accounts/create-from-private-key.md) or
+  [by using a wallet](../../userbook/wallet/create-account.md).
+* Create a second account (Account B) to participate in the swap.
+* Obtain <XYM:> for Account A to pay for the transaction fee, transfer amounts, and the hash lock deposit.
+  See [Getting Testnet Funds from the Faucet](../accounts/testnet-faucet.md).
+* Create a <mosaic:> owned by Account B for the swap.
```
segfaultxavi: Same comment as for the complete agg tx tutorial regarding the length of this list and the pre-funded accounts.

mkdocs/pages/en/devbook/transactions/bonded-aggregate.md
```diff
@@ -0,0 +1,299 @@
...
+Once the hash lock is confirmed, the bonded aggregate is announced to <put:/transactions/partial> using the
+`announce_transaction` helper.
+
+The <node:> validates the transaction, checks that a valid hash lock exists, and places it in a partial state.
+The `wait_for_status` helper monitors the transaction until it reaches this state, at which point it can collect
+cosignatures.
+
+### Submitting Cosignatures
+
+{{ tutorial.code_snippet(['py:198:252', 'js:206:267']) }}
```
segfaultxavi: This snippet is too long. As a rule of thumb, it should fit on a screen so it is easy to understand at a glance.
I would split it in 3:
1. Recovering the transaction:
    - You also need to explain that the tx hash is shared off-chain to make sure the tx we recover from the partials list is the one we want to sign, but Account B could use other means.
    - You should explain the difference between `partial_tx_hash` and `bonded_hash`. It is not clear to me.
2. Verifying it:
    - Not much to talk about, just say that this is where actual verification would happen, but we are only showing the number of embedded txes.
4. Signing it:
    - You can merge this one with the next one, Waiting for confirmation, as you did when announcing the bonded tx.

mkdocs/pages/en/devbook/transactions/bonded-aggregate.md
```diff
@@ -0,0 +1,299 @@
...
+
+Unlike complete aggregates where the transaction payload is shared off-chain, bonded aggregates enable true on-chain
```
segfaultxavi:
```suggestion
Unlike complete aggregates where the transaction payload is shared off-chain, bonded aggregates enable fully on-chain
```

"True" is a bit ambiguous. Also, it's not fully on-chain because we share the hash off-chain.

mkdocs/pages/en/devbook/transactions/bonded-aggregate.md
```diff
@@ -0,0 +1,299 @@
...
+coordination.
+Account B discovers and retrieves the partial transaction directly from the blockchain.
+
+First, Account B polls <get:/transactions/partial> with the `address` parameter to find transactions waiting for their
```
segfaultxavi:
```suggestion
First, Account B polls <get:/transactions/partial> using its own address to find transactions waiting for its
```

I'm happy calling an account "it" :)

mkdocs/pages/en/devbook/transactions/bonded-aggregate.md
```diff
@@ -0,0 +1,299 @@
...
+signature.
+This returns a list of partial transactions involving Account B.
+
+Once a transaction is found, Account B uses its hash to fetch the full details (including embedded transactions) from
+<get:/transactions/partial/{transactionId}>.
+
+After verifying the transaction content, Account B cosigns it using <dy:SymbolFacade.cosignTransactionHash> with the
+transaction hash and the `detached` parameter set to `true` (Python) or passed as the third argument (JavaScript).
```
segfaultxavi: You are describing here what the code does, which is fine, but it's more important to explain _why_.

So what is this `detached` parameter that you talk about? Why do we need it set to true? When would we set it to false?
This is the added value of the text descriptions, because this info is not in the code.

Also, no need to give Python and JS details. Just mention the parameter name, and readers can click on the method name to find it.

mkdocs/pages/en/devbook/transactions/bonded-aggregate.md
```diff
@@ -0,0 +1,299 @@
...
+
+## Output
+
+The output shown below corresponds to a typical run of the program.
+
+```text linenums="1" hl_lines="14 18 48 51 55 63 66 68 70 72"
+--8<-- 'devbook/transactions/bonded-aggregate.log'
+```
+
+Key points in the output:
```
segfaultxavi, line 297: Include line numbers.

segfaultxavi (changes requested):
Nice one! This one's longer, so more comments are to be expected :)
Also, check if any comments to the complete agg tx tutorial apply here.

mkdocs/pages/en/devbook/transactions/bonded-aggregate.md
```diff
@@ -0,0 +1,331 @@
...
+## Full Code
+
+{% import 'tutorial.jinja2' as tutorial with context %}
+
+{{ tutorial.code_full('devbook/transactions/bonded-aggregate', ['py', 'js']) }}
+
+The whole code is wrapped in a single `try` block to provide simple error handling,
+but applications will probably want to use more fine-grained control.
+
+## Account A: Initiator Workflow
```
zero4862, line 87: @segfaultxavi Split "Code explanation" in:

```
## Account A: Initiator Workflow

## Account B: Cosigner Workflow
```

To make it easier to follow.

But, removed "Code explanation" to keep it 3 levels max.

mkdocs/pages/en/devbook/transactions/bonded-aggregate.md
```diff
@@ -0,0 +1,331 @@
...
+Account B cosigns the transaction using <dy:SymbolFacade.cosignTransactionHash> with the transaction hash and the
+`detached` parameter set to `true`.
+
+A detached cosignature is a standalone object that can be submitted independently to the network.
+This is required for bonded aggregates because the cosigner submits directly to the node.
+
+The resulting detached cosignature payload includes:
+
+* **Version:** The cosignature format version.
+* **Signer pubic key:** Account B's public key, identifying who cosigned.
```
segfaultxavi:
```suggestion
* **Signer public key:** Account B's public key, identifying who cosigned.
```

mkdocs/overrides/devbook/transactions/bonded-aggregate.py
```diff
@@ -0,0 +1,257 @@
...
+		if not partial_txs['data']:
+			raise Exception('No partial transactions found')
+
+	print(f'Found {len(partial_txs["data"])} partial transaction(s)')
+
+	# Find the transaction matching our expected hash
+	partial_tx_hash = None
+	for tx in partial_txs['data']:
+		if tx['meta']['hash'] == str(bonded_hash):
+			partial_tx_hash = tx['meta']['hash']
```
segfaultxavi: Wait. So `bonded_hash` and `partial_tx_hash` are the same thing?
This is very confusing, I assumed they were different hashes and was reading the spec trying to find what they are :D

You can raise an exception if `bonded_hash` is not found in `partial_txs`, and proceed otherwise.
But don't use another var holding the same hash.

segfaultxavi (changes requested):

segfaultxavi:
Also, please rebase :)

mkdocs/pages/en/devbook/transactions/bonded-aggregate.md
segfaultxavi: Good idea!

segfaultxavi (approved):
🎉

PR #1784, opened 2025-11-13
[sdk/javascript] fix: noble imports and prohibit string in hash functions

Wayonb:
problem: @noble/hashes 2.0.1 enforces the correct import path, plus prohibits strings in hash functions
solution: update imports of sha256 and sha512 to sha2.js
                convert all strings to byte array for hashes


Jaguar0625 (approved):
👍

PR #1790, opened 2025-11-18
[monorepo] fix: configure multi-directory for dependabot

Wayonb:
problem: currently, Dependabot creates a PR for each project, which leads to a lot of PRs
solution: configure multi-directory for Dependabot, that allows grouping of multiple projects with the same package manager

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -2,13 +2,14 @@ version: 2
 updates:
 
   - package-ecosystem: github-actions
-    directory: /
     schedule:
       interval: weekly
       day: sunday
-    target-branch: dev
+    labels: ['dependencies']
     commit-message:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yaml
```diff
@@ -18,179 +19,62 @@ updates:
...
-      interval: weekly
-      day: sunday
-    target-branch: dev
-    labels: [sdk-python]
-    versioning-strategy: increase-if-necessary
-    commit-message:
-      prefix: '[dependency]'
     groups:
-      sdk-python-dependencies:
+      wasm-dependencies:
```
Jaguar0625: wasm => cargo might be more consistent with other naming in file?

Jaguar0625 (approved):
👍

PR #1803, opened 2025-12-15
[mkdocs] docs: add account metadata tutorial

zero4862:
Adds tutorial on how to add and modify account metadata.


mkdocs/pages/en/devbook/accounts/account-metadata.md
```diff
@@ -0,0 +1,202 @@
...
+The transaction hashes printed in the confirmation messages can be used to search for the transactions
+in the [Symbol Testnet Explorer](https://testnet.symbol.fyi/).
+
+## Conclusion
+
+This tutorial showed how to:
+
+| Step                                                                                  | Related documentation                              |
+| ------------------------------------------------------------------------------------- | -------------------------------------------------- |
+| [Define metadata key and value](#defining-the-metadata)                               | <dy:metadataGenerateKey>                           |
```
zero4862: @segfaultxavi `dy:metadataGenerateKey` and `dy:metadataUpdateValue` links do not seem to work. Could you please review?

mkdocs/pages/en/devbook/transactions/complete-aggregate.md
```diff
@@ -105,9 +105,8 @@ configuration.
 
 {{ tutorial.code_snippet(['py:39:57', 'js:34:52']) }}
 
-To prepare an aggregate, first retrieve the current network time from <get:/node/time> and the recommended fee
-multiplier from <get:/network/fees/transaction>, following the same steps described in the
-[Transfer Transaction](./transfer.md) tutorial.
+Network time and recommended fees are fetched from <get:/node/time> and <get:/network/fees/transaction> respectively,
```
zero4862, line 108: To make it less imperative and more descriptive.

mkdocs/overrides/devbook/accounts/account-metadata.py
segfaultxavi, line 1: If the md file is in the `accounts` folder, then the source files should be in `overrides/devbook/accounts`.

mkdocs/pages/en/devbook/accounts/account-metadata.md
```diff
@@ -0,0 +1,202 @@
...
+title: Account Metadata
+---
+
+# Adding Metadata to an Account
+
+<Accounts:> can store <metadata:> as key-value pairs.
+
+This tutorial shows how to add metadata to an account, retrieve metadata from the network, and update existing values.
+
+In this example, an account adds a username to itself and then modifies it.
```
segfaultxavi: A `username` looks like a specific property of accounts and might be confusing. How about:

```suggestion
In this example, the string `alice` is attached an account and then changed to `bob`.
```

mkdocs/pages/en/devbook/accounts/account-metadata.md
segfaultxavi: These are global functions that were not handled by my `<dy:>` code, sorry!
It's fixed in the `new-docs` branch, you'll need to rebase and use `<dy:Metadata.metadataGenerateKey>`.

It's pretty hackish, but this will generate `<js:metadataGenerateKey>` in JS and `<py:Metadata.metadata_generate_key>` in Python.

mkdocs/pages/en/devbook/accounts/account-metadata.md
```diff
@@ -0,0 +1,202 @@
...
+
+Network time and recommended fees are fetched from <get:/node/time> and <get:/network/fees/transaction> respectively, following the process described in the [Transfer Transaction](./transfer.md) tutorial.
+
+### Defining the Metadata
+
+{{ tutorial.code_snippet(['py:84:87', 'js:82:85']) }}
+
+Each metadata entry is identified by a **scoped metadata key**, a 64-bit value chosen by the metadata creator.
+The combination of the signer's address, the target account's address, and the scoped metadata key uniquely
+identifies the metadata entry.
```
segfaultxavi:
```suggestion
Each metadata entry is uniquely identified by the signer's address, the target account's address, and a **scoped metadata key**: a 64-bit value chosen by the metadata creator.
```

mkdocs/pages/en/devbook/accounts/account-metadata.md
```diff
@@ -0,0 +1,202 @@
...
+
+!!! note "Multiple entries with the same key"
+
+    Because the signer's address is part of the unique identifier, multiple accounts can add metadata with the same
+    scoped key to a target (e.g., Account A and Account B can add a different `username` to Account C).
+    Each entry remains separate and can only be updated by its signer.
```
segfaultxavi:
```suggestion
    Because the signer's address is part of the unique identifier, different accounts can use the same scoped metadata key 
    on the same target account without conflict.
    
    For example, Account A and Account B can both use the key `username` when adding metadata to Account C,
    resulting in two distinct metadata entries.

    Each entry is independent and can only be updated by the account that originally created it.
```

mkdocs/pages/en/devbook/accounts/account-metadata.md
```diff
@@ -0,0 +1,202 @@
...
+The SDK provides a <dy:metadataGenerateKey> helper function that generates a key from a
+human-readable string using SHA3-256 hashing.
+This approach makes keys more meaningful and reduces the chance of collisions.
+
+In this example, the key is derived from the string `username`.
+To ensure the metadata key is unique across multiple runs of the tutorial, a timestamp is appended to the key string.
+In practice, you would use a stable key that identifies the specific metadata entry you want to create or update.
+
+The metadata value can be any byte sequence.
+In this example, the username `alice` is UTF-8 encoded.
```
segfaultxavi:
```suggestion
In this example, the value is the string `alice` encoded in UTF-8.
```

mkdocs/pages/en/devbook/accounts/account-metadata.md
```diff
@@ -0,0 +1,202 @@
...
+
+### Creating the Embedded Metadata Transaction
+
+{{ tutorial.code_snippet(['py:89:101', 'js:87:100']) }}
+
+Symbol requires metadata transactions to be inside an <aggregate transaction:>.
```
segfaultxavi: It would be good to explain why.
I understand this is useful when the signer and the target accounts are different, but it looks cumbersome when they are not.
The Symbol Whitepaper does not mention metadata anywhere, so you'll have to ask in the chat.

mkdocs/pages/en/devbook/accounts/account-metadata.md
```diff
@@ -0,0 +1,202 @@
...
+
+For this reason, the code defines the metadata update as an <embedded transaction:>.
```
segfaultxavi: You have not talked about updates yet, this section is about adding new metadata.

I know it's the same transaction, but the reader might not. So you better not talk about updates yet, or explain now that additions and updates are the same thing.
If you don't do it now you'll have to do it later.
Doing it now has the benefit that explaining below why Value and Value size delta are so weird is easier.

mkdocs/pages/en/devbook/accounts/account-metadata.md
```diff
@@ -0,0 +1,202 @@
...
+The account metadata transaction specifies:
+
+* **Type:** Use `account_metadata_transaction_v1`.
+
+* **Signer public key:** The account creating the metadata entry.
+
+* **Target address:** The account to attach the metadata to.
+  When the target differs from the signer, the target account must cosign the aggregate transaction.
+
+* **Scoped metadata key:** The 64-bit identifier for this metadata entry.
```
segfaultxavi:
```suggestion
* **Scoped metadata key:** The 64-bit key used to identify this metadata entry.
```

Previously we separated the key from the identifier that is obtained by combining the key with the addresses, so let's keep that nomenclature to avoid confusion.

mkdocs/pages/en/devbook/accounts/account-metadata.md
```diff
@@ -0,0 +1,202 @@
...
+  [Modifying Existing Metadata](#modifying-existing-metadata) section).
+
+### Building the Aggregate Transaction
+
+{{ tutorial.code_snippet(['py:103:115', 'js:102:114']) }}
+
+The code adds the embedded transaction to a complete <aggregate transaction:>.
+
+Since the signer is modifying their own account, the transaction does not need cosignatures
+and the aggregate can be signed and announced immediately.
```
segfaultxavi: Mention that therefore this is an aggregate complete tx.

mkdocs/pages/en/devbook/accounts/account-metadata.md
```diff
@@ -0,0 +1,202 @@
...
+[Creating a Complete Aggregate Transaction](../transactions/complete-aggregate.md#building-the-aggregate-transaction).
+
+### Retrieving Metadata
+
+{{ tutorial.code_snippet(['py:130:147', 'js:130:147']) }}
+
+Updating an existing metadata entry requires the current value from the network.
+
+The code queries the <get:/metadata> endpoint with filters for `sourceAddress`, `targetAddress`,
+`scopedMetadataKey`, and `metadataType` (`0` for account metadata) to retrieve the specific entry.
```
segfaultxavi, line 154: Good that you mention that 0 is for account metadata, because the REST API docs say nothing about it 😩 

Are you interested in reviewing the REST API docs? 🙂

mkdocs/pages/en/devbook/accounts/account-metadata.md
```diff
@@ -0,0 +1,202 @@
...
+
+### Modifying Existing Metadata
```
segfaultxavi, line 156: Worth mentioning here that updating old metadata uses the same tx as adding new metadata, if you haven't explained it yet.

mkdocs/overrides/devbook/accounts/account-metadata.py
```diff
@@ -0,0 +1,191 @@
...
+	json_payload = facade.transaction_factory.attach_signature(
+		transaction, signature)
+
+	# Announce and wait for confirmation
+	transaction_hash = facade.hash_transaction(transaction)
+	announce_transaction(json_payload, 'aggregate transaction')
+	wait_for_confirmation(transaction_hash, 'Aggregate transaction')
+
+	# --- MODIFYING EXISTING METADATA ---
+	print('\n--- Modifying existing metadata ---')
```
segfaultxavi, line 130: A similar comment and output line for adding new metadata above would help separate the two different operations.

mkdocs/overrides/devbook/transactions/account-metadata.py
```diff
@@ -0,0 +1,191 @@
...
+		'type': 'aggregate_complete_transaction_v3',
+		'signer_public_key': signer_key_pair.public_key,
+		'deadline': timestamp.add_hours(2).timestamp,
+		'transactions_hash': facade.hash_embedded_transactions(
+			embedded_transactions),
+		'transactions': embedded_transactions
+	})
+	transaction.fee = Amount(fee_mult * transaction.size)
+	print('Built aggregate transaction:')
+	print(json.dumps(transaction.to_json(), indent=2))
```
segfaultxavi: We already explained the details of agg txes in another tutorial. Dumpling them here (and again below) makes this log too long. Maybe just printing the hash so it can be looked up in the explorer is enough.

segfaultxavi (changes requested):

mkdocs/overrides/devbook/accounts/account-metadata.py
zero4862: Good catch!

mkdocs/pages/en/devbook/accounts/account-metadata.md
zero4862: It is described, but the Swagger UI does not render the enum description.

image

I would definitely like to review the REST API docs. Let's add it to our list of follow-up items and discuss it in more detail later.

mkdocs/pages/en/devbook/accounts/account-metadata.md
```diff
@@ -0,0 +1,211 @@
...
+
+In this tutorial, the signer adds metadata to their own account.
+Adding metadata to a different account requires the target to cosign the transaction.
+
+### Fetching Network Time and Fees
+
+{{ tutorial.code_snippet(['py:64:82', 'js:62:80']) }}
+
+Network time and recommended fees are fetched from <get:/node/time> and <get:/network/fees/transaction> respectively,
+following the process described in the [Transfer Transaction](./transfer.md) tutorial.
```
segfaultxavi:
```suggestion
following the process described in the [Transfer Transaction](../transactions/transfer.md) tutorial.
```
There is a yellow warning in the mkdocs output log due to the missing file.

mkdocs/pages/en/devbook/accounts/account-metadata.md
```diff
@@ -0,0 +1,211 @@
...
+
+    Each entry is independent and can only be updated by the account that originally created it.
+
+The SDK provides a <dy:Metadata.metadataGenerateKey> helper function that generates a key from a
+human-readable string using SHA3-256 hashing.
+This approach makes keys more meaningful and reduces the chance of collisions.
+
+In this example, the key is derived from the string `username`.
+To ensure the metadata key is unique across multiple runs of the tutorial, a timestamp is appended to the key string.
+In practice, you would use a stable key that identifies the specific metadata entry you want to create or update.
```
segfaultxavi:
```suggestion
For demonstration purposes, a timestamp is appended to the key string,
so each time the code is executed a new entry is added to the account.
In practice, you would use a fixed key that identifies the specific metadata entry you want to create or update.
```

I made it extra clear that the timestamp is only required for the tutorial and is not something users will normally need to do.

mkdocs/pages/en/devbook/accounts/account-metadata.md
```diff
@@ -0,0 +1,211 @@
...
+
+The metadata value can be any byte sequence.
+In this example, the value is the string `alice` encoded in UTF-8.
+
+### Creating the Embedded Account Metadata Transaction
+
+{{ tutorial.code_snippet(['py:92:104', 'js:90:103']) }}
+
+An account metadata transaction attaches a key-value pair to an account on the blockchain.
+The same transaction type handles both adding new metadata and updating existing entries.
```
segfaultxavi:
```suggestion
The same transaction type handles both adding new metadata entries and updating existing ones.
```

Otherwise it looks like "metadata" and "entries" are different things.

mkdocs/pages/en/devbook/accounts/account-metadata.md
```diff
@@ -0,0 +1,211 @@
...
+
+Symbol requires these transactions to be inside an <aggregate transaction:> to prevent third parties from attaching
+unwanted metadata to an account without permission.
+Even when target and signer are the same account, an aggregate is still required to keep the transaction format
+uniform.
```
segfaultxavi:
```suggestion
Symbol requires these transactions to be inside an <aggregate transaction:> that includes the target account owner's signature.  
This prevents unwanted metadata to be attached to an account without its owner's permission.  
An aggregate is still required even when the transaction is initiated by the account owner,
to keep the transaction format uniform.
```

Clarified.

mkdocs/pages/en/devbook/accounts/account-metadata.md
```diff
@@ -0,0 +1,211 @@
...
+
+For this reason, the code defines the account metadata transaction as an <embedded transaction:>.
+This transaction specifies:
+
+* **Type:** Use `account_metadata_transaction_v1`.
+
+* **Signer public key:** The account creating the metadata entry.
+
```
segfaultxavi, line 108:
```suggestion
    In this case, this is the account receiving the metadata too.

```

mkdocs/pages/en/devbook/accounts/account-metadata.md
```diff
@@ -0,0 +1,211 @@
...
+* **Value:** The metadata content as bytes.
+  When creating new metadata, provide the raw value.
+  When updating, provide a computed value (explained in the
+  [Modifying Existing Metadata](#modifying-existing-metadata) section).
+
+### Building the Aggregate Transaction
+
+{{ tutorial.code_snippet(['py:106:116', 'js:105:115']) }}
+
+The code adds the embedded account metadata transaction to a <aggregate transaction:>.
```
segfaultxavi:
```suggestion
The code adds the embedded account metadata transaction to an <aggregate transaction:>.
```

mkdocs/pages/en/devbook/accounts/account-metadata.md
```diff
@@ -0,0 +1,211 @@
...
+
+Since the signer is modifying their own account, no <cosignatures:> are needed and the aggregate can be of type
+complete, which can be signed and announced immediately.
```
segfaultxavi:
```suggestion
Since the signer is modifying its own account, no <cosignatures:> are required and the aggregate can be created as
<complete aggregate transaction:|complete>, allowing it to be signed and announced immediately.
```

mkdocs/pages/en/devbook/accounts/account-metadata.md
```diff
@@ -0,0 +1,211 @@
...
+Updating an existing metadata entry requires the current value from the network.
+
+The code queries the <get:/metadata> endpoint with filters for `sourceAddress`, `targetAddress`,
+`scopedMetadataKey`, and `metadataType` (`0` for account metadata) to retrieve the specific entry.
+
+### Modifying Existing Metadata
+
+{{ tutorial.code_snippet(['py:151:177', 'js:152:180']) }}
+
+To demonstrate updating metadata, the code changes the username from `alice` to `bob` using an embedded account
```
segfaultxavi:
```suggestion
To demonstrate updating metadata, the code changes the username from `alice` to `bob` using another account
```

mkdocs/pages/en/devbook/accounts/account-metadata.md
```diff
@@ -0,0 +1,211 @@
...
+title: Account Metadata
+---
+
+# Adding Metadata to an Account
+
+<Accounts:> can store <metadata:> as key-value pairs.
+
+This tutorial shows how to add metadata to an account, retrieve metadata from the network, and update existing values.
+
+In this example, the string `alice` is attached an account and then changed to `bob`.
```
segfaultxavi:
```suggestion
In this example, the string `alice` is attached to an account and then changed to `bob`.
```

Do not accept my suggestions without reviewing them :P

segfaultxavi (changes requested):
Please always use 4 spaces to indent, even on lists and paragraphs, just like you do for notes.
Right now it works with 2 spaces, but I've found mkdocs plugins that break if we don't follow the Markdown standard exactly.

Only minor changes!

segfaultxavi (approved):
All good!

Except one repeated sentence that I guess it's a leftover. I'll fix it in another commit.

PR #1810, opened 2025-12-23
[mkdocs] docs: add namespace tutorials

zero4862:
Adds tutorials on how to:

- register a root namespace
- extend a root namespace
- create a subnamespace


mkdocs/pages/en/devbook/namespaces/register-root-namespace.md
```diff
@@ -0,0 +1,137 @@
...
+test key if not set.
+The signer's address is derived from the public key.
+This account will own the registered namespace.
+
+### Fetching Network Time and Fees
+
+{{ tutorial.code_snippet(['py:27:45', 'js:24:43']) }}
+
+Network time and recommended fees are fetched from <get:/node/time> and <get:/network/fees/transaction> respectively,
+following the process described in the [Transfer Transaction](./transfer.md) tutorial.
```
segfaultxavi:
```suggestion
following the process described in the [Transfer Transaction](../transactions/transfer.md) tutorial.
```

mkdocs/pages/en/devbook/namespaces/register-root-namespace.md
```diff
@@ -0,0 +1,137 @@
+---
+title: Register Root Namespace
+---
+
+# Registering a Root Namespace
+
+<Namespaces:> provide human-readable aliases for <mosaics:> and <accounts:>.
+
+This tutorial shows how to register a root namespace and set its
+<lease duration:../../textbook/namespaces.md#duration>.
```
segfaultxavi: This is the first time "root namespace" is used so it needs to be defined.
The best way to do it is to just link to its glossary definition, but we don't have one.
So my suggestion is that you add definitions for "root namespace" and "subnamespace" in the Textbook page about Namespaces, right after the paragraph that starts with "The first part is called the root namespace."

Something like (please extend):

```markdown
Root Namespace
:   A <namespace:> that has no parent and can be used on its own.
    It can be used to group <subnamespaces:> together in a hierarchical manner.
Subnamespace
:   A <namespace:> that belongs to a parent <root namespace:>.
```

You can then remove the following paragraph in the Textbook and move its information (about the number of subnamespaces and their duration) to the above two definitions.

This would also fix the missing definition for <root namespace:> you have in the Registering a Subnamespace page.

mkdocs/overrides/devbook/namespaces/register-root-namespace.py
```diff
@@ -0,0 +1,120 @@
...
+			status_url = (
+				f'{NODE_URL}/transactionStatus/{transaction_hash}')
+			with urllib.request.urlopen(status_url) as response:
+				status = json.loads(response.read().decode())
+				print(f'  Transaction status: {status["group"]}')
+			if status['group'] == 'confirmed':
+				print('Namespace registration confirmed in',
+					attempt, 'seconds')
+				break
+				if status['group'] == 'failed':
```
segfaultxavi:
```suggestion
			if status['group'] == 'failed':
```

Indentation error!

mkdocs/pages/en/devbook/namespaces/register-root-namespace.md
```diff
@@ -0,0 +1,137 @@
...
+
+### Building the Transaction
+
+{{ tutorial.code_snippet(['py:47:59', 'js:45:57']) }}
+
+The namespace registration transaction specifies:
+
+* **Type:** Use `namespace_registration_transaction_v1`.
+
+* **Registration type:** The value `root` indicates a root namespace is being created.
```
segfaultxavi, line 62: You can mention the other option, and link to the other tutorial. For completeness.

mkdocs/pages/en/devbook/namespaces/register-root-namespace.md
```diff
@@ -0,0 +1,137 @@
...
+
+* **Duration:** The number of blocks for which the namespace will be leased.
+    The minimum duration is 86,400 blocks (approximately 30 days on mainnet), and the maximum is
+    15,552,000 blocks (approximately 5 years).
+
+* **Name:** The name of the root namespace.
+    Names can only contain lowercase letters, numbers, hyphens, and underscores, must start with a letter or number,
+    and can be at most 64 characters long.
+
+    To ensure the namespace name is unique across multiple runs, the example appends a timestamp to the name.
```
segfaultxavi:
```suggestion
    To ensure the namespace name is unique across multiple runs of the tutorial, a timestamp is added to the name.
    Applications will typically use a fixed name instead.
```

mkdocs/pages/en/devbook/namespaces/register-root-namespace.md
```diff
@@ -0,0 +1,137 @@
...
+
+!!! note "Namespace lease fees"
+
+    Registering a root namespace requires a lease fee in addition to the transaction fee.
+    The lease fee is proportional to the requested duration.
```
segfaultxavi: But I don't see the lease fee being paid in the transaction, only the standard transaction fee:
`transaction.fee = Amount(fee_mult * transaction.size)`

How come the code works?

The tutorial should show how to calculate the lease fee, but we have to figure out what is going on first.

mkdocs/pages/en/devbook/namespaces/register-root-namespace.md
```diff
@@ -0,0 +1,137 @@
...
+To verify the namespace was registered, the code retrieves it from the network
+using the <get:/namespaces/{namespaceId}> endpoint and displays its properties.
+
+The namespace ID is computed using <dy:IdGenerator.generateNamespaceId>.
+This function applies a deterministic hashing algorithm to the namespace name,
+producing the ID needed to query the namespace information.
+
+The response includes:
+
+* **Registration type:** Shows `0` for a namespace.
```
segfaultxavi:
```suggestion
* **Registration type:** Shows `0` for a root namespace.
```

mkdocs/pages/en/devbook/namespaces/register-root-namespace.md
```diff
@@ -0,0 +1,137 @@
...
+* **Owner address:** The account that registered and owns the namespace.
```
segfaultxavi: There's only one account involved in this tutorial, however, I am getting 3 different numbers in the output:
```
Signer address: TCHBDENCLKEBILBPWP3JPB2XNY64OE7PYHHE32I
"signer_public_key": "3B6A27BCCEB6A42D62A3A8D02A6F0D73653215771DE243A63AC048A18B59DA29"
Owner address: 988E1191A25A88142C2FB3F69787576E3DC713EFC1CE4DE9
```
Why doesn't the last one (retrieved from the chain) match the first one?

mkdocs/overrides/devbook/namespaces/register-root-namespace.py
```diff
@@ -0,0 +1,120 @@
...
+				raise Exception('Namespace registration failed:',
+					status['code'])
+		except urllib.error.HTTPError:
+			print('  Transaction status: unknown')
+
+	# Retrieve the namespace
+	namespace_id = generate_namespace_id(namespace_name)
+	print(f'Namespace ID: {hex(namespace_id)}')
```
segfaultxavi: In the transaction dumped on line 66 the id is in decimal so it's hard to see that they match.

mkdocs/pages/en/devbook/namespaces/register-root-namespace.md
```diff
@@ -0,0 +1,137 @@
...
+* **Start height:** The block height when the namespace was registered.
+* **End height:** The block height when the namespace will expire.
```
segfaultxavi: Why doesn't End height - Start height match 86400?

In my own experiments I get the exact same difference, 2880 blocks (1 day). Any idea where that comes from?

mkdocs/pages/en/devbook/namespaces/register-root-namespace.md
```diff
@@ -0,0 +1,137 @@
...
+For root namespaces, you can calculate the remaining duration by subtracting the current block height from the end
+height.
+
+## Output
+
+The output shown below corresponds to a typical run of the program.
+
+```text
+--8<-- 'devbook/namespaces/register-root-namespace.log'
+```
```
segfaultxavi: Add line highlights and explain some of the interesting points :)
For example:
- The chosen namespace name (and how users can look it up in the explorer)
- The fee reflecting the transaction and the lease fees
- The id and the name (explain the difference, and that the name is hex-encoded)
- The star and end heights, and how the difference matches the requested lease (once we know why they don't match)

mkdocs/pages/en/devbook/namespaces/register-root-namespace.md
```diff
@@ -0,0 +1,137 @@
...
+in the [Symbol Testnet Explorer](https://testnet.symbol.fyi/).
+
+## Conclusion
+
+This tutorial showed how to:
+
+| Step                                                          | Related documentation                              |
+| ------------------------------------------------------------- | -------------------------------------------------- |
+| [Generate namespace ID](#building-the-transaction)            | <dy:IdGenerator.generateNamespaceId>               |
+| [Build a namespace registration](#building-the-transaction)   | <dy:SymbolTransactionFactory.create>               |
```
segfaultxavi:
```suggestion
| [Build a namespace registration transaction](#building-the-transaction)   | <dy:SymbolTransactionFactory.create>               |
```

mkdocs/pages/en/devbook/namespaces/register-root-namespace.md
```diff
@@ -0,0 +1,137 @@
...
+| [Retrieve the namespace](#retrieving-the-namespace)           | <get:/namespaces/{namespaceId}>                    |
+
+## Next Steps
+
+Now that you have a root namespace, you can:
+
+- Link your namespace to a mosaic or account to create an alias
```
segfaultxavi, line 166: This point is important enough that I would promote it to an info box at the beginning of the Conclusions section.
Something like:

**⚠ The namespace is registered but it is not in use yet**
Namespaces are really useful when they can be used instead of the cumbersome identifiers required to refer to <mosaics:> and <accounts:>.
To be able to do this, the namespace needs to be linked to one such identifier, as explained in the guides linked below.

mkdocs/pages/en/devbook/namespaces/register-root-namespace.md
```diff
@@ -0,0 +1,137 @@
+---
+title: Register Root Namespace
+---
+
+# Registering a Root Namespace
+
+<Namespaces:> provide human-readable aliases for <mosaics:> and <accounts:>.
```
segfaultxavi:
```suggestion
<Namespaces:> provide human-readable aliases for <mosaics:> and <accounts:>,
which can be used instead of their cumbersome identifiers.
```

mkdocs/pages/en/devbook/namespaces/register-root-namespace.md
```diff
@@ -0,0 +1,137 @@
...
+
+This tutorial shows how to register a root namespace and set its
+<lease duration:../../textbook/namespaces.md#duration>.
+
```
segfaultxavi, line 14: Add a last paragraph to the intro saying that once a namespace is registered, additional steps are required to link it to a mosaic or account, as explained in the Conclusions section.

mkdocs/pages/en/devbook/namespaces/register-subnamespace.md
```diff
@@ -0,0 +1,125 @@
...
+
+- Set up your development environment.
+  See [Setting Up a Development Environment](../start/setup.md).
+- Have <account:> with an existing root namespace.
+  See [Registering a Root Namespace](./register-root-namespace.md).
+
+    !!! note
+        The examples in this tutorial use a root namespace named `ns_root`.
+        Make sure to update the code to use your own root namespace name.
+- Obtain <XYM:> to pay for the transaction and lease fees.
```
segfaultxavi, line 27: This bullet point is rendered indented because it's missing a blank line after the `!!! note`.

segfaultxavi (changes requested):
I've only reviewed the Register Root Namespace tutorial, but some of my comments will apply to the other tutorials too.
So let's finish reviewing this one, apply the comments to the others when needed, and when we're done with the first I'll review the rest.

mkdocs/pages/en/devbook/namespaces/register-root-namespace.md
zero4862: The network [deducts it automatically](https://github.com/symbol/symbol/blob/dev/client/catapult/plugins/txes/namespace/src/plugins/NamespaceRegistrationTransactionPlugin.cpp#L54) when the transaction is confirmed, so it is not needed to specify it in the transaction.

zero4862: Added some more details in the tutorial:

>     The network deducts it automatically when the transaction is confirmed, so you do not need to specify it in the transaction.

mkdocs/pages/en/devbook/namespaces/register-root-namespace.md
zero4862: It's due to the end grade period. For testnet, it is configured to one day.  See [namespaceGracePeriodDuration](https://reference.symboltest.net:3001/network/properties). 

Added some more context:

> * **Start and end heights** (lines 35-36): The namespace is active from block `2984442` to block `3073722`.
    The end height includes a [grace period](../../textbook/namespaces.md#duration) (1 day on <testnet:>,
    30 days on <mainnet:>) beyond the requested duration, giving owners time to renew before the namespace
    becomes available to others.

mkdocs/pages/en/devbook/namespaces/register-root-namespace.md
zero4862: Edited. Explained the interesting points in the conclusion,  but reduced the content in *Retrieve the namespace* to avoid repetition.

mkdocs/pages/en/devbook/namespaces/register-root-namespace.md
zero4862: Now added, but under "Retrieve the namespace".

mkdocs/pages/en/devbook/namespaces/register-root-namespace.md
zero4862: Fixed with:

> 		owner_address_hex = bytes.fromhex(namespace_info["ownerAddress"])
>		owner_address = Address(owner_address_hex)

The response returns the address hash without network byte nor checksum.

mkdocs/pages/en/devbook/namespaces/register-root-namespace.md
segfaultxavi: Ooooohhh... I understand. This was pretty confusing to me, because it's different from how transaction fees work.
I believe it will be confusing to other readers too, so it is worth highlighting the difference even more:

```md
    In addition to the standard [transaction fee](#fetching-network-time-and-fees),
    registering a namespace requires a [lease fee](../../textbook/namespaces.md#lease-fee) proportional to
    the requested duration.

    Unlike the transaction fee, the lease fee is **not** included in the transaction request.
    It is calculated and deducted automatically by the network from the **transaction signer’s account**
    when the registration transaction is confirmed.

    The amount of the lease fee can be calculated beforehand using the <get:/network/fees/rental> endpoint.
```

Make the same changes to the subnamespace tutorial.

mkdocs/pages/en/devbook/namespaces/register-root-namespace.md
```diff
@@ -0,0 +1,165 @@
...
+* **Duration:** The number of blocks for which the namespace will be leased.
+    The minimum duration is 86,400 blocks (approximately 30 days on mainnet), and the maximum is
+    15,552,000 blocks (approximately 5 years).
+
+* **Name:** The name of the root namespace.
+    Names can only contain lowercase letters, numbers, hyphens, and underscores, must start with a letter or number,
+    and can be at most 64 characters long.
+
+    To ensure the namespace name is unique across multiple runs of the tutorial, a timestamp is added to the name.
```
segfaultxavi, line 73:
```suggestion
    To ensure the namespace name is unique across multiple runs of the tutorial, a timestamp is added to the name.
    In practice, programs would use a fixed name for their namespaces.
```

Let's be extra clear, as we were in the Metadata tutorial.
Add the same comment to the subnamespace tutorial.

mkdocs/pages/en/devbook/namespaces/register-subnamespace.md
```diff
@@ -0,0 +1,166 @@
...
+Once registered, additional steps are required to link the namespace to a mosaic or account,
+as explained in [Next Steps](#next-steps).
+
+## Prerequisites
+
+Before you start, make sure to:
+
+- Set up your development environment.
+  See [Setting Up a Development Environment](../start/setup.md).
+- Have <account:> with an existing root namespace.
```
segfaultxavi:
```suggestion
- Have an <account:> with an existing root namespace.
```

mkdocs/overrides/devbook/namespaces/register-root-namespace.py
```diff
@@ -0,0 +1,123 @@
...
+import urllib.request
+
+from symbolchain.CryptoTypes import PrivateKey
+from symbolchain.facade.SymbolFacade import SymbolFacade
+from symbolchain.symbol.Network import NetworkTimestamp
+from symbolchain.symbol.IdGenerator import generate_namespace_id
+from symbolchain.sc import Amount
+from symbolchain.symbol.Network import Address
+
+NODE_URL = os.environ.get(
```
segfaultxavi:
```suggestion
NODE_URL = os.getenv(
```

For consistency with the next line. Also, make sure this is used consistently in all tutorials! I just noticed it is not.

mkdocs/pages/en/devbook/namespaces/register-subnamespace.md
```diff
@@ -0,0 +1,166 @@
...
+
+{{ tutorial.code_full('devbook/namespaces/register-subnamespace', ['py', 'js']) }}
+
+## Code Explanation
+
+Most of the code follows the same pattern as registering a root namespace.
+This section focuses on the key differences.
+
+For detailed explanations of the common steps (setting up the account, fetching network time and fees, and announcing),
+see [Registering a Root Namespace](./register-root-namespace.md).
```
segfaultxavi:
```suggestion
The code follows the same pattern as the [Registering a Root Namespace](./register-root-namespace.md) tutorial.
This section focuses only on the key differences.
```

Good call, avoiding the repetition. I just made the intro a bit shorter.

mkdocs/pages/en/devbook/namespaces/register-subnamespace.md
```diff
@@ -0,0 +1,166 @@
...
+
+* **Start and end heights** (lines 39-40): These values are inherited from the root namespace, not set independently.
+
+The transaction hash printed in the output can also be used to search for the transaction
+in the [Symbol Testnet Explorer](https://testnet.symbol.fyi/).
+
+!!! info "Subnamespace registered but not linked yet"
+
+    A namespace becomes useful when it serves as an alias for a <mosaic:> or an <account:>.
+    Link the namespace to an identifier using the guides in [Next Steps](#next-steps).
```
segfaultxavi: I believe you only wanted to have this info box once, in the previous section.

mkdocs/pages/en/devbook/namespaces/extend-root-namespace.md
```diff
@@ -0,0 +1,30 @@
...
+
+- An <account:> that owns a root namespace.
+- <XYM:> to pay for the transaction and lease fees.
+
+Review the [Registering a Root Namespace](./register-root-namespace.md) tutorial
+for the complete registration process.
+
+## Step by Step
+
+To extend a namespace, follow the [registration process](./register-root-namespace.md) with these parameters:
```
segfaultxavi:
```suggestion
To extend a namespace, repeat the [registration process](./register-root-namespace.md) with these parameters:
```

To emphasize that it's the exact same process.

mkdocs/pages/en/devbook/namespaces/extend-root-namespace.md
```diff
@@ -0,0 +1,30 @@
+---
+title: Extend Root Namespace
+---
+
+# Extending a Root Namespace
+
+This tutorial shows how to extend a root <namespace:> before it expires.
```
segfaultxavi: I agree this can be a very short, codeless tutorial. In addition to what it already has, I would add:

- Explain why you would want to extend a ns: because upon registration the maximum lease time is limited to 5 years, so you need to renew periodically if you want more than that.
- Explain that you can never extend a ns beyond 5 years in the future, you have to wait and renew (please confirm this is true!).
- Highlight that only root namespaces need to be extended, because children inherit the duration of their parents.
- Add links to the Duration section of the Namespaces page.
- Explain that the extension must happen either while the ns is active, and therefore the duration is effectively increased, or during the grace period, in which case the ns becomes active immediately (but double check this).

Do not add all these items to the intro, put them in the appropriate places :D

segfaultxavi (changes requested):
Good! Almost there.
(God, I sound just like ChatGPT 😅)

segfaultxavi (approved):
👍

PR #1825, opened 2026-01-14
[jenkins] fix: force pip to reinstall packages

Wayonb:
problem: internal Python packages with the same version are failing to install
         due to different hashes between Nexus and PyPi.
solution: for private builds, only use the Nexus registry and force pip to reinstall the package.

### DIFF
--- a/jenkins/shared-library/vars/configureArtifactRepository.groovy
+++ b/jenkins/shared-library/vars/configureArtifactRepository.groovy
@@ -64,7 +64,8 @@ void configurePypi(Map info) {
 
 	env.PIP_INDEX_URL = "${info.url}simple"
 	env.PIP_TRUSTED_HOST = "${hostName} pypi.org"
-	env.PIP_EXTRA_INDEX_URL = 'https://pypi.org/simple'
+	env.PIP_NO_CACHE_DIR = 1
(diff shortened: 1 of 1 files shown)


jenkins/shared-library/vars/configureArtifactRepository.groovy
```diff
@@ -64,7 +64,8 @@ void configurePypi(Map info) {
 
 	env.PIP_INDEX_URL = "${info.url}simple"
 	env.PIP_TRUSTED_HOST = "${hostName} pypi.org"
-	env.PIP_EXTRA_INDEX_URL = 'https://pypi.org/simple'
```
AnthonyLaw, line 67: Was this intentional to ensure Nexus is always the primary source?

jenkins/shared-library/vars/configureArtifactRepository.groovy
```diff
@@ -64,7 +64,8 @@ void configurePypi(Map info) {
...
+	env.PIP_NO_CACHE_DIR = 1
```
Jaguar0625, line 67: this could slow down the builds, force reinstall isn't sufficient on its own?

Wayonb:
close for now

PR #1826, opened 2026-01-16
[mkdocs] fix: Japanese translation

daoka:
I reviewed and fixed the Japanese translation of the textbook.


segfaultxavi (approved):
Thanks for this work!

PR #1828, opened 2026-01-20
[jenkins] fix: update Clang complier to 21

Wayonb:
problem: Catapult is not built with the latest Clang version
solution: Upgrade to the latest Clang in the Jenkins build

### DIFF
--- a/client/catapult/CMakeGlobalSettings.cmake
+++ b/client/catapult/CMakeGlobalSettings.cmake
@@ -160,7 +160,9 @@ elseif("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
 	# - Wno-switch-enum: do not require enum switch statements to list every value (this setting is also incompatible with GCC warnings)
 	# - Wno-weak-vtables: vtables are emitted in all translation units for virtual classes with no out-of-line virtual method definitions
 	# - Wno-unsafe-buffer-usage: allow unsafe buffer usage https://reviews.llvm.org/D137379
-	# = Wno-shadow-uncaptured-local: allow shadowing of local variables in lambdas https://github.com/llvm/llvm-project/issues/81307
+	# - Wno-shadow-uncaptured-local: allow shadowing of local variables in lambdas https://github.com/llvm/llvm-project/issues/81307
+	# - Wnrvo: error: not eliding copy on return
+	# - Wno-thread-safety-negative:
 	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} \
 		-stdlib=libc++ \
 		-Weverything \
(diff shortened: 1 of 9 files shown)


client/catapult/CMakeGlobalSettings.cmake
```diff
@@ -160,7 +160,9 @@ elseif("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
 	# - Wno-switch-enum: do not require enum switch statements to list every value (this setting is also incompatible with GCC warnings)
 	# - Wno-weak-vtables: vtables are emitted in all translation units for virtual classes with no out-of-line virtual method definitions
 	# - Wno-unsafe-buffer-usage: allow unsafe buffer usage https://reviews.llvm.org/D137379
-	# = Wno-shadow-uncaptured-local: allow shadowing of local variables in lambdas https://github.com/llvm/llvm-project/issues/81307
+	# - Wno-shadow-uncaptured-local: allow shadowing of local variables in lambdas https://github.com/llvm/llvm-project/issues/81307
+	# - Wnrvo: error: not eliding copy on return
```
Jaguar0625: Wno-nrvo

client/catapult/CMakeGlobalSettings.cmake
```diff
@@ -278,9 +282,11 @@ function(catapult_set_test_compiler_options)
 	elseif("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
 		# - Wno-global-constructors: required for GTEST test definition macros
 		# - Wno-zero-as-null-pointer-constant: workaround for GTEST NULL/nullptr mismatch https://github.com/google/googletest/issues/1323
+		# - Wno-missing-noreturn: test function FailToCreateSecondFuture does not return a value
```
Jaguar0625: if it is only one function, would disable in code with pragmas vs disabling for whole build

Jaguar0625 (approved):
👍

client/catapult/CMakeGlobalSettings.cmake
```diff
@@ -173,7 +174,14 @@ elseif("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
 		-Wno-weak-vtables \
 		-Wno-unsafe-buffer-usage \
 		-Wno-shadow-uncaptured-local \
-		-Wno-switch-default")
+		-Wno-switch-default \
+		-Wno-thread-safety-negative")
+
+	if("${CMAKE_CXX_COMPILER_VERSION}" MATCHES "^21.")
```
AndreaLanfranchi, line 180: I assume the intention here is to check for versions GE 21.
This instead says "starts with 21" which eventually will fail on versions 22 and beyond.

```suggestion
if (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "21")
    # -Wno-nrvo: suppress "not eliding copy on return"
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-nrvo")
endif()
```

AndreaLanfranchi (changes requested):

client/catapult/CMakeGlobalSettings.cmake
Wayonb: No, I actually wanted to just disable for any Clang 21.x version.

Usually, I don't disable for more than one version unless it's a pattern (i.e., two or more versions have the same issue).  What I have noticed in the past is that some features change depending on how they affect end users on initial release.

Jaguar0625 (approved):
👍

PR #1829, opened 2026-01-21
[jenkins]: fix ninja build dependencies

Wayonb:
problem: tools implicitly depend on the publishing of the Catapult SDK
                  With Ninja 1.12 and later, this causes the tool project to not find the catapult header file.
solution: add explicit dependency

### DIFF
--- a/client/catapult/CMakeGlobalSettings.cmake
+++ b/client/catapult/CMakeGlobalSettings.cmake
@@ -503,6 +503,7 @@ function(catapult_define_tool TOOL_NAME)
 	target_link_libraries(${TARGET_NAME} catapult.tools)
 	catapult_target(${TARGET_NAME})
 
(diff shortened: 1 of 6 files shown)


Jaguar0625:
shouldn't this be based on `fix/catapult/macos_sdk_location` ? 🤔

Wayonb:
> shouldn't this be based on `fix/catapult/macos_sdk_location` ? 🤔

No, this fix is not specific to macOS but applies to later versions of Ninja.

Jaguar0625:
> > shouldn't this be based on `fix/catapult/macos_sdk_location` ? 🤔
> 
> No, this fix is not specific to macOS but applies to later versions of Ninja.

😅 my local branch was on wrong commit, so i was confused 😅

Jaguar0625 (commented):
👍

Jaguar0625 (approved):
👍

PR #1838, opened 2026-01-27
[mkdocs] docs: add link namespace tutorials

zero4862:
Adds tutorials on how to:

- link and unlink namespaces to address
- link and unlink namespaces to mosaics

### DIFF
--- a/mkdocs/overrides/devbook/namespaces/link-namespace-to-address.py
+++ b/mkdocs/overrides/devbook/namespaces/link-namespace-to-address.py
@@ -25,16 +25,16 @@ signer_address = facade.network.public_key_to_address(
 	signer_key_pair.public_key)
 print(f'Signer address: {signer_address}')
 
-namespace_name = f'nsaddr_{int(time.time())}'
+namespace_name = os.getenv('NAMESPACE_NAME', 'my_namespace')
 print(f'Namespace name: {namespace_name}')
 
 namespace_id = generate_namespace_path(namespace_name)[-1]
 print(f'Namespace ID: {namespace_id} ({hex(namespace_id)})')
 
 # Target address to link the namespace to
-target_address_hex = (
-	'TCWYXKVYBMO4NBCUF3AXKJMXCGVSYQOS7ZG2TLI')
-target_address = Address(target_address_hex)
+target_address = Address(
+	os.getenv('TARGET_ADDRESS',
+	'TCWYXKVYBMO4NBCUF3AXKJMXCGVSYQOS7ZG2TLI'))
 print(f'Target address: {target_address}')
 
 try:
@@ -58,7 +58,7 @@ try:
 		fee_mult = max(median_mult, minimum_mult)
 		print(f'  Fee multiplier: {fee_mult}')
 
-	# Build the transaction
+	# Build the alias transaction
 	transaction = facade.transaction_factory.create({
 		'type': 'address_alias_transaction_v1',
 		'signer_public_key': signer_key_pair.public_key,
@@ -117,39 +117,38 @@ try:
 			f'{NODE_URL}{namespace_path}') as response:
 		response_json = json.loads(response.read().decode())
 		namespace_info = response_json['namespace']
-		print('Namespace alias information:')
+		print('Alias information:')
 		alias_type = namespace_info['alias']['type']
 		print(f'  Alias type: {alias_type}')
 		if alias_type == 2:  # ADDRESS type
 			aliased_address_hex = namespace_info['alias']['address']
 			aliased_address = Address(bytes.fromhex(aliased_address_hex))
-			print(f'  Aliased address: {aliased_address}')
+			print(f'  Linked address: {aliased_address}')
 
-	# Use the alias in a transfer transaction
-	print(f'Using namespace alias in transfer: {namespace_name}')
+	# Send a transfer using the alias instead of a raw address
+	print(f'Using alias in transfer: {namespace_name}')
 
-	# Convert namespace to aliased address format (24 bytes)
-	# Byte 0: network byte (0x98) | alias flag (0x01) = 0x99 for mainnet
-	#         or 0x90 | 0x01 = 0x91 for testnet
+	# Encode the namespace ID as a recipient address (24 bytes)
+	# Byte 0: network byte (Network Identifier | 0x01 to indicate alias)
 	# Bytes 1-8: namespace ID in little-endian
 	# Bytes 9-23: zero padding
-	ns_id = generate_namespace_path(namespace_name)[-1]
-	alias_address = Address(
+	recipient_id = generate_namespace_path(namespace_name)[-1]
+	recipient_address = Address(
 		bytes([facade.network.identifier | 0x01]) +
-		ns_id.to_bytes(8, 'little') + bytes(15))
+		recipient_id.to_bytes(8, 'little') + bytes(15))
 
 	transfer_tx = facade.transaction_factory.create({
 		'type': 'transfer_transaction_v1',
 		'signer_public_key': signer_key_pair.public_key,
 		'deadline': timestamp.add_hours(2).timestamp,
-		'recipient_address': alias_address,
+		'recipient_address': recipient_address,
 		'mosaics': [{
 			'mosaic_id': generate_mosaic_alias_id('symbol.xym'),
 			'amount': 1_000_000  # 1 XYM
 		}]
 	})
 	print('Transfer transaction:')
-	print(f'  Recipient address (alias): {alias_address}')
+	print(f'  Recipient address (alias): {recipient_address}')
 
 except Exception as e:
 	print(e)
--- a/mkdocs/overrides/devbook/namespaces/link-namespace-to-address.py
+++ b/mkdocs/overrides/devbook/namespaces/link-namespace-to-address.py
@@ -1,0 +1,155 @@
+import json
+import os
+import time
+import urllib.request
+
+from symbolchain.CryptoTypes import PrivateKey
+from symbolchain.facade.SymbolFacade import SymbolFacade
+from symbolchain.symbol.Network import NetworkTimestamp
+from symbolchain.symbol.IdGenerator import (
+	generate_namespace_path, generate_mosaic_alias_id)
+from symbolchain.sc import Amount
+from symbolchain.symbol.Network import Address
+
+NODE_URL = os.getenv(
+	'NODE_URL', 'https://reference.symboltest.net:3001')
+print(f'Using node {NODE_URL}')
+
+SIGNER_PRIVATE_KEY = os.getenv(
+	'SIGNER_PRIVATE_KEY',
+	'0000000000000000000000000000000000000000000000000000000000000000')
+signer_key_pair = SymbolFacade.KeyPair(PrivateKey(SIGNER_PRIVATE_KEY))
+
+facade = SymbolFacade('testnet')
+signer_address = facade.network.public_key_to_address(
+	signer_key_pair.public_key)
+print(f'Signer address: {signer_address}')
+
+namespace_name = f'nsaddr_{int(time.time())}'
+print(f'Namespace name: {namespace_name}')
+
+namespace_id = generate_namespace_path(namespace_name)[-1]
+print(f'Namespace ID: {namespace_id} ({hex(namespace_id)})')
+
+# Target address to link the namespace to
+target_address_hex = (
+	'TCWYXKVYBMO4NBCUF3AXKJMXCGVSYQOS7ZG2TLI')
+target_address = Address(target_address_hex)
+print(f'Target address: {target_address}')
+
+try:
+	# Fetch current network time
+	time_path = '/node/time'
+	print(f'Fetching current network time from {time_path}')
+	with urllib.request.urlopen(f'{NODE_URL}{time_path}') as response:
+		response_json = json.loads(response.read().decode())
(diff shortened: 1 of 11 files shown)


mkdocs/overrides/devbook/namespaces/link-namespace-to-address.py
```diff
@@ -0,0 +1,155 @@
...
+namespace_name = f'nsaddr_{int(time.time())}'
+print(f'Namespace name: {namespace_name}')
+
+namespace_id = generate_namespace_path(namespace_name)[-1]
+print(f'Namespace ID: {namespace_id} ({hex(namespace_id)})')
+
+# Target address to link the namespace to
+target_address_hex = (
+	'TCWYXKVYBMO4NBCUF3AXKJMXCGVSYQOS7ZG2TLI')
+target_address = Address(target_address_hex)
```
segfaultxavi: Initialize `target_address` straight from the string literal and skip `target_address_hex`, like you do in JS. The name is confusing because that's not hexadecimal.

mkdocs/pages/en/devbook/namespaces/link-namespace-to-address.md
```diff
@@ -0,0 +1,172 @@
...
+### Defining the Namespace and Target Address
+
+{{ tutorial.code_snippet(['py:28:38', 'js:24:36']) }}
+
+The code defines:
+
+* **Namespace name:** The name of the namespace to be linked. A timestamp is appended to ensure uniqueness across runs.
+* **Namespace ID:** The ID is generated from the namespace name using <dy:IdGenerator.generateNamespacePath>,
+    which returns an array of IDs for each level in the hierarchy.
+    The last element (`[-1]`) is selected to get the final namespace ID.
```
segfaultxavi: Remove the parenthesis, since it's Python specific and looks weird when looking at JS code.

Once we understand why we only use the last component and how this relates to subnamespaces, we might need to add an info box explaining it.

mkdocs/pages/en/devbook/namespaces/link-namespace-to-address.md
```diff
@@ -0,0 +1,172 @@
...
+
+The namespace ID is then encoded into a 24-byte address with the following structure:
+
+* **Byte 0:** Network byte with alias flag set (`network_identifier | 0x01`)
+* **Bytes 1-8:** Namespace ID in little-endian byte order
+* **Bytes 9-23:** Zero padding
+
+For more details on how to announce transfer transactions, see the
+[Transfer Transaction](../transactions/transfer.md) tutorial.
+
```
segfaultxavi, line 148: It would also be good to add a note regarding the Address Resolution Receipt that is generated every time an alias is used,
and link to the [Textbook](https://docs.symboltest.net/en/textbook/blocks/#resolution-statements).
And later link to the tutorial about receipts.

mkdocs/overrides/devbook/namespaces/link-namespace-to-address.py
```diff
@@ -0,0 +1,155 @@
...
+	'SIGNER_PRIVATE_KEY',
+	'0000000000000000000000000000000000000000000000000000000000000000')
+signer_key_pair = SymbolFacade.KeyPair(PrivateKey(SIGNER_PRIVATE_KEY))
+
+facade = SymbolFacade('testnet')
+signer_address = facade.network.public_key_to_address(
+	signer_key_pair.public_key)
+print(f'Signer address: {signer_address}')
+
+namespace_name = f'nsaddr_{int(time.time())}'
```
segfaultxavi: Use a fixed namespace, and add a warning note that this must match an existing ns, for example, the one created in the previous tutorial.
If we use time, it will not match anything (I am ashamed to say this happened to me).

mkdocs/overrides/devbook/namespaces/link-namespace-to-address.py
```diff
@@ -0,0 +1,155 @@
...
+print(f'Namespace name: {namespace_name}')
+
+namespace_id = generate_namespace_path(namespace_name)[-1]
```
segfaultxavi, line 31: I don't understand why we take only the last component. How will this work with subnamespaces? I mean, what happens when you have `foo.xym` and `bar.xym` and use only the last part?

mkdocs/overrides/devbook/namespaces/link-namespace-to-address.py
```diff
@@ -0,0 +1,155 @@
...
+			print(f'  Aliased address: {aliased_address}')
+
+	# Use the alias in a transfer transaction
+	print(f'Using namespace alias in transfer: {namespace_name}')
+
+	# Convert namespace to aliased address format (24 bytes)
+	# Byte 0: network byte (0x98) | alias flag (0x01) = 0x99 for mainnet
+	#         or 0x90 | 0x01 = 0x91 for testnet
+	# Bytes 1-8: namespace ID in little-endian
+	# Bytes 9-23: zero padding
```
segfaultxavi:
```suggestion
	# Convert namespace name to aliased address format (24 bytes)
	# Byte 0: network byte (Network Identifier | 0x01 to indicate alias)
	# Bytes 1-8: namespace ID in little-endian
	# Bytes 9-23: zero padding
```

I think it's clearer this way.
It was too much info, and the network ID bytes were not correct
(They are 0x68 for mainnet and 0x98 for testnet).

mkdocs/overrides/devbook/namespaces/link-namespace-to-address.py
```diff
@@ -0,0 +1,155 @@
...
+	ns_id = generate_namespace_path(namespace_name)[-1]
+	alias_address = Address(
+		bytes([facade.network.identifier | 0x01]) +
+		ns_id.to_bytes(8, 'little') + bytes(15))
```
segfaultxavi:
```suggestion
		ns_id.to_bytes(23, 'little'))
```

I think it's simpler this way, unless you wanted to highlight the padding bytes?

mkdocs/pages/en/devbook/namespaces/link-namespace-to-address.md
```diff
@@ -0,0 +1,172 @@
...
+{{ tutorial.code_snippet(['py:128:152', 'js:147:179']) }}
+
+Once the namespace is linked to an address, the namespace can be used in place of the address in transactions.
+The code demonstrates creating a <transfer transaction:> using the namespace as the recipient address instead of
+the full hexadecimal address.
+
+To use a namespace as a recipient address, the namespace ID must be encoded into a 24-byte address. The namespace ID
+is obtained using `generate_namespace_path()[-1]`, as described in the
+[previous section](#defining-the-namespace-and-target-address).
```
segfaultxavi: `generate_namespace_path()[-1]` is Python-specific, better just say "the last component".
(But I really want to understand how subnamespaces work first)

And also, line break after period :)

mkdocs/pages/en/devbook/namespaces/link-namespace-to-address.md
```diff
@@ -0,0 +1,172 @@
...
+specified address.
+
+### Using the Alias
+
+{{ tutorial.code_snippet(['py:128:152', 'js:147:179']) }}
+
+Once the namespace is linked to an address, the namespace can be used in place of the address in transactions.
+The code demonstrates creating a <transfer transaction:> using the namespace as the recipient address instead of
+the full hexadecimal address.
+
```
segfaultxavi, line 132:
```suggestion
For simplicity, this example creates the transaction but does not announce it or wait for its confirmation.

```

mkdocs/overrides/devbook/namespaces/link-namespace-to-address.py
```diff
@@ -0,0 +1,155 @@
...
+			f'{NODE_URL}{namespace_path}') as response:
+		response_json = json.loads(response.read().decode())
+		namespace_info = response_json['namespace']
+		print('Namespace alias information:')
+		alias_type = namespace_info['alias']['type']
+		print(f'  Alias type: {alias_type}')
+		if alias_type == 2:  # ADDRESS type
+			aliased_address_hex = namespace_info['alias']['address']
+			aliased_address = Address(bytes.fromhex(aliased_address_hex))
+			print(f'  Aliased address: {aliased_address}')
```
segfaultxavi: I would print "Target address" so it matches what we said above.
Otherwise it looks like this is the namespace address, which is confusing because it does not match the line below (Recipient address).

mkdocs/overrides/devbook/namespaces/link-namespace-to-address.py
```diff
@@ -0,0 +1,155 @@
...
+
+	# Use the alias in a transfer transaction
```
segfaultxavi: I understand you obtain again the ns_id to show the whole process an app would follow, from namespace name to address.
But this is a bit confusing, and the variable names do not help (`alias_address` vs `aliased_address`, and `namespace_id` vs `ns_id`).

Maybe indicate clearly in the comment what is about to happen ("Show the whole process to send a tx to an alias, starting from the alias name"),
and use variables called `recipient_name`, `recipient_id` and `recipient_address` ?

segfaultxavi (changes requested):
In addition to the inline comments, add links to these two new tutorials in the other Namespace tutorials.

mkdocs/overrides/devbook/namespaces/link-namespace-to-address.py
```diff
@@ -0,0 +1,155 @@
...
+	fee_path = '/network/fees/transaction'
+	print(f'Fetching recommended fees from {fee_path}')
+	with urllib.request.urlopen(f'{NODE_URL}{fee_path}') as response:
+		response_json = json.loads(response.read().decode())
+		median_mult = response_json['medianFeeMultiplier']
+		minimum_mult = response_json['minFeeMultiplier']
+		fee_mult = max(median_mult, minimum_mult)
+		print(f'  Fee multiplier: {fee_mult}')
+
+	# Build the transaction
```
segfaultxavi:
```suggestion
	# Build the link transaction
```

Better to be explicit.

mkdocs/pages/en/devbook/namespaces/link-namespace-to-mosaic.md
```diff
@@ -0,0 +1,166 @@
...
+
+{{ tutorial.code_snippet(['py:124:144', 'js:143:162']) }}
+
+Once the namespace is linked to a mosaic, the namespace can be used in place of the mosaic ID in transactions.
+The code demonstrates creating a <transfer transaction:> using the namespace in the mosaics array instead of
+the full hexadecimal mosaic ID.
+
+To use a namespace as a mosaic ID, the namespace name is converted to its mosaic alias ID using
+<dy:IdGenerator.generateMosaicAliasId>.
+
```
segfaultxavi, line 138: Add a note saying that this example sends the mosaic back to the sender, for simplicity.
Otherwise some readers might wonder what is going on with the `recipient_address`.

mkdocs/pages/en/devbook/namespaces/link-namespace-to-address.md
```diff
@@ -0,0 +1,172 @@
...
+To use a namespace as a recipient address, the namespace ID must be encoded into a 24-byte address. The namespace ID
+is obtained using `generate_namespace_path()[-1]`, as described in the
+[previous section](#defining-the-namespace-and-target-address).
+
+The namespace ID is then encoded into a 24-byte address with the following structure:
```
segfaultxavi: We better explain that "at this point the SDK lacks a method to perform this low-level operation so it needs to be done manually."

segfaultxavi (changes requested):
Almost all comments to the Address tutorial apply to the Mosaic tutorial too.

mkdocs/overrides/devbook/namespaces/link-namespace-to-address.py
zero4862: Subnamespace IDs are derived hierarchically from the root namespace, so `xym` from `foo.xym` and `bar.xym` should lead to different results.

We use `generate_namespace_path()[-1]` instead of `generate_namespace_id()` because the latter does not work well with subnamespaces (more details in `docsà channel).

For example:

* `generate_namespace_id('symbol.xym')`: https://testnet.symbol.fyi/namespaces/F7DAAA67B0536F75 ✗
* `generate_namespace_path('symbol.xym')[-1]`: https://testnet.symbol.fyi/namespaces/E74B99BA41F4AFEE ✓

segfaultxavi: Ah, OK, gotcha.
So `generate_namespace_path('foo.xym')[-1]` and `generate_namespace_path('bar.xym')[-1]` will produce different IDs.

Still worth mentioning this in the tutorials, probably in a "Subnamespaces" info box.

mkdocs/overrides/devbook/namespaces/link-namespace-to-address.py
zero4862: smart, but I feel the current version makes it easier to map with the description:

```
# Bytes 1-8: namespace ID in little-endian
# Bytes 9-23: zero padding
```

segfaultxavi: Fair enough!

segfaultxavi (approved):
All good, except a couple internal formatting issues that I've fixed in an additional commit:

- List should use a consistent character, either - or *
- Line breaks after periods.

PR #1839, opened 2026-01-30
[jenkins] fix: publish target is not needed on Windows

Wayonb:
problem: Windows build is failing to build the target publish
solution: remove as this is not needed anymore.


Jaguar0625 (approved):
👍

PR #1845, opened 2026-02-05
[jenkins] fix: add missing Python packages

Wayonb:
problem: colorama was not found during linting because it was not installed in the virtual environment.
solution: install colorama after the Python virtual environment is created.


jenkins/catapult/templates/FedoraTestBaseImage.Dockerfile
```diff
@@ -31,3 +30,5 @@ WORKDIR ${HOME_DIR}
 ENV VIRTUAL_ENV=${HOME_DIR}/venv
 RUN python3 -m venv $VIRTUAL_ENV
 ENV PATH="$VIRTUAL_ENV/bin:$PATH"
+
+RUN pip3 install -U colorama conan cryptography gitpython pycodestyle pylint ply PyYAML
```
Jaguar0625, line 34: i assume you intentionally added ply?

Jaguar0625 (approved):
👍

jenkins/catapult/templates/FedoraTestBaseImage.Dockerfile
Wayonb: Yeah, the ply package is required also.

PR #1846, opened 2026-02-05
[client/catapult] fix: disable stringop-overread for GCC 14

Wayonb:
problem: GCC 14 FP on stringop-overread
solution: disable stringop-overread for GCC 14

### DIFF
--- a/client/catapult/tests/catapult/ionet/PacketEntityUtilsTests.cpp
+++ b/client/catapult/tests/catapult/ionet/PacketEntityUtilsTests.cpp
@@ -162,7 +162,7 @@ namespace catapult { namespace ionet {
 
 // disable this warning which happens only on GCC 14
 // error: reading 1 or more bytes from a Region of size 0 [-Werror=stringop-overread]
-#if 14 == __GNUC__ && !defined(__clang__)
+#if 14 <= __GNUC__ && !defined(__clang__)
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wstringop-overread"
 #endif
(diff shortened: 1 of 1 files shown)


client/catapult/tests/catapult/ionet/PacketEntityUtilsTests.cpp
```diff
@@ -174,6 +181,10 @@ namespace catapult { namespace ionet {
 			EXPECT_TRUE(TTraits::IsEmpty(extractResult)) << "packet size " << size;
 		}
 
+#if 14 == __GNUC__ && !defined(__clang__)
```
Jaguar0625: should be same condition as in lnie 165

client/catapult/tests/catapult/ionet/PacketEntityUtilsTests.cpp
```diff
@@ -160,6 +160,13 @@ namespace catapult { namespace ionet {
 			}
 		};
 
+// disable this warning which happens only on GCC 14
+// error: reading 1 or more bytes from a Region of size 0 [-Werror=stringop-overread]
+#if 14 <= __GNUC__ && !defined(__clang__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wstringop-overread"
```
Jaguar0625, line 167: what is warning output? 🤔

client/catapult/tests/catapult/ionet/PacketEntityUtilsTests.cpp
Wayonb: yea, good catch

client/catapult/tests/catapult/ionet/PacketEntityUtilsTests.cpp
Wayonb: The error is above, but I had to remove parts of it since it triggers linting.

Jaguar0625 (approved):
👍

PR #1848, opened 2026-02-06
[client/catapult] fix: disable globals variablie san for macOS

Wayonb:
problem: for Clang 16 and 17 on macOS, AddressSanitizer failed due to false positive detection of ODR and global buffer overflow
solution: disable the global variable in AddressSanitizer on macOS
https://maskray.me/blog/2023-10-15-address-sanitizer-global-variable-instrumentation


Jaguar0625 (approved):
👍

PR #1858, opened 2026-02-13
[mkdocs] docs: add modify mosaic definition tutorial

zero4862:
Adds modify mosaic definition tutorial based on https://github.com/symbol/symbol/blob/main/docs/index.md#other


mkdocs/pages/en/devbook/mosaics/modify-mosaic-definition.md
```diff
@@ -0,0 +1,164 @@
...
+This account must be the original creator of the mosaic.
+
+### Fetching Network Time and Fees
+
+{{ tutorial.code_snippet(['py:27:47', 'js:24:42']) }}
+
+Network time and recommended fees are fetched from <get:/node/time> and <get:/network/fees/transaction> respectively,
+following the process described in the [Transfer Transaction](../transactions/transfer.md) tutorial.
+
+### Building the Modification Transaction
```
segfaultxavi, line 66: Given how the new values are XORed with the current ones, I believe it would be a more realistic use case to first fetch the current values and then modify them.
You can just move the "Retrieving the Mosaic" code before the "Building the Modification Tx".

mkdocs/pages/en/devbook/mosaics/modify-mosaic-definition.md
```diff
@@ -0,0 +1,164 @@
+---
+title: Modify Mosaic Definition
+---
+
+# Modifying a Mosaic Definition
+
+After creating a <mosaic:>, its definition can be modified by sending a second definition transaction with the same
+nonce.
```
segfaultxavi: Can't use the word "nonce" before defining it. Maybe just say "identifier" here and later explain that the Mosaic ID is built from the signer's address and this thing called nonce.

Also, it would be helpful to give some examples of the kind of things that can be modified (duration, divisibility and flags).
The most obvious modification is the supply, but that's not what this tutorial is about. It's worth explaining this and point to the supply change tutorial.

mkdocs/pages/en/devbook/mosaics/modify-mosaic-definition.md
```diff
@@ -0,0 +1,164 @@
+---
+title: Modify Mosaic Definition
+---
+
+# Modifying a Mosaic Definition
```
segfaultxavi, line 5: The whole concept of modifying a mosaic definition after it has been created is pretty shocking.
The first thing that comes to mind is that it can be abused, so you need to explain right after the first paragraph that this operation is only possible when the supply is zero, so there are no circulating mosaics.

And then, the obvious following question is: if the supply is zero, why should I modify an existing mosaic instead of creating a new one? We need to answer that too.

mkdocs/pages/en/devbook/mosaics/modify-mosaic-definition.md
```diff
@@ -0,0 +1,164 @@
...
+* Have a mosaic with **zero supply**.
+    Either [create a new mosaic](./create-mosaic.md) without minting supply,
+    or [decrease the supply](./change-mosaic-supply.md) of an existing mosaic to `0`.
+* Obtain <XYM:> to pay for the transaction and lease fees.
+    See [Getting Testnet Funds from the Faucet](../accounts/testnet-faucet.md).
+
+Additionally, review the [Transfer transaction](../transactions/transfer.md) tutorial to understand how
+transactions are announced and confirmed.
+
+!!! warning "Zero supply requirement"
```
segfaultxavi: Yeah, this needs to be part of the intro, right at the top.

The fact that to decrease the supply the owner must own it is a distracting detail. I would put it in parenthesis or remove it completely. The important part is that there cannot be a single mosaic out in the world before changing its definition.

mkdocs/pages/en/devbook/mosaics/modify-mosaic-definition.md
```diff
@@ -0,0 +1,164 @@
...
+{{ tutorial.code_snippet(['py:27:47', 'js:24:42']) }}
+
+Network time and recommended fees are fetched from <get:/node/time> and <get:/network/fees/transaction> respectively,
+following the process described in the [Transfer Transaction](../transactions/transfer.md) tutorial.
+
+### Building the Modification Transaction
+
+{{ tutorial.code_snippet(['py:49:65', 'js:44:61']) }}
+
+The `MOSAIC_NONCE` environment variable specifies the nonce of the mosaic to modify.
```
segfaultxavi: I think this is the right place to quickly define what a nonce is, and add a proper glossary definition in the Mosaic textbook page, in the Mosaic ID section.

mkdocs/pages/en/devbook/mosaics/modify-mosaic-definition.md
```diff
@@ -0,0 +1,164 @@
...
+
+When processing the transaction, each property is combined with the mosaic's current value using the following rules:
+
+* **Flags** are XOR'd with the current flags.
+    Setting a flag that is already active removes it; setting a flag that is not active adds it.
+    For a description of each available flag, see
+    [Building the Mosaic Definition Transaction](./create-mosaic.md#building-the-mosaic-definition-transaction).
+* **Divisibility** is XOR'd with the current divisibility.
+    The resulting value must be between `0` and `6`.
+* **Duration** is added to the current remaining duration.
```
segfaultxavi, line 92: Huh. But the Mosaics textbook page says that duration cannot be changed!
This needs to be clarified. Does that warning in the textbook mean that the duration cannot be changed _if the supply is not zero_?

mkdocs/pages/en/devbook/mosaics/modify-mosaic-definition.md
```diff
@@ -0,0 +1,164 @@
...
+    A value of `0` leaves the duration unchanged.
```
segfaultxavi, line 94: What about negative duration? Can the duration be reduced?
(I don't think so because the `duration` field is unsigned, but we should check.)

mkdocs/pages/en/devbook/mosaics/modify-mosaic-definition.md
```diff
@@ -0,0 +1,164 @@
...
+    Eternal mosaics (duration `0`) cannot have their duration modified.
+    The resulting duration cannot exceed approximately 10,512,000 blocks (approximately 10 years).
+
+In this example, the existing mosaic has flags `transferable restrictable`
+([numeric value](./create-mosaic.md#conclusion) `6`).
+The modification sets `flags: 'revokable'` (numeric value `8`).
+XOR produces `6 ^ 8 = 14`, which corresponds to `transferable restrictable revokable`.
```
segfaultxavi: Using `^` for XOR in text is a bit confusing, even if that's the right operator in Python and JS code.
You used it to indicate a power in another tutorial :)
Maybe it's better to use ⊕ in the text?

mkdocs/pages/en/devbook/mosaics/modify-mosaic-definition.md
```diff
@@ -0,0 +1,164 @@
...
+| Flag            | Current | Modification | Result (XOR) |
+| --------------- | ------- | ------------ | ------------ |
+| `transferable`  | on      | off          | on           |
+| `restrictable`  | on      | off          | on           |
+| `revokable`     | off     | on           | on           |
+
+Setting `divisibility` to `0` results in `2 ^ 0 = 2` (no change), and `duration: 0` adds nothing to the current
+duration.
+
+!!! note "Lease fee"
```
segfaultxavi, line 114: So every time you modify a mosaic you pay the fee? In other words, the Mosaic Definition Transaction always has a cost of 50 XYM? It would be good to clarify.

segfaultxavi (changes requested):
I request some structural changes, so I will do a deeper review after those are made.

mkdocs/pages/en/devbook/mosaics/modify-mosaic-definition.md
zero4862: We can't retrieve a mosaic and guarantee reproducible results if the script modifies it, unless it also includes the creation step which I think it will be redundant with the creation tutorial.

Instead, I'm adding some more details explaining that retrieving the current definition should be the first step before modifying (via API or explorer), with a link to the creation tutorial where the values are set.

mkdocs/pages/en/devbook/mosaics/modify-mosaic-definition.md
zero4862: It links to the creation tutorial where it was originally defined. However, I agree it also makes sense to include the entry in the textbook.

mkdocs/pages/en/devbook/mosaics/modify-mosaic-definition.md
zero4862: Updated the intro

mkdocs/pages/en/devbook/mosaics/modify-mosaic-definition.md
zero4862: Updated textbook

mkdocs/pages/en/devbook/mosaics/modify-mosaic-definition.md
segfaultxavi: OK. I've just changed it slightly to clarify that it's about total supply being zero, not that the supply has not been distributed yet.
Because the creator could still recall all mosaics, destroy them, and then change the definition.

segfaultxavi (approved):
The confirmation loop in the JS code is not the same as in other tutorials, it's longer and looks more convoluted.
But anyway, it's not critical.

PR #1862, opened 2026-02-17
[client/rest]: fix supply endpoints to accept text/plain

cryptoBeliever:

- Endpoints /network/currency/supply/circulating, /network/currency/supply/max, /network/currency/supply/total return content-type: text/plain.
- Requests with Accept: text/plain receive 406 with "Server accepts: application/json".
- Requests with Accept: application/json succeed and return plain text, which is inconsistent.

## What's the issue?

- The API returns plain text, but the server only accepts Accept: application/json.
- Clients that send Accept: `text/plain` get 406.
- The response format and Accept handling are inconsistent.

- Added a path-specific accept parser in bootstrapper.js:
  - Supply endpoints accept only Accept: text/plain.
  - Other endpoints accept only Accept: application/json.

**Breaking change**: Supply endpoints no longer accept Accept: application/json; clients must send Accept: text/plain.

### DIFF
--- a/client/rest/src/server/bootstrapper.js
+++ b/client/rest/src/server/bootstrapper.js
@@ -99,9 +99,11 @@ const catapultRestifyPlugins = {
 
 		if (req.accepts(accepts))
 			return next();
+
 		const error = isTextPlainPath
 			? new restifyErrors.NotAcceptableError('Endpoint accepts only text/plain')
 			: new restifyErrors.NotAcceptableError('Endpoint accepts only application/json');
+
 		return next(error);
 	}
 };
--- a/client/rest/src/server/bootstrapper.js
+++ b/client/rest/src/server/bootstrapper.js
@@ -62,7 +62,7 @@ const createCrossDomainHeaderAdder = crossDomainConfig => (req, res) => {
 	}
 };
 
-const TEXT_PLAIN_SUPPLY_PATHS = [
+const TEXT_PLAIN_PATHS = [
 	'/network/currency/supply/circulating',
 	'/network/currency/supply/max',
 	'/network/currency/supply/total'
@@ -94,14 +94,14 @@ const catapultRestifyPlugins = {
 		next();
 	},
 	acceptParser: () => (req, res, next) => {
-		const isSupplyPath = TEXT_PLAIN_SUPPLY_PATHS.includes(req.path());
-		const accepts = isSupplyPath ? ['text/plain'] : ['application/json'];
-		const error = isSupplyPath
-			? new restifyErrors.NotAcceptableError('Endpoint accepts only text/plain')
-			: new restifyErrors.NotAcceptableError('Endpoint accepts only application/json');
+		const isTextPlainPath = TEXT_PLAIN_PATHS.includes(req.path());
+		const accepts = isTextPlainPath ? ['text/plain'] : ['application/json'];
 
 		if (req.accepts(accepts))
 			return next();
+		const error = isTextPlainPath
+			? new restifyErrors.NotAcceptableError('Endpoint accepts only text/plain')
+			: new restifyErrors.NotAcceptableError('Endpoint accepts only application/json');
 		return next(error);
 	}
 };
--- a/client/rest/src/server/bootstrapper.js
+++ b/client/rest/src/server/bootstrapper.js
@@ -62,6 +62,12 @@ const createCrossDomainHeaderAdder = crossDomainConfig => (req, res) => {
 	}
 };
 
+const TEXT_PLAIN_SUPPLY_PATHS = [
+	'/network/currency/supply/circulating',
+	'/network/currency/supply/max',
+	'/network/currency/supply/total'
+];
+
 const catapultRestifyPlugins = {
 	crossDomain: addCrossDomainHeaders => (req, res, next) => {
 		addCrossDomainHeaders(req, res);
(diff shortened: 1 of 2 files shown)


client/rest/src/server/bootstrapper.js
```diff
@@ -62,6 +62,12 @@ const createCrossDomainHeaderAdder = crossDomainConfig => (req, res) => {
 	}
 };
 
+const TEXT_PLAIN_SUPPLY_PATHS = [
```
Jaguar0625: TEXT_PLAIN_PATHS or TEXT_PLAIN_ROUTES

the code you're adding is more general for handling correctly application/json and text/plain routes. currently, only supply routes are text/plain, but it would be better to keep naming more general in case other (non-supply) text/plain routes are added.

client/rest/src/server/bootstrapper.js
```diff
@@ -86,6 +92,17 @@ const catapultRestifyPlugins = {
 		}
 
 		next();
+	},
+	acceptParser: () => (req, res, next) => {
+		const isSupplyPath = TEXT_PLAIN_SUPPLY_PATHS.includes(req.path());
```
Jaguar0625: similarly, isTextPlainPath

client/rest/src/server/bootstrapper.js
```diff
@@ -86,6 +92,17 @@ const catapultRestifyPlugins = {
...
+		const accepts = isSupplyPath ? ['text/plain'] : ['application/json'];
+		const error = isSupplyPath
```
Jaguar0625: i think you can move this down after if?

client/rest/test/server/bootstrapper_spec.js
```diff
@@ -423,8 +430,25 @@ describe('server (bootstrapper)', () => {
...
+					expect(body).to.deep.equal({ code: 'NotAcceptable', message: 'Endpoint accepts only application/json' });
+				}));
+
+			it('accepts text/plain for supply endpoints', () => makeWrappedJsonRequest('/network/currency/supply/circulating', 'get')
+				.header('Accept', 'text/plain')
+				.expectStatus(200)
+				.end((headers, body) => {
+					// Assert:
+					expect(headers['content-type']).to.include('text/plain');
+					expect(String(body)).to.equal('12345.678');
```
Jaguar0625: is String needed? what is body?

client/rest/test/server/bootstrapper_spec.js
```diff
@@ -423,8 +430,25 @@ describe('server (bootstrapper)', () => {
...
+				}));
+
+			it('rejects application/json for supply endpoints', () => makeWrappedJsonRequest('/network/currency/supply/circulating', 'get')
+				.header('Accept', 'application/json')
+				.expectStatus(406)
+				.end((headers, body) => {
+					// Assert:
+					expect(body).to.deep.equal({ code: 'NotAcceptable', message: 'Endpoint accepts only text/plain' });
```
Jaguar0625: add call to assertPayloadHeaders?

client/rest/test/server/bootstrapper_spec.js
```diff
@@ -423,8 +430,25 @@ describe('server (bootstrapper)', () => {
...
+			it('accepts text/plain for supply endpoints', () => makeWrappedJsonRequest('/network/currency/supply/circulating', 'get')
+				.header('Accept', 'text/plain')
+				.expectStatus(200)
+				.end((headers, body) => {
+					// Assert:
+					expect(headers['content-type']).to.include('text/plain');
+					expect(String(body)).to.equal('12345.678');
+				}));
+
+			it('rejects application/json for supply endpoints', () => makeWrappedJsonRequest('/network/currency/supply/circulating', 'get')
```
Jaguar0625: would generalize test names too:

rejects application/json for text/plain endpoint
rejects text/plain for application/json endpoint (above)

and add a third one with some other accept type (that should be rejected), which is default behavior (originally tested by rejects request with invalid accept header)

Jaguar0625 (commented):
👍

client/rest/test/server/bootstrapper_spec.js
cryptoBeliever: It returns a number, so casting to a String is required (or comparing to a number).

client/rest/src/server/bootstrapper.js
cryptoBeliever: Make sense, changed and applied other comments :+1:

client/rest/test/server/bootstrapper_spec.js
Jaguar0625: i'd probably just compare to number unless good reason not to

Jaguar0625 (approved):
👍

client/rest/src/server/bootstrapper.js
```diff
@@ -86,6 +92,17 @@ const catapultRestifyPlugins = {
 		}
 
 		next();
+	},
+	acceptParser: () => (req, res, next) => {
+		const isTextPlainPath = TEXT_PLAIN_PATHS.includes(req.path());
+		const accepts = isTextPlainPath ? ['text/plain'] : ['application/json'];
+
+		if (req.accepts(accepts))
+			return next();
```
Wayonb, line 101: Usually, there is an empty line after an if statement...so after the ``return``

Wayonb (approved):
👍🏾

Jaguar0625 (approved):
👍

PR #1864, opened 2026-02-19
[client/catapult]: migrate boost asio usage to use non-deprecated APIs

Wayonb:
problem: catapult is calling deprecated Boost functions
solution: disable deprecated functions by adding -DBOOST_ASIO_NO_DEPRECATED flag
                 update Boost to 1.90.0

### DIFF
--- a/client/catapult/CMakeGlobalSettings.cmake
+++ b/client/catapult/CMakeGlobalSettings.cmake
@@ -33,6 +33,7 @@ endif()
 ### set boost settings
 add_definitions(-DBOOST_ALL_DYN_LINK)
 add_definitions(-DBOOST_ASIO_USE_TS_EXECUTOR_AS_DEFAULT)
+add_definitions(-DBOOST_ASIO_NO_DEPRECATED)
 
 if(Boost_VERSION VERSION_LESS 1.84)
 	# workaround for https://github.com/boostorg/phoenix/issues/111
--- a/client/catapult/CMakeGlobalSettings.cmake
+++ b/client/catapult/CMakeGlobalSettings.cmake
@@ -35,11 +35,6 @@ add_definitions(-DBOOST_ALL_DYN_LINK)
 add_definitions(-DBOOST_ASIO_USE_TS_EXECUTOR_AS_DEFAULT)
 add_definitions(-DBOOST_ASIO_NO_DEPRECATED)
 
-if(Boost_VERSION VERSION_LESS 1.84)
-	# workaround for https://github.com/boostorg/phoenix/issues/111
-	add_definitions(-DBOOST_PHOENIX_STL_TUPLE_H_)
-endif()
-
 set(Boost_USE_STATIC_LIBS OFF)
 set(Boost_USE_MULTITHREADED ON)
 set(Boost_USE_STATIC_RUNTIME OFF)
--- a/client/catapult/CMakeGlobalSettings.cmake
+++ b/client/catapult/CMakeGlobalSettings.cmake
@@ -125,7 +125,7 @@ if(MSVC)
 	add_compile_options(/MP)            # Enable parallel compilation
 	add_compile_options(/GA)            # Optimizes for Windows applications
 
-	add_definitions(-D_WIN32_WINNT=0x0601)
+	add_definitions(-D_WIN32_WINNT=0x0A00)
 
 	add_compile_options(/w44287)		# 'operator' : unsigned/negative constant mismatch
 	add_compile_options(/w44388)		# 'token' : signed/unsigned mismatch
--- a/client/catapult/conanfile.py
+++ b/client/catapult/conanfile.py
@@ -54,8 +54,8 @@ class CatapultConan(ConanFile):
 		self.options["boost*"].without_math = False
 		self.options["boost*"].without_mpi = True
 		self.options["boost*"].without_nowide = True
-		self.options["boost*"].without_program_options = False
 		self.options["boost*"].without_process = True
+		self.options["boost*"].without_program_options = False
 		self.options["boost*"].without_python = True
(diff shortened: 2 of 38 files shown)


client/catapult/extensions/mongo/tests/mappers/MapperUtilsTests.cpp
```diff
@@ -35,7 +35,7 @@ namespace catapult { namespace mongo { namespace mappers {
 
 	TEST(TEST_CLASS, CanConvertRawPointerToBinary) {
 		// Arrange:
-		uint8_t data[10];
+		uint8_t data[10] = {};
```
Jaguar0625, line 38: change is fine, but is this related to boost somehow?

client/catapult/src/catapult/thread/StrandOwnerLifetimeExtender.h
```diff
@@ -21,28 +21,85 @@
 
 #pragma once
 #include "catapult/utils/WrappedWithOwnerDecorator.h"
+#include "catapult/exceptions.h"
 #include <boost/asio.hpp>
 
 namespace catapult { namespace thread {
 
+	using Strand = boost::asio::strand<boost::asio::io_context::executor_type>;
```
Jaguar0625: can you move into class to avoid thread::Strand symbol?

client/catapult/conanfile.py
```diff
@@ -38,7 +38,7 @@ def configure(self):
 		self.options["boost*"].without_atomic = False
 		self.options["boost*"].without_chrono = False
 		self.options["boost*"].without_container = False
-		self.options["boost*"].without_context = True
+		self.options["boost*"].without_context = False
```
Jaguar0625: is this new asio dependency?

client/catapult/CMakeGlobalSettings.cmake
```diff
@@ -33,11 +33,7 @@ endif()
 ### set boost settings
 add_definitions(-DBOOST_ALL_DYN_LINK)
 add_definitions(-DBOOST_ASIO_USE_TS_EXECUTOR_AS_DEFAULT)
-
-if(Boost_VERSION VERSION_LESS 1.84)
-	# workaround for https://github.com/boostorg/phoenix/issues/111
-	add_definitions(-DBOOST_PHOENIX_STL_TUPLE_H_)
-endif()
+add_definitions(-DBOOST_ASIO_NO_DEPRECATED)
```
Jaguar0625, line 36: i assume we still need this?

Jaguar0625 (commented):
👍

client/catapult/extensions/mongo/tests/mappers/MapperUtilsTests.cpp
Wayonb: No, this and some other variable initialization were needed to build a debug version.

client/catapult/conanfile.py
Wayonb: conan seems to be doing an extra check.  I thought there was an issue but let me check.

```
ERROR: There are invalid packages:
boost/1.90.0: Invalid: process requires ['context', 'filesystem']: context is disabled  
```

client/catapult/CMakeGlobalSettings.cmake
Wayonb: Good question, maybe not, but it might be good to have when we upgrade each release 🤔

client/catapult/src/catapult/thread/StrandOwnerLifetimeExtender.h
Wayonb: I think that I defined this in the thread namespace because it is used in TimedCallBack.h and scheduler.cpp.  So I am assuming I should use the actual type instead of the alias in these files?

client/catapult/CMakeGlobalSettings.cmake
Wayonb: I would keep this to prevent regression.

```
BOOST_ASIO_NO_DEPRECATED

Disables Boost.Asio's deprecated interfaces and functionality.

See [Networking TS Compatibility](https://www.boost.org/doc/libs/latest/doc/html/boost_asio/net_ts.html) for a list of older interfaces that have been deprecated, and their replacements.
```

Wayonb: Also of interest is the old vs new interfaces - https://www.boost.org/doc/libs/latest/doc/html/boost_asio/net_ts.html

client/catapult/conanfile.py
Wayonb: I guess the Boost Process V2 requires context.  I don't see any use of the Process component in the catapult though? 🤔

client/catapult/src/catapult/thread/StrandOwnerLifetimeExtender.h
Jaguar0625: in that case, you can make it public in class;  so you could use something like:

`StrandOwnerLifetimeExtender<StrandedTimedCallback>::Strand`

client/catapult/conanfile.py
Jaguar0625: if you remove process is there compiler error?

Jaguar0625 (approved):
👍

AndreaLanfranchi (approved):
👍

client/catapult/conanfile.py
AndreaLanfranchi: I'm not sure how conan works but it should be able to build a tree of boost requirements on behalf of Boost's recipe

Wayonb: I removed the process from the boost list, plus updated the local build.

Jaguar0625 (approved):
👍

PR #1865, opened 2026-02-20
[mkdocs] docs: add resolve namespace from receipt tutorial

zero4862:
Adds retrieving namespace info from receipts tutorial.

### DIFF
--- a/mkdocs/overrides/devbook/namespaces/resolve-namespace-from-receipt.py
+++ b/mkdocs/overrides/devbook/namespaces/resolve-namespace-from-receipt.py
@@ -1,0 +1,109 @@
+import json
+import os
+import urllib.request
+
+from symbolchain.symbol.Network import Address
+
+NODE_URL = os.getenv(
+	'NODE_URL', 'https://reference.symboltest.net:3001')
+print(f'Using node {NODE_URL}')
+
+# Hash of a confirmed tx that used a namespace alias
+TX_HASH = os.getenv('TRANSACTION_HASH',
+	'BA0C65DB752A3BF1B25285540642537ECE8C2CA716577EDF8BF0F8597A85ADC4')
+print(f'Transaction hash: {TX_HASH}')
+
+try:
+	# Retrieve the confirmed transaction
+	tx_path = f'/transactions/confirmed/{TX_HASH}'
+	print(f'Fetching transaction from {tx_path}')
+	with urllib.request.urlopen(
+			f'{NODE_URL}{tx_path}') as response:
+		tx_data = json.loads(response.read().decode())
+
+	block_height = tx_data['meta']['height']
+	print(f'  Block height: {block_height}')
+
+	# primaryId is 1-based, meta.index is 0-based
+	tx_index = int(tx_data['meta']['index'])
+	tx_primary = tx_index + 1
+	print(f'  Transaction index: {tx_index} (primaryId: {tx_primary})')
+
+	recipient_hex = tx_data['transaction']['recipientAddress']
+	recipient_bytes = bytes.fromhex(recipient_hex)
+	is_address_alias = (recipient_bytes[0] & 0x01) == 1
+	print(f'  Recipient: {recipient_hex}')
+	print(f'  Is address alias: {is_address_alias}')
+
+	aliased_mosaics = set()
+	mosaics = tx_data['transaction']['mosaics']
+	for mosaic in mosaics:
+		mosaic_id = int(mosaic['id'], 16)
+		is_mosaic_alias = (mosaic_id >> 63) & 1 == 1
+		if is_mosaic_alias:
+			aliased_mosaics.add(mosaic['id'])
+		print(f'  Mosaic: {mosaic["id"]}')
+		print(f'  Is mosaic alias: {is_mosaic_alias}')
+
+	# Query address resolution statements
+	if is_address_alias:
+		address_path = ('/statements/resolutions/address'
+			f'?height={block_height}')
+		print(f'\nFetching address resolutions from'
+			f' {address_path}')
+		with urllib.request.urlopen(
+				f'{NODE_URL}{address_path}') as response:
+			address_data = json.loads(
+				response.read().decode())
+
+		address_statements = address_data['data']
+		print(f'  Found {len(address_statements)}'
+			' resolution statement(s)')
+
+		for item in address_statements:
+			statement = item['statement']
+			if statement['unresolved'] != recipient_hex:
+				continue
+			resolved = None
+			for entry in statement['resolutionEntries']:
+				source = entry['source']
+				if source['primaryId'] <= tx_primary:
+					resolved = entry['resolved']
+			if resolved:
+				address = Address(
+					bytes.fromhex(resolved))
+				print('\nAddress resolution:')
+				print(f'  Unresolved:  {statement["unresolved"]}')
+				print(f'  Resolved:   {address}')
+
+	# Query mosaic resolution statements
+	if len(aliased_mosaics):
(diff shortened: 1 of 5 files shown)


mkdocs/pages/en/devbook/namespaces/resolve-namespace-from-receipt.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Code Explanation
+
+### Defining the Transaction Hash
+
+{{ tutorial.code_snippet(['py:11:14', 'js:7:10']) }}
+
+The code defines the hash of a confirmed transaction that used a namespace alias instead of a plain address.
+This hash is read from the `TRANSACTION_HASH` environment variable, which defaults to a known transaction on the
+Symbol <testnet:>.
```
segfaultxavi, line 40: As the task description in the Scope of work document says, the tutorial will stop working if the testnet is ever reset (which has already happened in the past).
So we better use `mainnet` instead.

mkdocs/overrides/devbook/namespaces/resolve-namespace-from-receipt.py
```diff
@@ -0,0 +1,109 @@
...
+NODE_URL = os.getenv(
+	'NODE_URL', 'https://reference.symboltest.net:3001')
+print(f'Using node {NODE_URL}')
+
+# Hash of a confirmed tx that used a namespace alias
+TX_HASH = os.getenv('TRANSACTION_HASH',
+	'BA0C65DB752A3BF1B25285540642537ECE8C2CA716577EDF8BF0F8597A85ADC4')
+print(f'Transaction hash: {TX_HASH}')
+
+try:
```
segfaultxavi, line 16: The whole file is formatted too narrow, and therefore it is needlessly long :)
I have my ruler in VSCode set to 75 chars.

mkdocs/pages/en/devbook/namespaces/resolve-namespace-from-receipt.md
```diff
@@ -0,0 +1,160 @@
...
+
+{{ tutorial.code_full('devbook/namespaces/resolve-namespace-from-receipt', ['py', 'js']) }}
+
+## Code Explanation
+
+### Defining the Transaction Hash
+
+{{ tutorial.code_snippet(['py:11:14', 'js:7:10']) }}
+
+The code defines the hash of a confirmed transaction that used a namespace alias instead of a plain address.
```
segfaultxavi: Not only an address alias, but also a mosaic alias, right?
We better not be misleading here then.

mkdocs/pages/en/devbook/namespaces/resolve-namespace-from-receipt.md
```diff
@@ -0,0 +1,160 @@
...
+Each resolution entry contains:
+
+* **Source:** Indicates from which transaction the resolved value applies, using a `primaryId`
+    (1-based index of the transaction in the block) and `secondaryId`
+    (1-based index within an aggregate, or `0` for standalone transactions).
+* **Resolved:** The actual address the alias pointed to from the indicated source onward.
+
+To determine the resolved value for the analyzed transaction, the code finds the last entry whose `primaryId` is
+less than or equal to the transaction's `primaryId`.
+This is the value the network used for that specific transaction, regardless of what the alias currently points to.
```
segfaultxavi: I would further explain why we need to keep that last one that is still lower than `primaryId`.
Something like:

```suggestion
This is the value the network used for that specific transaction, regardless of what the alias currently points to,
even if the alias was defined multiple times in the same block.
```

mkdocs/pages/en/devbook/namespaces/resolve-namespace-from-receipt.md
```diff
@@ -0,0 +1,160 @@
...
+
+* **Recipient and alias checks** (lines 6-9): The recipient field contains the encoded namespace alias, and the alias
+    flag check confirms it is a namespace alias rather than a regular address.
+    The mosaic alias check confirms that `E74B99BA41F4AFEE` is also a namespace alias.
+
+* **Resolved address** (lines 15-16): The address resolution shows that the namespace alias resolved to
+    `TCWYXKVYBMO4NBCUF3AXKJMXCGVSYQOS7ZG2TLI` for this transaction.
+
+* **Resolved mosaic ID** (lines 22-23): The mosaic resolution shows that `symbol.xym` (unresolved
+    `E74B99BA41F4AFEE`) resolved to mosaic ID `72C0212E67A08BCE` for this transaction.
```
segfaultxavi: It is worth explaining that neither the alias ID (or unresolved mosaic ID) nor the resolved mosaic ID are text strings.
It is impossible to know what the original alias string was (for example `symbol.xym`), because apps never provide it to the chain, they just provide an ID, probably generated with `generate_mosaic_id()` (link).
I say this is worth explaining, because some readers might be expecting to see `symbol.xym` recovered at the end.

The only thing an app can do (and probably should do) is check that the recovered mosaic ID matches the one they are expecting.

mkdocs/pages/en/devbook/namespaces/resolve-namespace-from-receipt.md
```diff
@@ -0,0 +1,160 @@
...
+
+Each statement can also have multiple entries if the alias was re-linked to a different value between transactions
+in the same block.
+If consecutive transactions resolve to the same value, only a single entry is stored.
+
+Each resolution entry contains:
+
+* **Source:** Indicates from which transaction the resolved value applies, using a `primaryId`
+    (1-based index of the transaction in the block) and `secondaryId`
+    (1-based index within an aggregate, or `0` for standalone transactions).
```
segfaultxavi:
```suggestion
    (1-based index within an <aggregate transaction:>, or `0` for standalone transactions).
```

segfaultxavi (changes requested):
Nice one, short and sweet!

mkdocs/overrides/devbook/namespaces/resolve-namespace-from-receipt.py
zero4862: TIL about rulers in VSCode 😅

mkdocs/pages/en/devbook/namespaces/resolve-namespace-from-receipt.md
zero4862: Good point

mkdocs/pages/en/devbook/namespaces/resolve-namespace-from-receipt.md
zero4862: Moving to mainnet adds some complexity. We would need to select a maintained node from Symbol Syndicate (https://symbol.fyi/nodes), get mainnet funds, and reproduce the transactions. We can add a note in our shared doc if we still want to apply the change in the future, but note it would also make the tutorial less consistent with others.  

Note that in case of a testnet reset, we would need to review other tutorials too. For example, keys like `0000` would need to be funded again, and we might want to update any hardcoded transaction hashes, mosaic IDs, or namespaces shown.

segfaultxavi: OK, agreed. I'll just add a note (just like in other tutorials we say that if you use address 0000 it _might_ already be funded.)

mkdocs/pages/en/devbook/namespaces/resolve-namespace-from-receipt.md
segfaultxavi: OK, yes, there is a way to recover the name from the ID. Good catch 😅

segfaultxavi (approved):

PR #1854, opened 2026-02-10
[mkdocs] add namespace and mosaic metadata tutorials

zero4862:
Add tutorials for namespace metadata and mosaic metadata.

Appplies the latest improvements from those tutorials to the former account metadata tutorial.

**Note:** It could make sense to review  the mosaic metadata tutorial after I submit the PR that includes the creation tutorial.

PR #1872, opened 2026-02-24
[mkdocs] chore: review consistency between tutorials

zero4862:
Addresses some improvements added in the namespaces and mosaic metadata tutorials back to the account metadata tutorial: https://github.com/symbol/symbol/pull/1854

Additionally, makes sure tutorial lines lenght are 120 chars or less.

### DIFF
--- a/mkdocs/overrides/devbook/accounts/account-metadata.log
+++ b/mkdocs/overrides/devbook/accounts/account-metadata.log
@@ -27,7 +27,7 @@ Waiting for aggregate transaction confirmation...
 aggregate transaction confirmed in 11 seconds
 
 --- Modifying existing metadata ---
-Fetching current metadata from /metadata?targetAddress=TCHBDENCLKEBILBPWP3JPB2XNY64OE7PYHHE32I&metadataType=0
+Fetching current metadata from /metadata?sourceAddress=TCHBDENCLKEBILBPWP3JPB2XNY64OE7PYHHE32I&targetAddress=TCHBDENCLKEBILBPWP3JPB2XNY64OE7PYHHE32I&scopedMetadataKey=E7F71DA42B54B310&metadataType=0
   Current value: alice
 Created embedded update transaction:
 {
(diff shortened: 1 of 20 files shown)


zero4862:
@segfaultxavi I'm not sure why `sourceAddress` was removed in [this review](https://github.com/symbol/symbol/pull/1854/changes/cb1f74cfd6eca5ca7d5ddcaa68dfaabc44bdc909#diff-69a34218926de4b831be3459c9a5a92e3cd35756c25a1fd72de9ecdc7708b1b1R145) (mosaic and metadata tutorials). Without it, the query could return multiple entries if another account adds the same metadata entry to a mosaic or namespace. Let me know if you want me to add it back (or just add it back in this PR).

mkdocs/overrides/devbook/accounts/account-metadata.log
```diff
@@ -27,7 +27,7 @@ Waiting for aggregate transaction confirmation...
 aggregate transaction confirmed in 11 seconds
 
 --- Modifying existing metadata ---
-Fetching current metadata from /metadata?targetAddress=TCHBDENCLKEBILBPWP3JPB2XNY64OE7PYHHE32I&metadataType=0
+Fetching current metadata from /metadata?sourceAddress=TCHBDENCLKEBILBPWP3JPB2XNY64OE7PYHHE32I&targetAddress=TCHBDENCLKEBILBPWP3JPB2XNY64OE7PYHHE32I&scopedMetadataKey=E7F71DA42B54B310&metadataType=0
```
zero4862, line 30: To match the current implementation, it was run before the latest edit.

segfaultxavi (approved):
Thank you very much for keeping track of these chores 🙏

Just a few comments:
- 120 columns is a soft limit (unlike 75, which is strict). But I agree it's good to be consistent. I'll respect this limit in the future!
- I've fixed a couple places were the indentation was 2 spaces instead of 4.
- It is Markdown good practice to only use - or * for lists, but not both on the same file. I've fixed a couple of them.
- We don't use two consecutive blank lines in the Markdown files.

PR #1873, opened 2026-02-25
[mkdocs] docs: revoke mosaic tutorial

zero4862:
Adds revoke mosaic tutorial based on https://github.com/symbol/symbol/blob/main/docs/index.md#tutorial-working-with-rules-restrictions

### DIFF
--- a/mkdocs/overrides/devbook/mosaics/create-mosaic.mjs
+++ b/mkdocs/overrides/devbook/mosaics/create-mosaic.mjs
@@ -81,7 +81,7 @@ try {
 	// --- CREATING MOSAIC DEFINITION ---
 	console.log('\n--- Creating mosaic definition ---');
 
-	const nonce = Date.now() & 0xFFFFFFFF;
+	const nonce = Math.floor(Date.now() / 1000) & 0x7FFFFFFF;
 	console.log('Mosaic nonce:', nonce);
 
 	const definitionTx =
--- a/mkdocs/overrides/devbook/mosaics/revoke-mosaic.py
+++ b/mkdocs/overrides/devbook/mosaics/revoke-mosaic.py
@@ -16,11 +16,8 @@ print(f'Using node {NODE_URL}')
 # Helper function to fetch account mosaic balances
 def get_account_mosaics(address):
 	account_path = f'/accounts/{address}'
-	print(
-		f'Fetching account information from {account_path}')
-	with urllib.request.urlopen(
-		f'{NODE_URL}{account_path}'
-	) as response:
+	print(f'Fetching account information from {account_path}')
+	with urllib.request.urlopen(f'{NODE_URL}{account_path}') as response:
 		response_json = json.loads(response.read().decode())
 		return response_json['account']['mosaics']
 
--- a/mkdocs/overrides/devbook/mosaics/revoke-mosaic.py
+++ b/mkdocs/overrides/devbook/mosaics/revoke-mosaic.py
@@ -12,6 +12,19 @@ NODE_URL = os.getenv(
 	'NODE_URL', 'https://reference.symboltest.net:3001')
 print(f'Using node {NODE_URL}')
 
+
+# Helper function to fetch account mosaic balances
+def get_account_mosaics(address):
+	account_path = f'/accounts/{address}'
+	print(
+		f'Fetching account information from {account_path}')
+	with urllib.request.urlopen(
+		f'{NODE_URL}{account_path}'
+	) as response:
+		response_json = json.loads(response.read().decode())
+		return response_json['account']['mosaics']
+
+
 SIGNER_PRIVATE_KEY = os.getenv('SIGNER_PRIVATE_KEY',
 	'0000000000000000000000000000000000000000000000000000000000000000')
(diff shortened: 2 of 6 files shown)


mkdocs/overrides/devbook/mosaics/create-mosaic.mjs
```diff
@@ -81,7 +81,7 @@ try {
 	// --- CREATING MOSAIC DEFINITION ---
 	console.log('\n--- Creating mosaic definition ---');
 
-	const nonce = Date.now() & 0xFFFFFFFF;
+	const nonce = Math.floor(Date.now() / 1000) & 0x7FFFFFFF;
```
zero4862, line 84: Converted to seconds to avoid negative values from JS signed 32-bit bitwise ops.

segfaultxavi: It's not enough to just mask with 0x7FFF_FFFF ?

mkdocs/overrides/devbook/mosaics/revoke-mosaic.py
```diff
@@ -0,0 +1,123 @@
...
+		response_json = json.loads(response.read().decode())
+		median_mult = response_json['medianFeeMultiplier']
+		minimum_mult = response_json['minFeeMultiplier']
+		fee_mult = max(median_mult, minimum_mult)
+		print(f'  Fee multiplier: {fee_mult}')
+
+	# --- REVOKING MOSAIC ---
+	print('\n--- Revoking mosaic ---')
+
+	mosaic_id_hex = os.getenv('MOSAIC_ID', '0')
```
segfaultxavi: Can we provide an actual mosaic ID so the tutorial works out of the box?
(Until the next testnet reset)

mkdocs/overrides/devbook/mosaics/revoke-mosaic.py
```diff
@@ -0,0 +1,123 @@
...
+					attempt, 'seconds')
+				break
+			if status['group'] == 'failed':
+				raise Exception(
+					f'Mosaic revocation failed: {status["code"]}')
+		except urllib.error.HTTPError:
+			print('  Transaction status: unknown')
+
+	# --- VERIFYING REVOCATION ---
+	print('\n--- Verifying revocation ---')
```
segfaultxavi, line 128: This is not very useful if we don't show the balance before the revoke, no?
Maybe you can put this in a function and just call it before and after executing the revoke tx.

mkdocs/pages/en/devbook/mosaics/revoke-mosaic.md
```diff
@@ -0,0 +1,135 @@
...
+
+* **Type:** Mosaic supply revocation transactions use the type `mosaic_supply_revocation_transaction_v1`.
+
+* **Source address:** The address of the account holding the mosaic units to revoke.
+    This can be any account that currently holds units of the specified mosaic.
+
+* **Mosaic:** An object containing the <mosaic ID:> and the amount to revoke.
+
+* **Amount:** The number of atomic units to revoke from the source account.
+    Since the mosaic in this example has a [divisibility](../../textbook/mosaics.md#divisibility) of `2`,
```
segfaultxavi: We don't know the divisibility of the mosaic that the user will choose, right?
Either we provide a default mosaic or rephrase this paragraph.

mkdocs/pages/en/devbook/mosaics/revoke-mosaic.md
```diff
@@ -0,0 +1,135 @@
...
+<get:/transactionStatus/{hash}> endpoint until the status changes to `confirmed`.
+
+### Verifying the Revocation
+
+{{ tutorial.code_snippet(['py:112:120', 'js:123:132']) }}
+
+To verify the revocation, the code retrieves the source account's mosaic balances from the
+<get:/accounts/{accountId}> endpoint.
+
+A successful response shows that the revoked mosaic's balance has decreased by the specified amount.
```
segfaultxavi: Not unless we also show the balance before executing the transaction :)

segfaultxavi (changes requested):
Another nice little one!
My only concerns are about providing a default mosaic ID and showing the balance before and after the transaction.

mkdocs/overrides/devbook/mosaics/create-mosaic.mjs
zero4862: It would work too. However, I converted the value to seconds to stay consistent with the Python tutorial, which also uses seconds.

segfaultxavi: Got it, makes sense, yeah.

segfaultxavi (changes requested):
Nice! The only remaining complaint is that the code is again too narrow :P

zero4862:
@segfaultxavi fixed in https://github.com/symbol/symbol/pull/1873/commits/989e87d9800d07d118a5b2397318d2d47fe9f1a7  🤞

segfaultxavi (approved):
👍

PR #1892, opened 2026-03-05
[mkdocs] docs: add listen to transaction flow tutorial

zero4862:
Adds listen to new transaction flow tutorial (WebSockets) based on https://github.com/symbol/symbol/blob/main/docs/index.md?rgh-link-date=2026-03-04T12%3A22%3A31Z#tutorial-working-with-websockets


mkdocs/pages/en/devbook/reference/websockets/index.md
```diff
@@ -77,9 +77,9 @@ All channels share the same response format, which is:
 
     [FinalizedBlockDTO](../rest/symbol.md#model-FinalizedBlockDTO)
 
-### `confirmedAdded`
+### `confirmedAdded/{address}`
 
-`ws:confirmedAdded/{address}`
+`ws:confirmedAdded`
```
zero4862: Glossary item links containing `/` could not be resolved correctly, so `/{address}` was moved to the title as a workaround.

mkdocs/pages/en/devbook/websockets/listen-transaction-flow.md
```diff
@@ -0,0 +1,158 @@
...
+| Step                                                        | Related documentation                                                      |
+| ----------------------------------------------------------- | ---------------------------------------------------------------------------|
+| [Subscribe to confirmedAdded](#subscribing-to-channels)     | <ws:confirmedAdded>                                                        |
+| [Subscribe to unconfirmedAdded](#subscribing-to-channels)   | <ws:unconfirmedAdded>                                                      |
+| [Subscribe to unconfirmedRemoved](#subscribing-to-channels) | <ws:unconfirmedRemoved>                                                    |
+| [Handle transaction messages](#handling-messages)           | [TransactionInfoDTO](../reference/rest/symbol.md#model-TransactionInfoDTO) |
+
+## Next Steps
+
+To detect rejected transactions and their error codes, see the Listening to Transaction Errors tutorial.
```
zero4862, line 174: Todo: I'll add the link to the next tutorial in the next PR (once it's created).

mkdocs/pages/en/devbook/reference/websockets/index.md
segfaultxavi: Yeah, I had not noticed `/` is a special character and needs to be escaped:
https://realtimeprojects.github.io/mkdocs-ezglossary/usage/definition/#special-characters

Try using `&#47;` as the docs suggest, and compare which solution is more comfortable to use.
I am not against moving `/{address}` to the title, but then `<ws:confirmedAdded|confirmedAdded/{address}>` is pretty horrible.

mkdocs/pages/en/devbook/websockets/listen-transaction-flow.md
```diff
@@ -0,0 +1,158 @@
...
+---
+
+# Listening to Transaction Flow
+
+Symbol provides WebSocket channels that send real-time notifications as a <transaction:> moves through the confirmation
+process for a specific <account:>.
+Compared to polling the <get:/transactionStatus/{hash}> endpoint, WebSockets push updates as they happen without
+the overhead of repeated API calls.
+
+This tutorial shows how to subscribe to all three channels and display updates as a transaction progresses from
```
segfaultxavi: What three channels?
If you don't want to introduce them yet, you can say "... subscribe to the three channels required to display updates as ..."

segfaultxavi (changes requested):
My main concern is that if the reader does not initiate a transaction, nothing is shown, which is a departure from the rest of tutorials that are self-contained.

If we announce the transaction in the code (after subscribing to ws), we can show an alternative method to wait for its confirmation, and this will add a lot of value to the tutorial.

I am not familiar with asyncio, though, so I don't know how cumbersome the resulting code will be. Can you try it?

zero4862:
@segfaultxavi great suggestion on making it self-contained. I tried it and it works well. The important part is to make the announcement after subscribing to the channels to make sure there are not race conditions. 

Let's agree on the code first before editing the tutorial text (as it will also affect the other websocket tutorial I'm writing). Here's the Python snippet: https://gist.github.com/zero4862/00903747ed98a93df668df2158d33ae2

segfaultxavi:
The code looks good. Maybe group together everything related to sending the transaction (fetching time and fees, building, announcing and confirming). In this way in the text we can say "this is a regular transfer, built and announced as usual".
Otherwise, it looks like there is a reason for subscribing to the ws right before announcing, but after building the tx.

mkdocs/overrides/devbook/websockets/listen-new-blocks.mjs
```diff
@@ -3,56 +3,61 @@ const NODE_URL = process.env.NODE_URL ||
 const WS_URL = NODE_URL.replace('http', 'ws') + '/ws';
 console.log(`Using node ${NODE_URL}`);
 
-const websocket = new WebSocket(WS_URL);
-
-// Connect to websocket endpoint
-const uid = await new Promise((resolve) => {
+try {
```
zero4862, line 6: Added a try / catch to match listen-transaction-flow.mjs

segfaultxavi (approved):
I made:
- Whitespace fixes
- Clarified that the sender and the receiver are the same account only if you use the default values.
- Changed `!!! important` to `!!! warning` which is valid admonition type.

The JS code shows the scrollbar because it's 1 character too wide in a couple of places, but it's no big deal and I don't want to have to re-number all snippets :D

PR #1897, opened 2026-03-09
[catapult/sdk] Amend out-of-bounds vector access on MemoryStream

AndreaLanfranchi:
When m_buffer is pristine and input buffer is empty then m_buffer[0] is out-of-bounds.
Early exit also saves void calls to resize and memcpy.

### DIFF
--- a/client/catapult/sdk/src/extensions/MemoryStream.cpp
+++ b/client/catapult/sdk/src/extensions/MemoryStream.cpp
@@ -32,6 +32,8 @@ namespace catapult { namespace extensions {
 	{}
 
 	void MemoryStream::write(const RawBuffer& buffer) {
+		if(0u == buffer.Size)
+			return;
 		m_buffer.resize(std::max<size_t>(m_buffer.size(), m_position + buffer.Size));
 		utils::memcpy_cond(&m_buffer[m_position], buffer.pData, buffer.Size);
 		m_position += buffer.Size;
--- a/client/catapult/sdk/src/extensions/MemoryStream.cpp
+++ b/client/catapult/sdk/src/extensions/MemoryStream.cpp
@@ -32,11 +32,7 @@ namespace catapult { namespace extensions {
 	{}
 
 	void MemoryStream::write(const RawBuffer& buffer) {
-		if(0u == buffer.Size)
-			return;
-
-		m_buffer.resize(std::max<size_t>(m_buffer.size(), m_position + buffer.Size));
-		std::memcpy(&m_buffer[m_position], buffer.pData, buffer.Size);
+		m_buffer.insert(m_buffer.end(), buffer.pData, buffer.pData + buffer.Size);
 		m_position += buffer.Size;
 	}
 
--- a/client/catapult/sdk/src/extensions/MemoryStream.cpp
+++ b/client/catapult/sdk/src/extensions/MemoryStream.cpp
@@ -32,7 +32,11 @@ namespace catapult { namespace extensions {
 	{}
 
 	void MemoryStream::write(const RawBuffer& buffer) {
-		m_buffer.insert(m_buffer.end(), buffer.pData, buffer.pData + buffer.Size);
+		if(0u == buffer.Size)
+			return;
+
+		m_buffer.resize(std::max<size_t>(m_buffer.size(), m_position + buffer.Size));
+		std::memcpy(&m_buffer[m_position], buffer.pData, buffer.Size);
 		m_position += buffer.Size;
 	}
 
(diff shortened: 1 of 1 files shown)


client/catapult/sdk/src/extensions/MemoryStream.cpp
```diff
@@ -32,8 +32,10 @@ namespace catapult { namespace extensions {
 	{}
 
 	void MemoryStream::write(const RawBuffer& buffer) {
+		if(0u == buffer.Size)
+			return;
```
Wayonb: empty line after the if statement.

client/catapult/sdk/src/extensions/MemoryStream.cpp
```diff
@@ -32,8 +32,10 @@ namespace catapult { namespace extensions {
...
 		m_buffer.resize(std::max<size_t>(m_buffer.size(), m_position + buffer.Size));
-		utils::memcpy_cond(&m_buffer[m_position], buffer.pData, buffer.Size);
+		std::memcpy(&m_buffer[m_position], buffer.pData, buffer.Size);
```
Wayonb: Should we also change the below to prevent future problems?
``&m_buffer[m_position]``  -> ``m_buffer.data() + m_position``

Wayonb (changes requested):

client/catapult/sdk/src/extensions/MemoryStream.cpp
AndreaLanfranchi: Thank you. [c2e8a5f](https://github.com/symbol/symbol/pull/1897/commits/c2e8a5f4bcdf7a9c57d1b664e66f0310ba9fbab5)

client/catapult/sdk/src/extensions/MemoryStream.cpp
AndreaLanfranchi: Imo definitely not. For several reasons:

- We already guarded the possible out-of-bounds and ...
- This `&m_buffer[m_position]` internally IS NOT equivalent to this `m_buffer.data() + m_position`: well actually it is in this speciific case as the underlying element type is 1 byte but the correct form should be `m_buffer.data() + m_position * sizeof(element_type)`
- on an empty vector the data() method may return different implementation defined values. It could be nullptr or a non-null pointer. Doing ptr math on such a value is dangerous as, per cpp definition, If vector is empty, data() is not dereferenceable.

While is obvious the edge case defined here might cause no serious harm (if input is empty the destination vector is not resized and memcpy tries to copy a zero count - hence nothing) from the POV of correctness is better to dereference random access operator.

But we actually can be more concise by reducing it all to two lines:

```
	void MemoryStream::write(const RawBuffer& buffer) {
		m_buffer.insert(m_buffer.end(), buffer.pData, buffer.pData + buffer.Size);
		m_position += buffer.Size;
	}
```

This will always be consistent even in the edge case described here (even if I'm not a fan about doing pointer math when one operand is possibly nullptr).

Wdyt ?

Edit : The last suggestion has been overridden by #5058a17 and #f5d537f

client/catapult/sdk/src/extensions/MemoryStream.cpp
```diff
@@ -32,8 +32,7 @@ namespace catapult { namespace extensions {
 	{}
 
 	void MemoryStream::write(const RawBuffer& buffer) {
-		m_buffer.resize(std::max<size_t>(m_buffer.size(), m_position + buffer.Size));
-		utils::memcpy_cond(&m_buffer[m_position], buffer.pData, buffer.Size);
+		m_buffer.insert(m_buffer.end(), buffer.pData, buffer.pData + buffer.Size);
```
Jaguar0625: seems like you're changing behavior by always appending?

previously it was writing at m_position, which might not be the end. seems like this is why CI is failing now.

AndreaLanfranchi: You're damn right ... sorry.
Amending now

client/catapult/sdk/src/extensions/MemoryStream.cpp
Wayonb: - Yes, I agree, the guard will prevent any out-of-bounds issue 👍🏾 
- As you state, the size of the element_type is 1 byte, so this is why I left this off.
- If an empty vector returns nullptr (nullptr + 0 is still safe) vs &m_buffer[m_position] is UB.

The guard is actually fixed, so I am fine with it.

Wayonb (approved):
👍🏾

PR #1900, opened 2026-03-12
[jenkins] fix: install the exact package version for 'npm install'

Wayonb:
problem: for builds that use the internal Nexus repo, Jenkins will delete the package-lock.json.
         npm install will then install the latest version of dependencies.
         In certain cases, the newer updates can break or have bugs that cause failures in CI.
solution: add the save-exact flag to npm to not upgrade the packages

Just for reference, this was causing the REST build to fail due to an update to Sinon package(https://github.com/sinonjs/sinon/issues/2678)


Jaguar0625 (approved):
👍

PR #1905, opened 2026-03-18
[jenkins] fix: switch the default to public

Wayonb:
problem: There have been a lot of dependency-related failures for JavaScript.
                  This is due to Jenkins removing the package-lock.json for private builds,
                  and some projects fail with the newer deps.
solution: Set the default build to public, which will not delete the package-lock file.


Jaguar0625 (approved):
👍

AnthonyLaw (approved):
👍🏼

PR #1906, opened 2026-03-19
docs: add prove mosaic state tutorial

zero4862:
Adds prove mosaic state tutorial based on https://github.com/symbol/symbol/blob/main/docs/index.md#tutorial-working-with-proofs


mkdocs/pages/en/devbook/chain/prove-mosaic-state.md
```diff
@@ -0,0 +1,187 @@
...
+| `0x8001` | `STATE_HASH_DOES_NOT_MATCH_ROOTS` | `SHA3-256(roots)` does not match `state_hash`. The block header and sub-cache roots are inconsistent.                                                                |
+| `0x8002` | `UNANCHORED_PATH_TREE`            | The tree path's root node does not hash to any of the sub-cache roots.                                                                                               |
+| `0x8003` | `LEAF_VALUE_MISMATCH`             | The leaf's value does not match `hashed_value`.                                                                                                                      |
+| `0x8004` | `UNLINKED_NODE`                   | A node in the path is not linked by its parent.                                                                                                                      |
+| `0x8005` | `PATH_MISMATCH`                   | The path reconstructed from root to leaf does not match `encoded_key`.                                                                                               |
```
zero4862: In the future, we might want to check if we can generate this from the SDK reference and link to it.

mkdocs/pages/en/devbook/chain/prove-mosaic-state.md
```diff
@@ -0,0 +1,187 @@
...
+The `stateHashSubCacheMerkleRoots` array contains the individual root hash for each sub-cache
+(accounts, mosaics, namespaces, and so on).
+
+### Fetching the Tree Path
+
+{{ tutorial.code_snippet(['py:73:97', 'js:81:109']) }}
+
+The <get:/mosaics/{mosaicId}/merkle> endpoint returns the raw path through the mosaic sub-cache
+<Patricia tree:>, from the root down to the leaf that stores the mosaic's hashed value.
+<dy:Merkle.deserializePatriciaTreeNodes> converts this raw binary into a list of tree nodes.
```
zero4862: `dy:Merkle.deserializePatriciaTreeNodes` works for JS but **not** for Python:

> WARNING -  page 'devbook/chain/prove-mosaic-state/' refers to undefined glossary entry py:Merkle.deserialize_patricia_tree_nodes

However, `dy:Merkle.provePatriciaMerkle` works for both JS and Python.

@segfaultxavi any idea?

mkdocs/pages/en/textbook/blocks.md
segfaultxavi, line 1: Since you're touching this file, can you rename the header `## State Hashes` to `## Block Hashes`, and use proper word capitalization for its 3 sub-headers?

And then change links to #state-hashes accordingly in other pages.

mkdocs/pages/en/textbook/blocks.md
```diff
@@ -245,5 +245,61 @@ Each Patricia tree has its own root hash.
 The state hash is computed as `SHA3-256(root₁ || root₂ || ... || rootₙ)`, a single hash over the concatenated roots
 of all sub-caches.
```
segfaultxavi, line 247: "Sub-cache" needs to be defined. For example, in the paragraph above:

```
Symbol maintains separate <Patricia trees:> for each type of on-chain data, called _sub-caches_:
```

mkdocs/pages/en/textbook/blocks.md
```diff
@@ -245,5 +245,61 @@ Each Patricia tree has its own root hash.
...
 
+For example, to prove that a mosaic has a particular hashed value, only the nodes along the path from the sub-cache root
+to the leaf are needed.
+The mosaic's encoded key (`3A4C...` in the diagram) is the SHA3-256 hash of its identifier, and each nibble
+(half-byte, one hex digit) of this key determines which child to follow at each level of the tree:
```
segfaultxavi: This paragraph is very hard to follow. "Hashed value" needs to be defined.

I would try something like:

```md
For example, to prove that a mosaic had a particular definition in a certain block, the hash of the
definition must be present in the tree.
Locating hashes inside Patricia trees is very efficient due to their properties:
* The Mosaic's ID is hashed with SHA3-256 to obtain its _encoded key_.
* Each _nibble_ (half-byte, one hex digit) of this key indicates which branch to follow at each level of the tree.
* If a final leaf is reached, it contains the hash of the definition of the mosaic.
```

mkdocs/pages/en/textbook/blocks.md
```diff
@@ -245,5 +245,61 @@ Each Patricia tree has its own root hash.
...
+
+Each branch node has up to 16 links (one per nibble value `0`–`F`), each pointing to a child node.
+Each node also carries a **path**: a compressed sequence of nibbles from the encoded key,
+shared by all entries below it.
+In the example above the branch nodes have empty paths, but a node might carry a long path that skips many nibbles at
+once when there are no other entries that would require branching at those positions.
+The full encoded key is reconstructed by concatenating each followed nibble and node path from root to leaf:
+`3` + `A` + `4` + `C` + leaf path `540D7E...559E` = `3A4C540D7E...559E`.
+
+The leaf stores the SHA3-256 hash of the serialized entry (the value).
```
segfaultxavi: We need to be consistent with the nomenclature: Serialized entry, Hashed value, Mosaic definition... chose one :)

mkdocs/pages/en/devbook/chain/prove-mosaic-state.md
segfaultxavi: Yeah, it's because `deserialize_patricia_tree_nodes` does not have a help string (docstring).
The Python plugin (mkdocstrings-python) does not create an entry for it by default, whereas the JS plugin (typedoc) does.

If I add the option to generate entries for methods without help strings, a lot of chaff is added to the docs,
and what's worse, we get collisions between methods with the same name 🤦

Since these methods are useless in the docs anyway (they have no help text), I would just use code tags (backticks) instead of `<dy:>`.

mkdocs/pages/en/devbook/chain/prove-mosaic-state.md
segfaultxavi: We do have them in JS, but not in Python:
https://docs.symboltest.net/en/devbook/reference/ts/symbol/merkle/classes/PatriciaMerkleProofResult/
https://docs.symboltest.net/en/devbook/reference/py/symbol/Merkle/#symbolchain.symbol.Merkle.PatriciaMerkleProofResult

The Python source simply does not have this information in the docstring.
I've created a task to review all these docstrings.

mkdocs/pages/en/devbook/chain/prove-mosaic-state.md
```diff
@@ -0,0 +1,187 @@
+---
+title: Check Mosaic State
```
segfaultxavi: Mosaic State sounds a bit vague to me. Maybe:

```suggestion
title: Prove Mosaic Definition
```

And then the full title below can be "Prove a Mosaic's Definition".
I'm not sure about the "Against the Chain" part.

mkdocs/pages/en/devbook/chain/prove-mosaic-state.md
```diff
@@ -0,0 +1,187 @@
...
+---
+
+# Checking a Mosaic's State Against the Chain
+
+Every Symbol <block:> header includes a `stateHash` that covers the full chain state, including all <mosaic:>
+definitions.
```
segfaultxavi:
```suggestion
Every Symbol <block:> header has a `stateHash` that covers the full chain state, including all <mosaic:> definitions.
```

To avoid repeating "includes".

mkdocs/pages/en/devbook/chain/prove-mosaic-state.md
```diff
@@ -0,0 +1,187 @@
...
+By requesting a <Patricia tree:> proof and verifying it locally, you can confirm that the data a node returns matches
```
segfaultxavi:
```suggestion
By requesting a proof and verifying it locally, you can confirm that the data a node returns matches
```

Too soon to start giving details.

mkdocs/pages/en/devbook/chain/prove-mosaic-definition.md
```diff
@@ -0,0 +1,187 @@
...
+The response includes all the properties that make up the mosaic's on-chain state: `version`, `id`, `supply`,
+`startHeight`, `ownerAddress`, `revision`, `flags`, `divisibility`, and `duration`.
+
+### Computing the Key and Value Hashes
+
+{{ tutorial.code_snippet(['py:35:53', 'js:46:63']) }}
+
+To verify the mosaic's state, the code must reproduce the exact hash that the chain stores internally.
+This requires serializing all mosaic properties into a binary buffer in the exact field order and sizes defined by the
+<ser:MosaicEntry> schema, then computing the SHA3-256 hash of the result.
```
segfaultxavi, line 53: Nice link!
I would warn users that MosaicEntry contains a MosaicDefinition structure, which in turn contains a MosaicProperties structure.
Otherwise, on first glance, the struct only seems to have 4 fields and the code is serializing a few more.

mkdocs/pages/en/devbook/chain/prove-mosaic-definition.md
```diff
@@ -0,0 +1,187 @@
...
+{{ tutorial.code_snippet(['py:73:97', 'js:81:109']) }}
+
+The <get:/mosaics/{mosaicId}/merkle> endpoint returns the raw path through the mosaic sub-cache
+<Patricia tree:>, from the root down to the leaf that stores the mosaic's hashed value.
+<dy:Merkle.deserializePatriciaTreeNodes> converts this raw binary into a list of tree nodes.
+
+The code then walks the deserialized path and prints each node for inspection.
+Branch nodes show their active links and which [nibble](../../textbook/blocks.md#state-hash) was followed to reach the
+next node.
+The leaf node at the end stores the remaining path nibbles and the hashed value.
```
segfaultxavi, line 91: We should say that we are traversing and printing the tree here for educational purposes, because the SDK already does it for us, as shown below.

mkdocs/pages/en/devbook/chain/prove-mosaic-state.md
```diff
@@ -0,0 +1,187 @@
...
+| `0x8003` | `LEAF_VALUE_MISMATCH`             | The leaf's value does not match `hashed_value`.                                                                                                                      |
+| `0x8004` | `UNLINKED_NODE`                   | A node in the path is not linked by its parent.                                                                                                                      |
+| `0x8005` | `PATH_MISMATCH`                   | The path reconstructed from root to leaf does not match `encoded_key`.                                                                                               |
+
+!!! warning "Height consistency"
+
+    The mosaic properties, block header, and tree path must all reflect the same chain state.
+    If a new block is confirmed between requests, the mosaic state may have changed and the proof will fail.
+    When this happens, re-fetch all three pieces of data and try again.
+    If the proof still fails after two or three retries, the data itself may be invalid.
```
segfaultxavi: It's good that you call attention to this, but it sounds like a very scary situation.
You could clarify that this would only happen if the mosaic definition is changed between the code fetches it and the code fetches the block headers.
Given how infrequent mosaic redefinitions are, this is a very unlikely event.

There really is no bulletproof solution, right? Can we know when a mosaic was defined and check that it happened before the block headers were fetched?

segfaultxavi (changes requested):
Good job with this one, it is a tough one.

mkdocs/pages/en/devbook/chain/prove-mosaic-state.md
zero4862: The issue isn't limited to the mosaic being redefined though since the `stateHash` changes with every new block because it covers the entire chain state (all accounts, mosaics, namespaces, etc.). So if any block is confirmed between our API requests, the proof can fail.

There's no bulletproof solution via the REST API since we can't fetch all three pieces atomically.

I'm rewording the warning to be less alarming.

segfaultxavi (approved):

PR #1913, opened 2026-03-25
[mkdocs] docs: query block rewards

zero4862:
Adds Query block rewards tutorial.


segfaultxavi (approved):
Nothing to comment!

I liked the info boxes about the receipt type filter and signerPublicKey vs. receipt.
It's valuable information, I was wondering about that myself :)

PR #1914, opened 2026-03-25
[mkdocs] refactor: use SDK Address methods for hex-to-address conversion

zero4862:
Spotted this improvement while working in our of the latest tutorials.

Replaces manual `bytes.fromhex/Buffer.from` patterns with the SDK built-in methods across namespace tutorials.


segfaultxavi (approved):
Thanks for this :pray:

PR #1917, opened 2026-03-27
[mkdocs] docs: transaction batching

zero4862:


segfaultxavi (approved):
I just made minor changes. All good!

We need to do a general pass on all tutorials to add `<ser:>` links to the serialization tables.

PR #1918, opened 2026-03-29
[client/catapult] Expand limits on SpinReaderWriterLock

AndreaLanfranchi:
The `SpinReaderWriterLock` used a `std::atomic<uint16_t>` with a compact bit layout that limited the lock to 128 concurrent writers and 256 concurrent readers. Exceeding these limits would cause silent counter overflow, corrupting the lock state.

## What's the issue?
With the original 16-bit layout, counter overflow was silent: incrementing the reader counter past its maximum would carry bits into the writer-count field, incorrectly marking pending writers; similarly, incrementing the writer counter past its maximum would set the `Active_Writer_Flag` bit, causing inconsistent `isWriterActive`/`isWriterPending` results and potential deadlock. The `promoteToWriter()` path was also unprotected, and the CAS loop in `acquireWriter()` lacked backoff, causing tight CPU spin under contention.

- Switched internal lock state from `std::atomic<uint16_t>` to `std::atomic<uint32_t>` with a new bit layout:
  - Bit 31: Active writer flag
  - Bits 30–16 (15 bits): Writer count (pending + active), exposed as `Writer_Count_Mask` / `Writer_Count_Increment`
  - Bits 15–0 (16 bits): Active reader count
- This raises the representable limits to **32,768 max writers** and **65,535 max readers**.
- Added a fail-fast overflow guard in `acquireReader()`: if the reader count field is already at its maximum (`Reader_Mask`), a `catapult_runtime_error("max reader count (65535) reached")` is thrown before the CAS increment, preventing carry into writer bits.
- Replaced the unconditional `fetch_add` in `acquireWriter()` with a CAS loop (with `YieldStepper` backoff) that checks whether the writer count field is at its maximum (`Writer_Count_Mask`) before each attempt; throws `catapult_runtime_error("max writer count (32767) reached")` if saturated, preventing carry into the `Active_Writer_Flag` bit.
- Applied the same saturation guard and `YieldStepper`-backed CAS loop to `promoteToWriter()`, replacing the previously unprotected `fetch_add`, so the promotion path is equally safe under extreme contention.
- Renamed `Pending_Writer_Mask` → `Writer_Count_Mask` and `Pending_Writer_Increment` → `Writer_Count_Increment` to reflect that bits 30–16 track total writers (pending + active), not just pending; deprecated aliases are retained for internal backward compatibility.
- Corrected inline documentation: the note now states the actual limits (32,768 writers / 65,535 readers) and the bit-layout comment now reads "Writer count (pending + active)".

These changes are not breaking: existing callers that stay within the new higher limits see no behavioral difference.

Three new unit tests were added in a `// region counter bounds` section of `SpinReaderWriterLockTests.cpp`:

- **`CanAcquireMoreThan256ConcurrentReaderLocks`**: Acquires 300 simultaneous reader locks using `BasicSpinReaderWriterLock<NoOpReaderNotificationPolicy>` and asserts that no writer bits are set, then releases all locks and asserts the lock is fully clean.
- **`CanAcquireMoreThan128SequentialWriterLocks`**: Acquires and releases 200 writer locks sequentially, asserting the correct pending/active writer state on each acquire and a clean lock state after all releases.
- **`AcquireReaderThrowsWhenReaderCountSaturated`**: Fills all 65,535 reader slots, then asserts that the next `acquireReader()` call throws `catapult_runtime_error` and that the lock state remains intact after the throw.

An explicit `#include <vector>` was also added to the test file to avoid reliance on transitive includes.

---- 

## Requirements

* [ ] My pull request title is prefixed with the directory and subdirectory it affects.
* [ ] My pull request contains a single change or feature.
* [ ] My commits are clearly labeled with one of the following: **feat | bug | fix | build | perf | task**
* [ ] My code has been linted.
* [ ] My code has comments, particularly in any hard to understand areas.
* [ ] My code follows the style guidelines.
* [ ] I have written and supplied test cases (either unit tests or end-to-end, where relevant).
* [ ] I have made any corresponding changes to the documentation (where relevant).

---

📱 Kick off Copilot coding agent tasks wherever you are with [GitHub Mobile](https://gh.io/cca-mobile-docs), available on iOS and Android.


AndreaLanfranchi:
@copilot apply changes based on the comments in [this thread](https://github.com/symbol/symbol/pull/1918#pullrequestreview-4026894461)

AndreaLanfranchi:
@copilot apply changes based on the comments in [this thread](https://github.com/symbol/symbol/pull/1918#pullrequestreview-4029258582)

client/catapult/src/catapult/utils/SpinReaderWriterLock.h
```diff
@@ -30,21 +30,28 @@ namespace catapult { namespace utils {
...
+		// Mask for total writer count bits (pending + active)
+		static constexpr uint32_t Writer_Count_Mask = 0b01111111111111110000000000000000;
+
+		static constexpr uint32_t Reader_Mask = 0b00000000000000001111111111111111;
+		static constexpr uint32_t Writer_Mask = Writer_Count_Mask | Active_Writer_Flag;
+
+		static constexpr uint32_t Active_Reader_Increment = 0b00000000000000000000000000000001;
+
+		// Increment for total writer count (pending + active)
+		static constexpr uint32_t Writer_Count_Increment = 0b00000000000000010000000000000000;
```
Wayonb, line 61: Is there a reason to change the representation from hex to bytes?

client/catapult/tests/catapult/utils/SpinReaderWriterLockTests.cpp
```diff
@@ -41,6 +42,70 @@ namespace catapult { namespace utils {
...
+
+		// Act + Assert: the 65536th acquire must throw
+		EXPECT_THROW({ std::ignore = lock.acquireReader(); }, catapult_runtime_error);
+
+		// Assert: lock state is still intact after the throw attempt
+		EXPECT_FALSE(lock.isWriterPending());
+		EXPECT_FALSE(lock.isWriterActive());
+		EXPECT_TRUE(lock.isReaderActive());
+	}
+
```
Wayonb: Can you also add a test for when ``acquireWriter()`` throws?

Wayonb (changes requested):
Looks good just missing a test 👍🏾 
@Jaguar0625 should take a look also.

client/catapult/src/catapult/utils/SpinReaderWriterLock.h
AndreaLanfranchi: The main reason is visual representation of bytes layout into the atomic uint32 storage. Specifically for Reader_Mask is far easier to understand how bytes are accounted.

client/catapult/tests/catapult/utils/SpinReaderWriterLockTests.cpp
AndreaLanfranchi: Wilco

Edit ... after an excruciating attempt series to make linters happy I give up: too many conflicting inputs from ci.
I'll leave this open if you.

AndreaLanfranchi: @Wayonb ⬆️

Wayonb: I push some changes to fix linting

client/catapult/src/catapult/utils/SpinReaderWriterLock.h
```diff
@@ -25,11 +25,14 @@
 #include <atomic>
 #include <thread>
 
-namespace catapult { namespace utils {
+namespace catapult {
+	namespace utils {
+		template<typename TReaderNotificationPolicy>
+		struct SpinReaderWriterLockTestAccessor;
+	}
+}
```
AndreaLanfranchi, line 33: Thank you but let me shout ... WHY !!! 🤦

Wayonb (approved):
👍🏾

PR #1924, opened 2026-04-01
[jenkins] fix: switch the default nightly builds to public

Wayonb:
problem: a lot of dependency related failures for JavaScript projects
solution: change the nightly default build to public which will not delete the package-lock file


Jaguar0625 (approved):
👍

PR #1930, opened 2026-04-02
[sdk/python]: python sdk improvements suggested by docs team

Jaguar0625:
if/when approved, similar changes will be made to javascript sdk

### DIFF
--- a/sdk/python/symbolchain/nem/TransactionFactory.py
+++ b/sdk/python/symbolchain/nem/TransactionFactory.py
@@ -66,9 +66,13 @@ class TransactionFactory:
 	def attach_signature(transaction, signature):
 		"""Attaches a signature to a transaction."""
 		transaction.signature = nc.Signature(signature.bytes)
+		return TransactionFactory.to_json(transaction)
 
+	@staticmethod
+	def to_json(transaction):
+		"""Generates a JSON representation of transaction that can be sent to a node."""
 		transaction_hex = hexlify(TransactionFactory.to_non_verifiable_transaction(transaction).serialize()).decode('utf8').upper()
-		signature_hex = str(signature)
+		signature_hex = str(transaction.signature)
 		json_payload = f'{{"data":"{transaction_hex}", "signature":"{signature_hex}"}}'
 		return json_payload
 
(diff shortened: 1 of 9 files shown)


sdk/python/symbolchain/nem/TransactionFactory.py
```diff
@@ -67,8 +67,11 @@ def attach_signature(transaction, signature):
 		"""Attaches a signature to a transaction."""
 		transaction.signature = nc.Signature(signature.bytes)
 
+	@staticmethod
+	def to_json(transaction):
```
zero4862, line 72: @segfaultxavi note for us: this will break existing code examples once we update the SDK, so we’ll need to update them once released.

segfaultxavi: Not only will it break our tutorials. **This breaks EVERYBODY's code out there!**

I don't think we want to do that. I'd rather keep the existing functionality as is, and add the new `to_json` method. It's just a matter of calling `to_json` at the end of `attach_signature`, no?
@Jaguar0625 

We can say that `attach_signature` returns the JSON payload as a convenience, because that's the most common use case.

sdk/python/symbolchain/symbol/TransactionFactory.py
```diff
@@ -70,6 +70,9 @@ def attach_signature(transaction, signature):
 		"""Attaches a signature to a transaction."""
 		transaction.signature = sc.Signature(signature.bytes)
 
+	@staticmethod
```
segfaultxavi, line 74: Same comment as for NEM. This breaks every Symbol app out there.

segfaultxavi (changes requested):
My only concern is with the API break for `attach_signature`.
The rest of changes look OK.

sdk/python/examples/docs/__main__.py
```diff
@@ -111,7 +111,8 @@ async def create_account_with_tokens(facade, amount=500, private_key=None):  # p
 	print(f'seed transfer transaction hash {transaction_hash}')
 
 	# finally, construct the over wire payload
-	json_payload = facade.transaction_factory.attach_signature(transaction, signature)
+	facade.transaction_factory.attach_signature(transaction, signature)
```
segfaultxavi: No need for all these changes if we keep the current API, as I suggest in the other comment.

sdk/python/symbolchain/nem/TransactionFactory.py
segfaultxavi: Fixed with the latest changes.

segfaultxavi (approved):
Functionality looks good to me!

zero4862 (approved):

Wayonb (approved):
👍🏾

PR #1942, opened 2026-04-10
[client/catapult] fix: Amend buffer overflow on threads naming

AndreaLanfranchi:
This amends UB for thread name setter/getter.
MSVC debug builds were trapping this.

### DIFF
--- a/client/catapult/src/catapult/thread/ThreadInfo.cpp
+++ b/client/catapult/src/catapult/thread/ThreadInfo.cpp
@@ -20,11 +20,11 @@
 **/
 
 #include "ThreadInfo.h"
+#include <cstring>
+
 #ifdef _WIN32
 #include <array>
-#endif
-#include <cstring>
-#ifndef _WIN32
+#else
 #include <errno.h>
 #include <pthread.h>
 #endif
--- a/client/catapult/src/catapult/thread/ThreadInfo.cpp
+++ b/client/catapult/src/catapult/thread/ThreadInfo.cpp
@@ -33,17 +33,13 @@ namespace catapult { namespace thread {
 
 		constexpr std::size_t kMaxThreadNameBytes = 64; // Including the null terminator
 
-		constexpr int pthread_self() {
-			return 0;
-		}
-
-		int pthread_setname_np(int, const char* name) {
+		int pthread_setname_np(pthread_t, const char* name) {
 			/*
 			* On macOS, pthread_setname_np only accepts the thread name without the thread ID,
 			* and it applies to the current thread. Therefore, we can ignore the thread ID
 			* and directly set the thread name for the current thread.
 			*/
-			return pthread_setname_np(name);
+			return ::pthread_setname_np(name);
 		}
 
 #elif defined(__GLIBC__)
--- a/client/catapult/src/catapult/thread/ThreadInfo.cpp
+++ b/client/catapult/src/catapult/thread/ThreadInfo.cpp
@@ -116,7 +116,7 @@ namespace catapult { namespace thread {
 		if (0 == pthread_getname_np(pthread_self(), buffer, sizeof(buffer))) {
 			return std::string(buffer, buffer + std::strlen(buffer));
 		}
-        return std::string();
+		return std::string();
 	}
 }
 }
--- a/client/catapult/src/catapult/thread/ThreadInfo.cpp
+++ b/client/catapult/src/catapult/thread/ThreadInfo.cpp
@@ -24,8 +24,8 @@
 #include <array>
 #endif
 #include <cstring>
+#include <tuple>
 #ifndef _WIN32
-#include <errno.h>
 #include <pthread.h>
 #endif
 
--- a/client/catapult/src/catapult/thread/ThreadInfo.cpp
+++ b/client/catapult/src/catapult/thread/ThreadInfo.cpp
@@ -32,27 +32,29 @@
 
 namespace catapult { namespace thread {
 
-	/*
-	* Windows and Apple platforms support thread names up to 64 characters,
-	* while Linux (glibc) supports up to 16 characters
-	* Note this value does not include the NUL-terminator.
-	*/
-#if defined(_WIN32) || defined(__APPLE__)
-    constexpr std::size_t kMaxThreadNameLength = 63;
+	namespace {
+
+#if defined(PTHREAD_MAX_NAMELEN_NP)
+        constexpr std::size_t kMaxThreadNameBytes = PTHREAD_MAX_NAMELEN_NP;
+#elif defined(__APPLE__)
+        constexpr std::size_t kMaxThreadNameBytes = 64; // Including the null terminator
+#elif defined(__linux__) && HAVE_PTHREAD_SETNAME_NP
+        constexpr std::size_t kMaxThreadNameBytes = 16; // Including the null terminator
+#elif defined(_WIN32)
+        constexpr std::size_t kMaxThreadNameBytes = 64; // Including the null terminator (arbitrary choice)
+#define HAVE_PTHREAD_SETNAME_NP 1 // We provide our own implementation of pthread_setname_np on Windows
+#define HAVE_PTHREAD_GETNAME_NP 1 // We provide our own implementation of pthread_getname_np on Windows
+#elif defined(__GLIBC__)
+        constexpr std::size_t kMaxThreadNameBytes = 16; // Including the null terminator
 #else
-    constexpr std::size_t kMaxThreadNameLength = 15;
+        constexpr std::size_t kMaxThreadNameBytes = 0; // Disable thread naming on unsupported platforms
 #endif
 
-	size_t GetMaxThreadNameLength() {
-        return kMaxThreadNameLength;
-	}
-
-	namespace {
 #ifdef _WIN32
 		
-		thread_local std::array<char, kMaxThreadNameLength + /*NUL-terminator*/ 1> t_threadName = {'\0'};
+		thread_local std::array<char, kMaxThreadNameBytes> t_threadName = {'\0'};
 
-		int pthread_self() {
+		constexpr int pthread_self() {
 			return 0;
 		}
 
@@ -72,36 +74,59 @@ namespace catapult { namespace thread {
 				return 1;
 			/*
 			* From GetThreadName we already know `name` is a null terminated array of 
-			* at least kMaxThreadNameLength characters. We can safely copy it the
+			* at least kMaxThreadNameBytes characters. We can safely copy it the
             * whole content of t_threadName, which is also null terminated.
 			*/
-            std::memcpy(name, t_threadName.data(), kMaxThreadNameLength);
+            std::memcpy(name, t_threadName.data(), kMaxThreadNameBytes);
 			return 0;
 		}
-#elif !defined(__APPLE__)
-		// add overload that sets name of current thread
-		int pthread_setname_np(const char* name) {
-			return ::pthread_setname_np(pthread_self(), name);
-		}
+
 #endif
+	}
+
+	std::size_t GetMaxThreadNameLength() {
+		return kMaxThreadNameBytes ? kMaxThreadNameBytes - 1 : 0;
 	}
 
 	void SetThreadName(const std::string& name) {
         if (name.empty())
 			return;
 
-		auto truncatedName = name.substr(0, GetMaxThreadNameLength());
-		std::ignore = pthread_setname_np(truncatedName.c_str());
+		const std::size_t maxLength = GetMaxThreadNameLength();
+		if (0 == maxLength)
+			return;
+
+        /// We truncate from the front of the name to preserve any unique suffixes, 
+		/// which are often more useful for debugging than a common prefix.
+		auto startOffset = name.size() > maxLength ? name.size() - maxLength : 0;
+		auto truncatedName = name.substr(startOffset, maxLength);
+
+#if defined(HAVE_PTHREAD_SET_NAME_NP) && HAVE_PTHREAD_SET_NAME_NP
+		std::ignore = pthread_set_name_np(::pthread_self(), truncatedName.data());
+#elif defined(HAVE_PTHREAD_SETNAME_NP) && HAVE_PTHREAD_SETNAME_NP
+#if defined(__APPLE__) || defined(_WIN32)
+		std::ignore = pthread_setname_np(truncatedName.data());
+#else
+        std::ignore = pthread_setname_np(pthread_self(), truncatedName.c_str());
+#endif
+#endif
 	}
 
 	std::string GetThreadName() {
-#if defined(__APPLE__) || defined(__GLIBC__) || defined(_WIN32)
-        std::string ret(kMaxThreadNameLength, char(0)); // pre-allocate buffer for name retrieval
-		std::ignore = pthread_getname_np(pthread_self(), &ret[0], ret.size());
-		return std::string(ret.c_str());
-#else
-		// musl libc (from alpine) defines __GNU_SOURCE__ but it only has pthread_setname_np
-		return std::string();
+        std::string ret;
+        if (0 == kMaxThreadNameBytes)
+			return ret;
+
+		char buffer[kMaxThreadNameBytes] = { '\0' };
+        ret.reserve(kMaxThreadNameBytes);
+#if defined(HAVE_PTHREAD_GETNAME_NP) && HAVE_PTHREAD_GETNAME_NP
+        if (0 == pthread_getname_np(pthread_self(), buffer, sizeof(buffer))) {
+            ret.append(buffer, buffer + std::strlen(buffer));
+		}
+#elif defined(HAVE_PTHREAD_GET_NAME_NP) && HAVE_PTHREAD_GET_NAME_NP
+		pthread_get_name_np(pthread_self(), buffer, sizeof(buffer));
+		ret.append(buffer, buffer + std::strlen(buffer));
 #endif
+        return ret;
 	}
 }}
--- a/client/catapult/src/catapult/thread/ThreadInfo.cpp
+++ b/client/catapult/src/catapult/thread/ThreadInfo.cpp
@@ -20,9 +20,9 @@
 **/
 
 #include "ThreadInfo.h"
+#include <cstring>
 #ifdef _WIN32
 #include <array>
-#include <cstring>
 #else
 #include <errno.h>
 #include <pthread.h>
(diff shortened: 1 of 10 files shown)


client/catapult/src/catapult/thread/ThreadInfo.cpp
```diff
@@ -20,43 +20,62 @@
...
-			name[len - 1] = '\0';
+			if (!name || 0 == len)
+				return 1;
+			/*
+			* From GetThreadName we already know `name` is a null terminated array of 
+			* at least kMaxThreadNameLength characters. We can safely copy it the
+            * whole content of t_threadName, which is also null terminated.
+			*/
+            std::memcpy(name, t_threadName.data(), kMaxThreadNameLength);
 			return 0;
```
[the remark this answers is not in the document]
AndreaLanfranchi: This is not true. The flow clearly asserts the buffer is pre-allocated with enough space to fit the contents.

client/catapult/src/catapult/thread/ThreadInfo.cpp
```diff
@@ -68,16 +87,18 @@ namespace catapult { namespace thread {
...
+		std::ignore = pthread_setname_np(truncatedName.c_str());
 	}
 
 	std::string GetThreadName() {
 #if defined(__APPLE__) || defined(__GLIBC__) || defined(_WIN32)
-		std::string name;
-		name.resize(GetMaxThreadNameLength() + 1); // include space for NUL-terminator
-		pthread_getname_np(pthread_self(), &name[0], name.size());
-		return name.substr(0, name.find_first_of('\0'));
+        std::string ret(kMaxThreadNameLength, char(0)); // pre-allocate buffer for name retrieval
```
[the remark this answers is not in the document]
AndreaLanfranchi: Addressed in 08cc39c

client/catapult/src/catapult/thread/ThreadInfo.cpp
```diff
@@ -20,67 +20,110 @@
...
+		// which are often more useful for debugging than a common prefix.
+		auto startOffset = name.size() > maxLength ? name.size() - maxLength : 0;
+		auto truncatedName = name.substr(startOffset, maxLength);
+
+#if defined(HAVE_PTHREAD_SET_NAME_NP) && HAVE_PTHREAD_SET_NAME_NP
+		std::ignore = pthread_set_name_np(::pthread_self(), truncatedName.data());
+#elif defined(__APPLE__) || defined(_WIN32)
+		std::ignore = pthread_setname_np(truncatedName.data());
+#else
+		std::ignore = pthread_setname_np(pthread_self(), truncatedName.c_str());
```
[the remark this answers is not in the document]
AndreaLanfranchi: (void) cast is legacy C++. Prefer more idiomatic std::ignore.
@copilot add include <tuple> header as last include

client/catapult/src/catapult/thread/ThreadInfo.cpp
```diff
@@ -20,67 +20,110 @@
...
 
 		int pthread_setname_np(const char* name) {
-			t_threadName = name;
+			/*
+			* We've already truncated the name to fit the maximum length in SetThreadName,
+			* so we can safely copy it here without worrying about truncation.
+			* We automatically include the NUL-terminator in the copy length.
+			*/
+			auto bytesCount = std::strlen(name) + 1;
+			std::memcpy(t_threadName.data(), name, bytesCount);
```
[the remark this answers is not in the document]
AndreaLanfranchi: This is not accurate. The comment in code is self explaining.
Won't fix as not necessary

client/catapult/src/catapult/thread/ThreadInfo.h
```diff
@@ -25,7 +25,7 @@
 namespace catapult { namespace thread {
 
 	/// Gets the maximum supported thread name length (excluding NUL-terminator).
-	size_t GetMaxThreadNameLength();
+	std::size_t GetMaxThreadNameLength();
```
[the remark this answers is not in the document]
AndreaLanfranchi: @copilot apply changes based on [this feedback](https://github.com/symbol/symbol/pull/1942#discussion_r3068237896)

client/catapult/src/catapult/thread/ThreadInfo.cpp
```diff
@@ -20,67 +20,110 @@
...
-#else
-		// musl libc (from alpine) defines __GNU_SOURCE__ but it only has pthread_setname_np
-		return std::string();
+		std::string ret;
+		if (0 == kMaxThreadNameBytes)
+			return ret;
+
+		char buffer[kMaxThreadNameBytes] = { '\0' };
+		ret.reserve(kMaxThreadNameBytes);
+#if defined(__APPLE__) || defined(__linux__) || defined(_WIN32)
```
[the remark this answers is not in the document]
AndreaLanfranchi: @copilot please re-analyze this suggestion considering also the fact in the anon namespace we have set the max number of bytes allowed for name.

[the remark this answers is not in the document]
AndreaLanfranchi: @copilot do the removal of the `#elif defined(__GLIBC__)` branch which is unreachable.

client/catapult/src/catapult/thread/ThreadInfo.cpp
```diff
@@ -20,67 +20,108 @@
...
 	}
 
+	size_t GetMaxThreadNameLength() {
+		return kMaxThreadNameBytes ? kMaxThreadNameBytes - 1 : 0;
+	}
+
 	void SetThreadName(const std::string& name) {
-		auto truncatedName = name.substr(0, GetMaxThreadNameLength());
-		pthread_setname_np(truncatedName.c_str());
+		if (name.empty())
```
Wayonb: Should the thread name(t_threadName) also be set to an empty string, just in case the thread name was set before? 🤔

Wayonb: Plus, add a test to cover this check

client/catapult/src/catapult/thread/ThreadInfo.cpp
```diff
@@ -20,67 +20,108 @@
...
 #endif
-	}
 
-	namespace {
 #ifdef _WIN32
-		// add pthread shims using tls
-		thread_local std::string t_threadName;
 
-		int pthread_self() {
+		thread_local std::array<char, kMaxThreadNameBytes> t_threadName{ '\0' };
```
Wayonb: Is there a reason to use an array that always allocates the max thread name size even when the thread name is shorter?

client/catapult/src/catapult/thread/ThreadInfo.cpp
```diff
@@ -20,67 +20,108 @@
...
+		const size_t maxLength = GetMaxThreadNameLength();
+		if (0 == maxLength)
+			return;
+
+		// We truncate from the front of the name to preserve any unique suffixes,
+		// which are often more useful for debugging than a common prefix.
+		auto startOffset = name.size() > maxLength ? name.size() - maxLength : 0;
+		auto truncatedName = name.substr(startOffset, maxLength);
+
+#if defined(HAVE_PTHREAD_SET_NAME_NP) && HAVE_PTHREAD_SET_NAME_NP
```
Wayonb: The ``pthread_set_name_np`` seems to be for FreeBSD/OpenBSD environments.  Your changes look good, but since we don't test on these environments, we can't tell if this works.

client/catapult/src/catapult/thread/ThreadInfo.cpp
```diff
@@ -20,67 +20,108 @@
...
-#if defined(__APPLE__) || defined(__GLIBC__) || defined(_WIN32)
-		std::string name;
-		name.resize(GetMaxThreadNameLength() + 1); // include space for NUL-terminator
-		pthread_getname_np(pthread_self(), &name[0], name.size());
-		return name.substr(0, name.find_first_of('\0'));
-#else
-		// musl libc (from alpine) defines __GNU_SOURCE__ but it only has pthread_setname_np
-		return std::string();
+		std::string ret;
+		if (0 == kMaxThreadNameBytes)
```
Wayonb: This fails on [Windows build](
https://jenkins.symbolsyndicate.us/job/Symbol/job/catapult-client/job/catapult-client-build-catapult-project/8081/console)
```
ThreadInfo.cpp(110,9): warning C4127: conditional expression is constant
```

Wayonb (changes requested):

client/catapult/src/catapult/thread/ThreadInfo.cpp
AndreaLanfranchi: Actually yes ... TLS is a _special beast_: in our prvious code we were mixing TLS with heap allocation for non-short strings which exposed a fragility on thread destruction.
With this code the storage space is handled by OS specific runtime and associated with the thread. With this code we have fixed-size POD-like storage with no dynamic memory and trivial teardown.

client/catapult/src/catapult/thread/ThreadInfo.cpp
AndreaLanfranchi: > Should the thread name(t_threadName) also be set to an empty string, just in case the thread name was set before? 🤔

Not necessary: we always set and retrieve up to neaerest NUL-terminator.
So ... assuming for visualization our null-terminator is `~` we experience this behavior:

* initial assignement of `abcdef~` : stored value is `abcdef~~~~[...]` and returned value is `abcdef`
* subsequent assignement of `abcd~` : stored value becomes `abcd~f~~~~[...]` and the returned value is `abcd`
* another assignement of `abcdefgh~` : stored value becomes `abcdefgh~~[...]` and the returned value is `abcdefgh`

So there's no need to wipe the whole storage area before (re)assignement.

> Plus, add a test to cover this check

Nice hint which exposes something I didn't tink of: trying to assign a string.empty does not change any previous stored value. So if we assign `abcd` and then eventually an empty string the returned value is still `abcd`.
Would you rather prefer :
* to have the whole name wiped ?
* or to keep current behavior ?

client/catapult/src/catapult/thread/ThreadInfo.cpp
AndreaLanfranchi: I'm removing that.

client/catapult/src/catapult/thread/ThreadInfo.cpp
AndreaLanfranchi: Interesting ... on my VS 2026 this doesn't trigger (both in debug and release). Maybe they have improved the compiler to optimize away.

Would you rather prefer :

* Disable the warning so we don't pullute the code with too many directives or
* Refactor so the condition is erased be preprocessor directives ?

client/catapult/src/catapult/thread/ThreadInfo.cpp
AndreaLanfranchi: Just read some docs and apparently the correct behavior is to make the thread un-named when an empty string is passed.

Wayonb: > to have the whole name wiped ?

I would keep the behavior the same unless we are fixing an issue.  
According to the man page for ``pthread_setname_np``, the only error return is if the name exceeds the maximum length.

client/catapult/src/catapult/thread/ThreadInfo.cpp
Wayonb: > Actually yes ... TLS is a special beast: in our prvious code we were mixing TLS with heap allocation for non-short strings which exposed a fragility on thread destruction.

Hmm...can you explain more?  I thought the ``thread_local std::string x`` means that ``x`` would get destroyed when the thread exited, along with the heap data?  Or are you saying there is a race condition writing to the std::string data during the thread's exit?🤔

AndreaLanfranchi: Yes that's the **lifetime** of x.
But the storage space is quite different. TLS is not heap: is a special region statically initialized by OS when the thread is created. So what we're doing now is simply to take advantage of this area in a static way.
Previously instead the "string" container would be created in the TLS but its data was dynamically allocated on the heap cause there's was no notion about the required size until the first assignement took place.

client/catapult/src/catapult/thread/ThreadInfo.cpp
AndreaLanfranchi: Addressed in [ffd864c](https://github.com/symbol/symbol/pull/1942/commits/ffd864cec616505ab6df439c59fee792c3b0a1ea)

client/catapult/src/catapult/thread/ThreadInfo.cpp
AndreaLanfranchi: Addressed

client/catapult/src/catapult/thread/ThreadInfo.cpp
AndreaLanfranchi: Addressed

client/catapult/src/catapult/thread/ThreadInfo.cpp
Wayonb: Yes, I think both should be fine with C++11

client/catapult/src/catapult/thread/ThreadInfo.cpp
```diff
@@ -20,67 +20,104 @@
...
 
-		int pthread_setname_np(const char* name) {
-			t_threadName = name;
+		int pthread_setname_np(int, const char* name) {
+			/*
+			* On macOS, pthread_setname_np only accepts the thread name without the thread ID,
+			* and it applies to the current thread. Therefore, we can ignore the thread ID
+			* and directly set the thread name for the current thread.
+			*/
+			return pthread_setname_np(name);
```
Wayonb: It seems you are missing ``::pthread_setname_np(name)``
```
Developer/SDKs/MacOSX.sdk -fPIC -MD -MT src/catapult/thread/CMakeFiles/catapult.thread.dir/ThreadInfo.cpp.o -MF src/catapult/thread/CMakeFiles/catapult.thread.dir/ThreadInfo.cpp.o.d -o src/catapult/thread/CMakeFiles/catapult.thread.dir/ThreadInfo.cpp.o -c /Users/wayonblair/work/code/symbol/symbol-complier/client/catapult/src/catapult/thread/ThreadInfo.cpp
/Users/wayonblair/work/code/symbol/symbol-complier/client/catapult/src/catapult/thread/ThreadInfo.cpp:46:11: error: no matching function for call to 'pthread_setname_np'
   46 |                         return pthread_setname_np(name);
```

client/catapult/src/catapult/thread/ThreadInfo.cpp
```diff
@@ -20,67 +20,104 @@
...
 	namespace {
-#ifdef _WIN32
-		// add pthread shims using tls
-		thread_local std::string t_threadName;
+#if defined(__APPLE__)
 
-		int pthread_self() {
+		constexpr std::size_t kMaxThreadNameBytes = 64; // Including the null terminator
+
+		constexpr int pthread_self() {
```
Wayonb: ON apple [pthread_self()](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/pthread_self.3.html) returns pthread_t and not int

```
/Users/wayonblair/work/code/symbol/symbol-complier/client/catapult/src/catapult/thread/ThreadInfo.cpp:116:12: error: no matching function for call to 'pthread_getname_np'
  116 |                 if (0 == pthread_getname_np(pthread_self(), buffer, sizeof(buffer))) {
      |                          ^~~~~~~~~~~~~~~~~~
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/pthread.h:531:5: note: candidate function not viable: no known conversion from 'int' to 'pthread_t _Nonnull' (aka '_opaque_pthread_t *') for 1st argument
  531 | int     pthread_getname_np(pthread_t,char*,size_t);
```

client/catapult/src/catapult/thread/ThreadInfo.cpp
```diff
@@ -20,67 +20,104 @@
...
+		}
+
+		// We truncate from the front of the name to preserve any unique suffixes,
+		// which are often more useful for debugging than a common prefix.
+		auto startOffset = name.size() > maxLength ? name.size() - maxLength : 0;
+		auto truncatedName = name.substr(startOffset);
+		std::ignore = pthread_setname_np(pthread_self(), truncatedName.c_str());
+	}
+
+	void SetThreadName() {
```
Wayonb: This new function is fine, but it is not getting used anywhere

Wayonb (changes requested):

client/catapult/src/catapult/thread/ThreadInfo.cpp
AndreaLanfranchi: Is actually used in `TEST(TEST_CLASS, CanSetSelfThreadName)` to assert the unnaming capability.
Want me to remove it ?

client/catapult/src/catapult/thread/ThreadInfo.cpp
AndreaLanfranchi: Addressed

client/catapult/src/catapult/thread/ThreadInfo.cpp
AndreaLanfranchi: Addressed

client/catapult/src/catapult/thread/ThreadInfo.cpp
Wayonb: If its not used, remove it.
The same can be accomplish by using ``SetThreadName("")``.
if you prefer no argument then create a default ``void SetThreadName(const std::string& name={});``  then you can use this ``SetThreadName()``

AndreaLanfranchi: Don't like defaults but I will comply

Rationale of my dislike: `SetThreadName(std::string name)` can be enforced by assertion string is NOT empty while the overload without arguments semantically clearly states the intention.
However, as said, I will comply removing the overload.

Wayonb: i am ok with just removing the function since it is not used, and if someone wants the functionality can either
1. call SetThreadName("");
2. add the SetThreadName() back

AndreaLanfranchi: Addressed

Wayonb (approved):
👍🏾

PR #1948, opened 2026-04-15
[client/catapult] fix: Restore GCC 11.4 compatibility

AndreaLanfranchi:
A redundant std::move is detected.
Yet gcc 11.4 is default on latest, at this time, Ubuntu LTS


Wayonb (approved):
Looks good👍🏾
My only concern is why this doesn't happen with the newer GCC compilers? @Jaguar0625  thoughts?

PR #1954, opened 2026-04-20
[monorepo]: move .gitattributes file to the root and renormalize all files to LF

Wayonb:
Force LF for all files in the repo


AndreaLanfranchi (approved):
That's ok for me

Jaguar0625 (approved):
👍

PR #1956, opened 2026-04-21
[client/catapult] fix: Enable name return value optimization for MSVC in Debug mode

AndreaLanfranchi:
Added flag to enable name return value optimization in debug mode.
This amends failure of several tests ran in debug mode


Wayonb (approved):
👍🏾

PR #1962, opened 2026-04-28
[mkdocs] add: translate shoestring

daoka:
Translated these contents

* reference/config
* userbook/shoestring


segfaultxavi (approved):
:+1:

Issue #1985, opened 2026-05-11
[client/rest] Get Confirmed Transaction Information endpoint returns 500 when the provided hash is too short


segfaultxavi:
Other endpoints return 409 when provided data is incorrect.

Valid hash:
https://reference.symboltest.net:3001/transactions/confirmed/1F5EB8CD4AEC54C38AB993481B56C03BBBED0DF1E734292B648408DCCD162996

Invalid hash (I removed one character):
https://reference.symboltest.net:3001/transactions/confirmed/1F5EB8CD4AEC54C38AB993481B56C03BBBED0DF1E734292B648408DCCD16299

Nonexisting hashes return 404 as expected.
Non-hex hashes return 409 as expected.

Jaguar0625:
Fixed in https://github.com/symbol/symbol/pull/1986 .

PR #1986, opened 2026-05-11
[client/rest]: get transaction endpoint returns 500 when hash/id is malformed

Jaguar0625:
problem: 500 is returned indicating server error
solution: return 409 instead

   issue: #1985


segfaultxavi (commented):
I am not familiar enough with this codebase to review this PR :/

Wayonb (approved):
👍🏾

Issue #1915, opened 2026-03-25
[client/rest] Error fetching Merkle information for address by public key


cryptoBeliever:
## Summary

Fetching Merkle information for an account using a public key results in an internal server error, while the same request using an address works correctly.

## Steps to reproduce

Send a request using the account public key:

```
GET /accounts/{publicKey}/merkle
```

Example:
```
https://symbol.cryptobeliever.net:3001/accounts/E85D10BF47FFBCE2230F70CB43ED2DDE04FCF342524B383972F86EA1FF773C79/merkle
```

Fetching by address works correctly. Example: 
https://symbol.cryptobeliever.net:3001/accounts/NCHEST3QRQS4JZGOO64TH7NFJ2A63YA7TPM5PXI/merkle

## Expected Behavior

The API should return Merkle information for the account when queried by public key (same as when using address).

## Actual Behavior

The API returns an internal error:

```
{
  "code": "Internal",
  "message": "Cannot read properties of undefined (reading 'publicKeyToAddress')"
}
```

Jaguar0625:
Fixed in https://github.com/symbol/symbol/pull/1987.

PR #1987, opened 2026-05-12
[catapult/rest]: cannot fetch Merkle state information for an account by public key

Jaguar0625:
problem: lookup only works for address
solution: allow public key to be used for lookup

  issue: #1915


Wayonb (approved):
👍🏾

PR #1991, opened 2026-05-14
[jenkins] fix: update node and OS versions used in CI

Wayonb:
problem: CI is currently using an outdated NodeJS version
solution: update NodeJS version plus some OS


Jaguar0625 (approved):
👍

PR #2005, opened 2026-05-27
[linters]: relax gitlint to allow PR number

Wayonb:
problem: GitHub defaults to adding a PR number.
         This causes GitLint to fail in the dev branch.
solution: relax regix used by gitlint to include the PR number as an optional value


AndreaLanfranchi (approved):
LGTM

Jaguar0625 (approved):
👍

PR #2014, opened 2026-06-01
[mkdocs] docs: add error handling to query balance tutorial

zero4862:
- Adds error handling to query balance tutorial (`try / catch`)
- Indents `Account holds N mosaics` code under `else` in Python to match JS implementation.


segfaultxavi (approved):
👍

PR #2022, opened 2026-06-08
[mkdocs] fix: refine Japanese Symbol documentation

daoka:
This update reflects the Japanese translation of the Symbol documentation that was updated after the end of April.


segfaultxavi (approved):
Thanks! Good work keeping the English and Japnese versions in sync!

There shouldn't be much more changes now :)

PR #2025, opened 2026-06-09
[sdk/javascript]: add helper functions for calculating NEM fees

### LINKED ISSUE #2024: [sdk] Recommended-fee helper for NEM transactions
## Summary

It would be helpful if the SDK offered a built-in way to get the recommended fee for a NEM transaction, rather than each developer reimplementing it.

## Motivation

While creating NEM transactions for the docs, I've found it error-prone and repetitive to implement a fee-calculation function that changes depending on the transaction type and whatever is being transferred.

The only way to know what fee to use is to read the docs or the source code and then recreate the calculation yourself, since these values are defined at the protocol level. That means every consumer ends up reimplementing the same logic.

## Proposal

- [ ] Add some kind of `calculateRecommendedFee` helper to the SDK to cover this (name and shape to be defined).
- [ ] Expose the additional namespace and mosaic creation fees as constants.

Jaguar0625:
problem: NEM fee calculation rules are complicated
solution: add helper functions to the sdk

---

once approved will add similar in python sdk

### DIFF
--- a/sdk/javascript/src/nem/FeeCalculator.js
+++ b/sdk/javascript/src/nem/FeeCalculator.js
@@ -1,0 +1,101 @@
+import * as nc from './models.js';
+
+/**
+ * Calculates the minimum required mosaic rental fee.
+ * @returns {bigint} Rental fee.
+ */
+const calculateMosaicRentalFee = () => 10n * 1_000_000n;
+
+/**
+ * Calculates the minimum required namespace rental fee.
+ * @param {boolean} isRoot \c true if the fee should be calculated for a root namespace.
+ * @returns {bigint} Rental fee.
+ */
+const calculateNamespaceRentalFee = isRoot => ((isRoot ? 100n : 10n) * 1_000_000n);
+
+const decodeMosaicId = mosaicId => {
+	const decoder = new TextDecoder();
+	return { namespaceId: { name: decoder.decode(mosaicId.namespaceId.name) }, name: decoder.decode(mosaicId.name) };
+};
+
+const calculateUnweightedTransferFee = (transaction, mosaicInformationLookup) => {
+	const XEM_SUPPLY = 8_999_999_999n;
+	const MAX_MOSAIC_UNITS = 9_000_000_000_000_000n;
+
+	// Math.min and Math.max don't work with bigint
+	const min = (lhs, rhs) => (lhs < rhs ? lhs : rhs);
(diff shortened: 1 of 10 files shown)


Wayonb (approved):
Looks good 👍🏾 

Should we have similar functions on the Symbol side?(to keep parity)

sdk/javascript/src/nem/index.js
```diff
@@ -75,6 +76,27 @@ export {
...
+	 * Calculates the minimum required namespace rental fee.
+	 * @param {boolean} isRoot \c true if the fee should be calculated for a root namespace.
+	 * @returns {bigint} Rental fee.
+	 */
+	calculateNamespaceRentalFee,
+
+	/**
+	 * Calculates the minimum required transaction fee for a transaction.
+	 * @param {models.Transaction} transaction Transaction.
+	 * @param {Function|undefined} mosaicInformationLookup Looks up mosaic information ({supply, divisibility}) given mosaic identifier.
```
segfaultxavi: I fear passing a function as a parameter will make it a bit cumbersome for users.
Could it accept an object, so users can directly feed here what they get from the `namespace/mosaic/definition/page` REST endpoint, for example?

sdk/javascript/src/nem/FeeCalculator.js
```diff
@@ -0,0 +1,89 @@
+import * as nc from './models.js';
+
+/**
+ * Calculates the minimum required mosaic rental fee.
+ * @returns {bigint} Rental fee.
+ */
+const calculateMosaicRentalFee = () => 10n * 1_000_000n;
```
segfaultxavi, line 7: The name of these two methods is a bit misleading, since they do not calculate anything.
Can they be called `get * RentalFee` instead?

sdk/javascript/src/nem/index.js
zero4862: +1, an object feels easier to define (unless there is an upside for using the function).

Worth flagging that the `/namespace/mosaic/definition/page` response by itself isn't enough to compute the fee. It carries the definition, but the current supply comes from a separate endpoint, `/mosaic/supply?mosaicId=...`. Users will need to do a small assembly step, so we can't take the page response unchanged.

zero4862 (commented):
Looks good, one comment: https://github.com/symbol/symbol/pull/2025#discussion_r3387059984

zero4862 (approved):

segfaultxavi (approved):
👍

PR #2026, opened 2026-06-09
[linters]: fail lint_indents if rg is not installed

Jaguar0625:
problem: linter silently succeeds when rg is not installed
solution: explicitly check and fail if not installed


Wayonb (approved):
👍🏾

PR #2028, opened 2026-06-10
[sdk/python]: add helper functions for calculating NEM and Symbol fees

### LINKED ISSUE #2024: [sdk] Recommended-fee helper for NEM transactions

Jaguar0625:
problem: NEM fee calculation rules are complicated
solution: add helper functions to the sdk

---

Notes
1. this complements javascript changes https://github.com/symbol/symbol/pull/2025
2. auto string conversion is a bit more complicated in python, so i will make a followup PR for that

### DIFF
--- a/sdk/python/symbolchain/nem/FeeCalculator.py
+++ b/sdk/python/symbolchain/nem/FeeCalculator.py
@@ -1,0 +1,94 @@
+import math
+
+from symbolchain.nc import TransactionType
+
+
+def calculate_mosaic_rental_fee():
+	"""Calculates the minimum required mosaic rental fee."""
+
+	return 10 * 1_000_000
+
+
+def calculate_namespace_rental_fee(is_root):
+	"""
+	Calculates the minimum required namespace rental fee.
+	When is_root is True, calculates the root namespace fee.
+	Otherwise, calculates the child namespace fee.
+	"""
+
+	return (100 if is_root else 10) * 1_000_000
+
+
+def _decode_mosaic_id(mosaic_id):
+	return {'namespace_id': {'name': mosaic_id.namespace_id.name.decode('utf8')}, 'name': mosaic_id.name.decode('utf8')}
+
+
+def _calculate_unweighted_transfer_fee(transaction, mosaic_information_lookup):
+	xem_supply = 8_999_999_999
+	max_mosaic_units = 9_000_000_000_000_000
+
+	def _calculate_xem_transfer_fee(amount):
+		return int(min(25, max(1, amount // 10000)))
+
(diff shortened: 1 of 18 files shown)


sdk/python/symbolchain/nem/FeeCalculator.py
```diff
@@ -0,0 +1,94 @@
...
+		return amount * mosaic_amount * xem_supply / _calculate_mosaic_total_quantity(mosaic_information)
+
+	def _calculate_mosaic_transfer_fee(amount, mosaic, mosaic_information):
+		if 0 == mosaic_information['divisibility'] and 10_000 >= mosaic_information['supply']:
+			return 1
+
+		xem_equivalent = _calculate_xem_equivalent(amount, mosaic.mosaic.amount.value, mosaic_information)
+		xem_fee = _calculate_xem_transfer_fee(xem_equivalent)
+		mosaic_total_quantity = _calculate_mosaic_total_quantity(mosaic_information)
+		supply_related_adjustment = 0 if 0 == mosaic_total_quantity else 0.8 * math.log(max_mosaic_units / mosaic_total_quantity) // 1
```
zero4862:
```suggestion
		supply_related_adjustment = 0 if 0 == mosaic_total_quantity else int(0.8 * math.log(max_mosaic_units / mosaic_total_quantity))
```

The `else` branch returns a `float` (`// 1` floors the value but keeps the type).

For example, `fee = calculate_transaction_fee(...)` returns `350000.0` for a mosaic transfer. If I set `transaction.fee = Amount(fee)`, signing the transaction fails with `AttributeError: 'float' object has no attribute 'to_bytes'`, unless I cast it with `Amount(int(fee))`.

zero4862 (changes requested):

zero4862 (approved):

segfaultxavi (approved):
👍

Wayonb (approved):
👍🏾

PR #2048, opened 2026-06-18
[skd/java]: add base types for Java SDK

Wayonb:
1. Updated the utils to use Java API where possible to reduce code.  e.g. deepCompare is dropped since Arrays class have similar methods.
2. add ByteArray, BaseValue and other types needed by the generator.


Jaguar0625:
General Project Structure Comments:
* Presumably gradle/wrapper is needed for all java projects, so should it be in a more common area like linters or jenkins?
* afaikt, there aren't any python files, so lint_python (uncalled) and setup_lint seem unneeded?

sdk/java/src/test/java/org/symbol/sdk/impl/Ed25519Test.java
```diff
@@ -0,0 +1,188 @@
+package org.symbol.sdk.impl;
```
Jaguar0625: are you planning to leave these or migrate once the nem/symbol implementations are added? in the other sdks, we're only testing the crypto within nem/symbol and not at this level

sdk/java/src/test/java/org/symbol/sdk/impl/Ed25519Test.java
```diff
@@ -0,0 +1,188 @@
...
+
+		// Act:
+		final byte[] msg = Converter.hexToUint8("72");
+		final byte[] sig = Ed25519.sign(Tweetnacl.HashMode.SHA2_512, msg, kp);
+
+		// Assert:
+		final String expectedSignature = "92A009A9F0D4CAB8720E820B5F642540A2B27B5416503F8FB3762223EBDB69DA"
+				+ "085AC1E43E15996E458F3613D0F11D8C387B2EAEB4302AEEB00D291612BB0C00";
+		assertThat(Converter.uint8ToHex(sig), equalTo(expectedSignature));
+		assertThat(Ed25519.verify(Tweetnacl.HashMode.SHA2_512, msg, sig, kp.publicKey), is(true));
```
Jaguar0625: is there reason previous has expectedPublicKey var (instead of inline) and does not verify (like this one)?

sdk/java/src/test/java/org/symbol/sdk/impl/Ed25519Test.java
```diff
@@ -0,0 +1,188 @@
...
+		final byte[] sigKeccak = Ed25519.sign(Tweetnacl.HashMode.KECCAK_512, msg, kpKeccak);
+
+		final Tweetnacl.KeyPair kpSha = Ed25519.keyPairFromSeed(Tweetnacl.HashMode.SHA2_512, seed);
+
+		// Act + Assert:
+		assertThat(Ed25519.verify(Tweetnacl.HashMode.SHA2_512, msg, sigKeccak, kpSha.publicKey), is(false));
+	}
+
+	@Test
+	void rejectsNonCanonicalS() {
```
Jaguar0625: i am not sure what you're doing with this test? doesn't it invalidate the signature? 

the following test seems to be the non-canonical test?

sdk/java/src/test/java/org/symbol/sdk/impl/SharedKeyHelpersTest.java
```diff
@@ -0,0 +1,160 @@
...
+	void symbol_deriveSharedKeyMatchesOfficialVectors() {
+		// Arrange:
+		final SharedKeyHelpers.SharedKeyDeriver derive = SharedKeyHelpers.deriveSharedKeyFactory("catapult", Tweetnacl.HashMode.SHA2_512);
+
+		// Act + Assert:
+		for (DeriveVector v : SYMBOL_VECTORS) {
+			final byte[] privateKey = Converter.hexToUint8(v.privateKeyHex);
+			final CryptoTypes.PublicKey otherPublicKey = new CryptoTypes.PublicKey(v.otherPublicKeyHex);
+			final CryptoTypes.SharedKey256 result = derive.derive(privateKey, otherPublicKey);
+			assertThat(Converter.uint8ToHex(result.bytes()), equalTo(v.sharedKeyHex));
```
Jaguar0625: technically this line is assert and the above is act

sdk/java/src/test/java/org/symbol/sdk/impl/SharedKeyHelpersTest.java
```diff
@@ -0,0 +1,160 @@
+package org.symbol.sdk.impl;
```
Jaguar0625: similar question as to whether these tests are temporary or not?

sdk/java/README.md
```diff
@@ -0,0 +1,217 @@
...
+| Primitive          | Implementation                                                                  |
+| ------------------ | ------------------------------------------------------------------------------- |
+| SHA-256, SHA-512   | `java.security.MessageDigest`                                                   |
+| HMAC-SHA512        | `javax.crypto.Mac`                                                              |
+| AES-CBC, AES-GCM   | `javax.crypto.Cipher`                                                           |
+| Ed25519 (SHA-512)  | `java.security.Signature` ("Ed25519")                                           |
+| Keccak-256/-512    | Bouncy Castle (`org.bouncycastle.crypto.digests.KeccakDigest`)                  |
+| RIPEMD-160         | Bouncy Castle (`org.bouncycastle.crypto.digests.RIPEMD160Digest`)               |
+| HKDF-SHA256        | Bouncy Castle (`org.bouncycastle.crypto.generators.HKDFBytesGenerator`)         |
+| Ed25519 (Keccak)   | Bouncy Castle low-level `Ed25519` with hasher swap; small ported helpers        |
```
Jaguar0625: isn't ed25519 coming from the tweet-nacl port?

sdk/java/src/main/java/org/symbol/sdk/utils/BufferView.java
```diff
@@ -1,102 +1,166 @@
...
-	 * Gets the underlying buffer.
+	 * Returns a new limited view.
 	 *
-	 * @return the byte buffer.
+	 * @param size Length in bytes.
+	 * @return View limited to specified size.
 	 */
-	public ByteBuffer getBuffer() {
-		return this.bufferViewImpl;
+	public byte[] window(final int size) {
```
Jaguar0625: might want to rename as window generally conotates copy-less operation

sdk/java/src/main/java/org/symbol/sdk/BaseValue.java
```diff
@@ -0,0 +1,187 @@
...
+ * Fixed-size integer value (amount, height, mosaic id, …) extended by every generated integer POD. F-bounded so {@link Comparable} is
+ * type-safe per POD. The base class owns the value semantics (equals/hashCode/toString/serialize/compareTo) that an interface cannot
+ * provide for {@link Object} methods.
+ *
+ * @param <T> Self-type — the concrete POD extending this class.
+ */
+public abstract class BaseValue<T extends BaseValue<T>> implements Serializer, Comparable<T> {
+	private final BigInteger value;
+
+	private final int size;
```
Jaguar0625, line 18: any reason not to use isSigned?

Wayonb:
> General Project Structure Comments:
> 
> * Presumably gradle/wrapper is needed for all java projects, so should it be in a more common area like linters or jenkins?

Yes, this is the first Gradle project.  I guess we are planning to switch NEM from mvn?  Jenkins makes more sense as a shared location so I will take a look.

> * afaikt, there aren't any python files, so lint_python (uncalled) and setup_lint seem unneeded?

yes, currently not needed, but I will just add it back in the next PR 😅

sdk/java/src/main/java/org/symbol/sdk/Bip32.java
```diff
@@ -0,0 +1,335 @@
...
+	}
+
+	/** BIP-39 mnemonic ↔ seed and ↔ entropy operations. */
+	private static final class Mnemonic {
+		private Mnemonic() {
+		}
+
+		static byte[] toSeed(final String normalizedMnemonic, final String password) {
+			final String normalizedPassword = Normalizer.normalize(null == password ? "" : password, Normalizer.Form.NFKD);
+			final char[] passwordChars = normalizedMnemonic.toCharArray();
```
Jaguar0625: mnemonicChars?

sdk/java/src/main/java/org/symbol/sdk/Bip32.java
```diff
@@ -0,0 +1,335 @@
...
+		try {
+			final Mac mac = Mac.getInstance("HmacSHA512");
+			mac.init(new SecretKeySpec(key, "HmacSHA512"));
+			return mac.doFinal(data);
+		} catch (java.security.GeneralSecurityException ex) {
+			throw new CryptoException(ex);
+		}
+	}
+
+	/** BIP-39 mnemonic ↔ seed and ↔ entropy operations. */
```
Jaguar0625: you didn't want to use an existing third party implementation for this?

sdk/java/src/main/java/org/symbol/sdk/CatbufferType.java
```diff
@@ -0,0 +1,226 @@
...
+
+import org.symbol.sdk.utils.Writer;
+
+/**
+ * Shared base for every generated catbuffer struct (PODs and enums do not extend it): {@link #serialize()} allocates a {@link Writer} sized
+ * by {@link #size()} and threads it through {@link #serializeInto(Writer)}.
+ */
+public abstract class CatbufferType implements Serializer {
+	/**
+	 * * Calculates the size of the object when serialized. Every generated struct overrides.
```
Jaguar0625: double *

Jaguar0625: should this be abstract instead of having default impl?

sdk/java/src/main/java/org/symbol/sdk/CatbufferType.java
```diff
@@ -0,0 +1,226 @@
...
+	protected abstract void serializeInto(Writer buffer);
+
+	/**
+	 * Returns the JSON-serializable projection of this object (nested {@code Map}/{@code List} tree with string-encoded u64 values). Every
+	 * generated struct overrides; the projection is a wire snapshot, not a descriptor — byte fields render as hex while descriptor input
+	 * reads strings as UTF-8 text.
+	 *
+	 * @return JSON-serializable representation of this object.
+	 */
+	public Object toJson() {
```
Jaguar0625: abstract?

sdk/java/src/main/java/org/symbol/sdk/CatbufferType.java
```diff
@@ -0,0 +1,226 @@
...
+
+	/**
+	 * Reads a named field. Generated subclasses override with a {@code switch} chaining to {@code super.getField(name)}.
+	 *
+	 * @param name Field name.
+	 * @return Field value (callers cast to the appropriate type).
+	 */
+	public Object getField(final String name) {
+		throw new InvalidDescriptorException(String.format("no field \"%s\" on %s", name, getClass().getSimpleName()));
+	}
```
Jaguar0625, line 96: should these be abstract?

sdk/java/src/main/java/org/symbol/sdk/Bip32.java
```diff
@@ -0,0 +1,335 @@
...
+				throw new CryptoException(ex);
+			}
+			final byte[] hash = sha256.digest(entropy);
+			for (int i = 0; i < cs; ++i) {
+				final int expected = (hash[i / 8] >> (7 - (i % 8))) & 1;
+				final int actual = bits.charAt(ent + i) - '0';
+				if (expected != actual)
+					throw new IllegalArgumentException("Mnemonic string is invalid: checksum mismatch");
+			}
+		}
```
Jaguar0625: consider refactoring the calculation of the entropy checksum into a separate function, so you can just call it again here and do an array compare

sdk/java/src/test/java/org/symbol/sdk/utils/ArrayHelpersTest.java
```diff
@@ -1,675 +1,602 @@
...
-		assertAlignUp(new int[]{
-				9, 16
-		}, 8, 16);
-		assertAlignUp(new int[]{
-				257, 264
-		}, 8, 264);
-	}
+		@Test
+		void alwaysAlignsUp() {
+			// Act + Assert:
```
Jaguar0625: don't need to update, but you can skip "Act + Assert" when there are only asserts in test function

sdk/java/src/test/java/org/symbol/sdk/utils/ArrayHelpersTest.java
```diff
@@ -1,675 +1,602 @@
...
 
-	@Test
-	public void canReadArrayWhenUsingAccessorAndElementAreOrdered() {
-		// Arrange:
-		final ReadTestContext context = new ReadTestContext(readArraySizes);
+	@Nested
+	final class ReadVariableSizeElements {
+		@Test
+		void throwsWhenAnyElementHasZeroSize() {
+			// Arrange:
```
Jaguar0625: can't you call assertThrowsWhenAnyElementHasZeroSize ?

sdk/java/src/test/java/org/symbol/sdk/utils/ArrayHelpersTest.java
```diff
@@ -1,675 +1,602 @@
...
 
-	final List<String> writeArrayExpectedWrites = new ArrayList<>() {
-		{
-			add("type: value value: 101");
-			add("type: value value: 104");
-			add("type: value value: 107");
-			add("type: value value: 110");
-			add("type: value value: 113");
+			// Act:
+			ArrayHelpers.writeArrayCount(output, java.util.List.of(new MockElement(1), new MockElement(2), new MockElement(1)), 2, bySize);
```
Jaguar0625: should assert what is written

sdk/java/src/test/java/org/symbol/sdk/utils/ArrayHelpersTest.java
```diff
@@ -1,675 +1,602 @@
...
+	// region sorted-order validation
 
-	@Test
-	public void cannotReadAtBufferEndWhenLastReadResultsInOOB() {
-		// Arrange:
-		final ReadTestContext context = new ReadTestContext(new int[]{
-				23, 25
-		}, 49);
+	@Nested
+	final class OrderValidation {
```
Jaguar0625: might want to move these tests so all the write tests are next to each other (this includes both read and write tests)

sdk/java/src/test/java/org/symbol/sdk/utils/Base32Test.java
```diff
@@ -1,190 +1,103 @@
...
 
+	@Nested
+	final class Encode {
+		@Test
+		void canConvertEmptyInput() {
 			// Act + Assert:
-			final String message = String.format("decoded size must be multiple of 5");
-			IllegalArgumentException thrown = assertThrows(IllegalArgumentException.class, () -> Base32.encode(input));
-			assertThat(thrown.getMessage(), equalTo(message));
+			assertThat(Base32.encode(new byte[0]), equalTo(""));
```
Jaguar0625: would prefer if you split up the act and assert on separate lines, like the other sdks

sdk/java/src/test/java/org/symbol/sdk/utils/Base32Test.java
```diff
@@ -1,190 +1,103 @@
...
+			final byte[] data = new byte[260];
+			for (int i = 0; i < 260; ++i)
+				data[i] = (byte) (i & 0xFF);
 
-	@Test
-	public void decodeThrowsWhenInputSizeIsIncorrect() {
-		// Arrange:
-		for (int i = 1; 8 > i; ++i) {
-			final String input = "A".repeat(i);
+			final String expected = "" + "AAAQEAYEAUDAOCAJBIFQYDIOB4IBCEQTCQKRMFYY" + "DENBWHA5DYPSAIJCEMSCKJRHFAUSUKZMFUXC6MBR"
```
Jaguar0625: probably can drop ""

sdk/java/src/test/java/org/symbol/sdk/utils/Base32Test.java
```diff
@@ -1,190 +1,103 @@
...
+				if (0 == i % 8)
+					continue;
 
-			// Assert:
-			assertThat(input, equalTo(result));
-		});
-	}
+				final int size = i;
+				final IllegalArgumentException ex = assertThrows(IllegalArgumentException.class, () -> Base32.decode("A".repeat(size)));
+				assertThat(ex.getMessage(), containsString("encoded size must be multiple of 8"));
```
Jaguar0625: might want to include message with index

sdk/java/src/test/java/org/symbol/sdk/utils/Base32Test.java
```diff
@@ -1,190 +1,103 @@
...
 
 			// Assert:
-			assertThat(input, equalTo(Converter.bytesToHex(result)));
-		});
+			assertThat(ex.getMessage(), containsString("illegal base32 character"));
+		}
 	}
-
-	// endregion
 }
```
Jaguar0625, line 167: add roundtrip test(s)?

sdk/java/src/test/java/org/symbol/sdk/utils/BufferViewTest.java
```diff
@@ -4,260 +4,676 @@
...
+		@Test
+		void backingReturnsUnderlyingArray() {
+			// Arrange:
+			final byte[] bytes = {
+					0, 1, 2, 3, 4, 5, 6
+			};
+			final BufferView view = new BufferView(bytes);
+
+			// Act + Assert:
+			assertThat(view.backing() == bytes, equalTo(true));
```
Jaguar0625: use sameInstance matcher?

sdk/java/src/test/java/org/symbol/sdk/utils/BufferViewTest.java
```diff
@@ -4,260 +4,676 @@
...
-		final TestContext context = new TestContext();
-		context.view.shiftRight(4);
-
-		// Act + Assert:
-		final IllegalArgumentException exception = assertThrows(IllegalArgumentException.class, () -> context.view.shiftRight(2));
-		assertThat(exception.getMessage(), equalTo("size cannot be greater than the remaining bytes"));
+	// region Accessors
+
+	@Nested
+	final class Accessors {
```
Jaguar0625: i think it would be better to test accessors after each modification test, as i think that might be a bit more complete suite

sdk/java/src/test/java/org/symbol/sdk/utils/BufferViewTest.java
```diff
@@ -4,260 +4,676 @@
...
+
+	@Nested
+	final class ShiftRight {
+		@Test
+		void canShiftBy0() {
+			// Arrange:
+			final BufferView view = newSubView();
+
+			// Act:
+			view.shiftRight(0);
```
Jaguar0625: bit unusual. 1 << 0 typically is 1. 🤔

sdk/java/src/test/java/org/symbol/sdk/utils/BufferViewTest.java
```diff
@@ -4,260 +4,676 @@
...
+			final BufferView view = newSubView();
+
+			// Act + Assert:
+			assertThat(view.window(3), equalTo(new byte[]{
+					2, 3, 4
+			}));
+		}
+
+		@Test
+		void canCreateNonShrinkingSubview() {
```
Jaguar0625: do you rely on this behavior anywhere? 
this is inconsistent with how Uint8Array.subarray works in js (cannot reach outside bounds of subview)

sdk/java/src/test/java/org/symbol/sdk/utils/BufferViewTest.java
```diff
@@ -4,260 +4,676 @@
...
+			final BufferView view = new BufferView(new byte[]{
+					0x01, 0x02, 0x03
+			});
+
+			// Act + Assert:
+			assertThrows(IllegalArgumentException.class, () -> view.peekInt(3, false));
+		}
+
+		@Test
+		void cannotPeek5ByteInt() {
```
Jaguar0625, line 538: consider refactor with previous and add cannot peek 8 variant

sdk/java/src/test/java/org/symbol/sdk/utils/BufferViewTest.java
```diff
@@ -4,260 +4,676 @@
...
-
-		// Act + Assert:
-		final IllegalArgumentException exception = assertThrows(IllegalArgumentException.class, () -> context.view.window(6));
-		assertThat(exception.getMessage(), equalTo("size cannot be greater than the remaining bytes"));
-	}
+	// endregion
 
 	@Test
-	public void cannotShrinkNegativeSubview() {
+	void negativeSizesAreRejected() {
```
Jaguar0625: this seems redundant to other tests? e.g. cannotShrinkNegative

sdk/java/src/test/java/org/symbol/sdk/utils/BufferViewTest.java
```diff
@@ -4,260 +4,676 @@
...
+			final BufferView view = new BufferView(new byte[]{
+					0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08
+			});
+
+			// Act + Assert:
+			assertThrows(IllegalArgumentException.class, () -> view.peekBigInt(0, false));
+		}
+
+		@Test
+		void unsignedNegativeLongBecomesPositiveBigInteger() {
```
Jaguar0625: i think test name is misleading? aren't you testing that unsigned value > Long.MAX_VALUE can be read?

sdk/java/src/test/java/org/symbol/sdk/utils/ConverterTest.java
```diff
@@ -1,220 +1,590 @@
...
+		@Test
+		void returnsTrueForValidHexStrings() {
+			// Arrange:
+			final String[] inputs = {
+					"", "026ee415fc15", "abcdef0123456789ABCDEF"
+			};
+
+			// Act + Assert:
+			for (String input : inputs)
+				assertThat("input " + input, Converter.isHexString(input), is(true));
```
Jaguar0625: also think it would be cleaner to split act and assert lines (using temp var)

sdk/java/src/test/java/org/symbol/sdk/utils/ConverterTest.java
```diff
@@ -1,220 +1,590 @@
...
+			final byte[] bytes = {
+					(byte) 0xFF, (byte) 0xFF
+			};
+
+			// Act + Assert:
+			assertThat(Converter.bytesToInt(bytes, 2, true), equalTo(-1L));
+		}
+
+		@Test
+		void readsInt16Positive() {
```
Jaguar0625: i think test name is misleading since expected is negative?

sdk/java/src/test/java/org/symbol/sdk/utils/ConverterTest.java
```diff
@@ -1,220 +1,590 @@
...
+					0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x00
+			};
 
-	// endregion
+			// Act:
+			final BigInteger value = Converter.bytesToBigInt(bytes, 8, true);
 
-	// region bytesToHex
+			// Assert:
+			assertThat(value.signum(), equalTo(1)); // positive
```
Jaguar0625: can you explain purpose of this test? how is it different from readsUint64?

sdk/java/src/main/java/org/symbol/sdk/utils/Converter.java
```diff
@@ -1,90 +1,213 @@
...
+
+	/**
+	 * Converts bytes to a little-endian {@link BigInteger}.
+	 *
+	 * @param input Byte array.
+	 * @param size Number of bytes (must be 8).
+	 * @param isSigned {@code true} if number should be treated as signed.
+	 * @return Value corresponding to the input.
+	 */
+	public static BigInteger bytesToBigInt(final byte[] input, final int size, final boolean isSigned) {
```
Jaguar0625: why do you need BigInteger at all? shouldn't built-in Long be sufficient?

sdk/java/src/test/java/org/symbol/sdk/utils/WriterTest.java
```diff
@@ -0,0 +1,434 @@
...
+			// Arrange:
+			final Writer writer = new Writer(10);
+
+			// Act:
+			writer.writeBigInt(BigInteger.valueOf(0x123456789ABCDEFL), 8);
+
+			// Assert:
+			assertThat(writer.offset, equalTo(8));
+			// Little-endian: 0xEF 0xCD 0xAB 0x89 0x67 0x45 0x23 0x01
+			assertThat(writer.storage[0], equalTo((byte) 0xEF));
```
Jaguar0625: any reason you don't want to compare contents of storage vs each element individually ?

sdk/java/src/test/java/org/symbol/sdk/BaseValueTest.java
```diff
@@ -0,0 +1,499 @@
...
+			assertOutOfRange(() -> new Bv(BigInteger.valueOf(0x100000000L), 4, false), "32-bit");
+
+			assertOutOfRange(() -> new Bv(BigInteger.valueOf(-1), 8, false), "64-bit");
+			assertOutOfRange(() -> new Bv(new BigInteger("10000000000000000", 16), 8, false), "64-bit");
+		}
+
+		@Test
+		void canCreateSignedBaseValueAcrossSizes() {
+			// Act + Assert:
+			canCreateSigned(BigInteger.valueOf(-0x80), 1, BigInteger.valueOf(-0x80));
```
Jaguar0625: consider also signed values inbetween

sdk/java/src/test/java/org/symbol/sdk/BaseValueTest.java
```diff
@@ -0,0 +1,499 @@
...
+			canCreateSigned(min64, 8, min64);
+			canCreateSigned(max64, 8, max64);
+		}
+
+		@Test
+		void cannotCreateSignedWithValuesOutsideRange() {
+			// Act + Assert:
+			assertOutOfRange(() -> new Bv(BigInteger.valueOf(-0x81), 1, true), "8-bit");
+			assertOutOfRange(() -> new Bv(BigInteger.valueOf(0x80), 1, true), "8-bit");
+
```
Jaguar0625, line 192: check bounds of other sizes too, like in unsigned test?

sdk/java/src/test/java/org/symbol/sdk/BaseValueTest.java
```diff
@@ -0,0 +1,499 @@
...
+			assertOutOfRange(() -> new Bv(BigInteger.ONE.shiftLeft(63), 8, true), "64-bit");
+		}
+	}
+
+	// endregion
+
+	// region Accessors
+
+	@Nested
+	final class Accessors {
```
Jaguar0625, line 215: i'm not sure if this adds anything relative to canCreateUnsigned / canCreateSigned (and you can check signed in those too)

sdk/java/src/test/java/org/symbol/sdk/BaseValueTest.java
```diff
@@ -0,0 +1,499 @@
...
+			assertThat(v.isSigned(), equalTo(true));
+		}
+	}
+
+	// endregion
+
+	// region Serialization
+
+	@Nested
+	final class Serialization {
```
Jaguar0625, line 257: remind me reason for having serialize() on BaseValue vs calling it externally?

sdk/java/src/test/java/org/symbol/sdk/BaseValueTest.java
```diff
@@ -0,0 +1,499 @@
...
+			// Act + Assert:
+			assertThat(new Bv(BigInteger.ZERO, 1, false).toString(), equalTo("0x00"));
+			assertThat(new Bv(BigInteger.valueOf(0xFF), 1, false).toString(), equalTo("0xFF"));
+			assertThat(new Bv(BigInteger.valueOf(0x12345678L), 4, false).toString(), equalTo("0x12345678"));
+			assertThat(new Bv(new BigInteger("1234567890ABCDEF", 16), 8, false).toString(), equalTo("0x1234567890ABCDEF"));
+			assertThat(new Bv(new BigInteger("FFFFFFFFFFFFFFFF", 16), 8, false).toString(), equalTo("0xFFFFFFFFFFFFFFFF"));
+		}
+
+		@Test
+		void toStringOfSignedBaseValuesOutputsFixedWidthHex() {
```
Jaguar0625: based on test name, this should test that leading zeros are present but none of the tests seem to test that ?

sdk/java/src/test/java/org/symbol/sdk/BaseValueTest.java
```diff
@@ -0,0 +1,499 @@
...
+			assertThat(new Bv(BigInteger.valueOf(-128), 1, true).toString(), equalTo("0x80"));
+			assertThat(new Bv(BigInteger.valueOf(-5), 1, true).toString(), equalTo("0xFB"));
+			assertThat(new Bv(BigInteger.valueOf(-1), 1, true).toString(), equalTo("0xFF"));
+
+			assertThat(new Bv(BigInteger.valueOf(-1), 8, true).toString(), equalTo("0xFFFFFFFFFFFFFFFF"));
+			assertThat(new Bv(BigInteger.ONE.shiftLeft(63).negate(), 8, true).toString(), equalTo("0x8000000000000000"));
+		}
+
+		@Test
+		void toStringOfSignedPositiveBaseValueOutputsHexWithoutTwosComplement() {
```
Jaguar0625, line 343: for positive, two's complement should be same as unsigned? so not sure purpose of test?

sdk/java/src/test/java/org/symbol/sdk/BaseValueTest.java
```diff
@@ -0,0 +1,499 @@
...
+			// Act + Assert:
+			assertThat(new Bv(BigInteger.valueOf(0x7F), 1, true).toString(), equalTo("0x7F"));
+			assertThat(new Bv(BigInteger.valueOf(0x1234), 2, true).toString(), equalTo("0x1234"));
+		}
+
+		@Test
+		void toHexStringStaticMethodWorksForUnsigned() {
```
Jaguar0625, line 349: why does toHexString need to be public?

sdk/java/src/test/java/org/symbol/sdk/BaseValueTest.java
```diff
@@ -0,0 +1,499 @@
...
+			assertThat(hex8, equalTo("0xFFFFFFFFFFFFFFFF"));
+		}
+	}
+
+	// endregion
+
+	// region JSON Representation
+
+	@Nested
+	final class JsonRepresentation {
```
Jaguar0625, line 380: not sure if you want to test more variants like js tests?

sdk/java/src/test/java/org/symbol/sdk/BaseValueTest.java
```diff
@@ -0,0 +1,499 @@
...
+		}
+
+		@Test
+		void hashCodeConsistentWithEquals() {
+			// Arrange:
+			final Bv v1 = new Bv(BigInteger.valueOf(5), 4, false);
+			final Bv v2 = new Bv(BigInteger.valueOf(5), 4, false);
+
+			// Act + Assert:
+			if (v1.equals(v2))
```
Jaguar0625: remove conditional ?

sdk/java/src/test/java/org/symbol/sdk/BaseValueTest.java
```diff
@@ -0,0 +1,499 @@
...
+			// Arrange:
+			final Bv v1 = new Bv(BigInteger.valueOf(-10), 4, true);
+			final Bv v2 = new Bv(BigInteger.valueOf(10), 4, true);
+
+			// Act + Assert:
+			assertThat(v1.compareTo(v2) < 0, equalTo(true));
+		}
+
+		@Test
+		void compareToIgnoresSizeAndSignedness() {
```
Jaguar0625: should this be error?

sdk/java/src/test/java/org/symbol/sdk/Bip32Test.java
```diff
@@ -0,0 +1,240 @@
...
+		assertThrows(IllegalArgumentException.class, () -> new Bip32().fromMnemonic(bad, ""));
+	}
+
+	@Test
+	void fromMnemonic_rejectsBadChecksum() {
+		// Arrange:
+		// Take the deterministic mnemonic and swap two words to break the checksum.
+		final String bad = "abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon";
+		// Above is a valid all-zero entropy mnemonic ("abandon ... about" actually). Let's break it instead:
+		final String broken = "abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon ";
```
Jaguar0625: aren't bad and broken the same or am i blind? 😅

sdk/java/src/test/java/org/symbol/sdk/Bip32Test.java
```diff
@@ -0,0 +1,240 @@
...
+		// sanity: canonical phrase derives
+		bip32.fromMnemonic(mnemonic, "TREZOR");
+		// double space / trailing space derive different seed bytes — must be rejected
+		org.junit.jupiter.api.Assertions.assertThrows(IllegalArgumentException.class,
+				() -> bip32.fromMnemonic(mnemonic.replaceFirst(" ", "  "), "TREZOR"));
+		org.junit.jupiter.api.Assertions.assertThrows(IllegalArgumentException.class, () -> bip32.fromMnemonic(mnemonic + " ", "TREZOR"));
+	}
+
+	@Test
+	void allBundledWordlistsDeriveRoundTrip() {
```
Jaguar0625: not sure if this is round trip? for roundtrip you'd want to reextract mnemonic from node and compare to original?

sdk/java/src/test/java/org/symbol/sdk/ByteArrayTest.java
```diff
@@ -0,0 +1,421 @@
...
+		}
+
+		@Test
+		void equalsIsSymmetric() {
+			// Arrange:
+			final Fixed ba1 = new Fixed(TEST_BYTES);
+			final Fixed ba2 = new Fixed(TEST_HEX);
+
+			// Act + Assert:
+			assertThat(ba1.equals(ba2), equalTo(ba2.equals(ba1)));
```
Jaguar0625: would split into two asserts (should both be true)

sdk/java/src/test/java/org/symbol/sdk/ByteArrayTest.java
```diff
@@ -0,0 +1,421 @@
...
+
+		@Test
+		void equalsIsTransitive() {
+			// Arrange:
+			final Fixed ba1 = new Fixed(TEST_BYTES);
+			final Fixed ba2 = new Fixed(TEST_HEX);
+			final Fixed ba3 = new Fixed(Converter.hexToUint8(TEST_HEX));
+
+			// Act + Assert:
+			if (ba1.equals(ba2) && ba2.equals(ba3))
```
Jaguar0625: not sure purpose of if?

sdk/java/src/test/java/org/symbol/sdk/ByteArrayTest.java
```diff
@@ -0,0 +1,421 @@
...
+	@Nested
+	final class Hashing {
+		@Test
+		void hashCodeIsConsistentWithEquals() {
+			// Arrange:
+			final Fixed ba1 = new Fixed(TEST_BYTES);
+			final Fixed ba2 = new Fixed(TEST_HEX);
+
+			// Act + Assert:
+			if (ba1.equals(ba2))
```
Jaguar0625: remove if?

sdk/java/src/test/java/org/symbol/sdk/ByteArrayTest.java
```diff
@@ -0,0 +1,421 @@
...
+		@Test
+		void byteArrayInputReturnsDirectReference() {
+			// Arrange:
+			final byte[] input = TEST_BYTES.clone();
+
+			// Act:
+			final byte[] result = ByteArray.toBytes(input);
+
+			// Assert:
+			assertThat(result == input, equalTo(true)); // Same reference
```
Jaguar0625: sameInstance ?

sdk/java/src/test/java/org/symbol/sdk/CatbufferTypeTest.java
```diff
@@ -0,0 +1,173 @@
...
+		}
+
+		@Override
+		public Object toJson() {
+			return json;
+		}
+	}
+
+	@Test
+	void serializeSizesWriterAndThreadsSerializeInto() {
```
Jaguar0625: why is Threads in test name ?

sdk/java/src/test/java/org/symbol/sdk/CatbufferTypeTest.java
```diff
@@ -0,0 +1,173 @@
...
+		// Act + Assert:
```
Jaguar0625: can remove 		`// Act + Assert:`  for tests with only asserts

sdk/java/src/test/java/org/symbol/sdk/CatbufferTypeTest.java
```diff
@@ -0,0 +1,173 @@
...
+
+	@Test
+	void toJsonStringWrapsSerializationFailureAsIllegalState() {
+		// Arrange:
+		final java.util.Map<String, Object> selfReferential = new java.util.HashMap<>();
+		selfReferential.put("self", selfReferential); // Jackson rejects the direct cycle with a JsonProcessingException
+
+		// Act + Assert:
+		assertThrows(IllegalStateException.class, () -> new JsonFixture(selfReferential).toJsonString());
+	}
```
Jaguar0625, line 149: this seems to be testing Jackson and mock class, not base class behavior?

sdk/java/src/test/java/org/symbol/sdk/CatbufferTypeTest.java
```diff
@@ -0,0 +1,173 @@
...
+		// Act + Assert:
+		assertThat(fixture.typeHints(), anEmptyMap());
+	}
+
+	@Test
+	void asListChecksElementsAndPassesNullThrough() {
+		// Act + Assert:
+		assertThat(CatbufferType.asList(null, String.class), nullValue());
+		assertThat(CatbufferType.asList(List.of("a", "b"), String.class), contains("a", "b"));
+		assertThrows(ClassCastException.class, () -> CatbufferType.asList(List.of(1, 2), String.class));
```
Jaguar0625, line 183: for batch test cases, with failures and successes, consider new line between success and failures for readability

Jaguar0625: alternatively, split these into separate tests

sdk/java/src/test/java/org/symbol/sdk/CryptoExceptionTest.java
```diff
@@ -0,0 +1,187 @@
...
+			// Act:
+			final CryptoException ce = new CryptoException(message);
+
+			// Assert:
+			assertThat(ce.getMessage(), equalTo(message));
+			assertThat(ce.getCause(), nullValue());
+		}
+
+		@Test
+		void messageIsPreserved() {
```
Jaguar0625: isn't this redundant to previous

sdk/java/src/test/java/org/symbol/sdk/CryptoExceptionTest.java
```diff
@@ -0,0 +1,187 @@
...
+		void nullMessageWithCauseIsAccepted() {
+			// Arrange:
+			final IOException cause = new IOException("I/O error");
+
+			// Act:
+			final CryptoException ce = new CryptoException((String) null, cause);
+
+			// Assert:
+			assertThat(ce.getCause(), sameInstance(cause));
+			assertThat(ce.getMessage(), nullValue());
```
Jaguar0625: prefer to keep field checking order consistent (this is only one where cause and message seem swapped)

sdk/java/src/test/java/org/symbol/sdk/CryptoExceptionTest.java
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	// endregion
+
+	// region Cause-Only Constructor
+
+	@Nested
+	final class CauseOnlyConstructor {
+		@Test
+		void canCreateWithOnlyCause() {
```
Jaguar0625: seems redundant with following

sdk/java/src/test/java/org/symbol/sdk/CryptoExceptionTest.java
```diff
@@ -0,0 +1,187 @@
...
+			assertThat(ce.getCause(), sameInstance(cause));
+		}
+
+		@Test
+		void nullCauseIsAccepted() {
+			// Act:
+			final CryptoException ce = new CryptoException((Throwable) null);
+
+			// Assert: Should not throw
+			assertThat(ce.getCause(), nullValue());
```
Jaguar0625, line 162: also check message

sdk/java/src/test/java/org/symbol/sdk/CipherTypesTest.java
```diff
@@ -0,0 +1,172 @@
...
+		@Test
+		void canDecrypt() {
+			// Act + Assert:
+			for (int i = 0; i < testCases.length; ++i) {
+				final TestCase tc = testCases[i];
+				final CipherTypes.AesCbcCipher cipher = new CipherTypes.AesCbcCipher(tc.sharedKey());
+				final byte[] result = cipher.decrypt(tc.cipherTextWithTag(), tc.iv());
+				assertThat("id " + i, result, equalTo(tc.clearText()));
+			}
+		}
```
Jaguar0625, line 99: can you make these generic ?

sdk/java/src/test/java/org/symbol/sdk/CryptoTypesTest.java
```diff
@@ -0,0 +1,199 @@
...
+			assertThat(zero.bytes().length, equalTo(32));
+			for (byte b : zero.bytes())
+				assertThat(b, equalTo((byte) 0));
+		}
+
+		@Test
+		void cannotCreateFromInvalidLength() {
+			// Act + Assert:
+			assertThrows(IllegalArgumentException.class, () -> new CryptoTypes.Hash256(new byte[31]));
+		}
```
Jaguar0625, line 56: did you want similar test for other types too ?

sdk/java/src/test/java/org/symbol/sdk/CryptoTypesTest.java
```diff
@@ -0,0 +1,199 @@
...
+
+		@Test
+		void canCreateFromHexString() {
```
Jaguar0625, line 59: is this redundant by following test?

sdk/java/src/test/java/org/symbol/sdk/CryptoTypesTest.java
```diff
@@ -0,0 +1,199 @@
...
+		private static final String HEX = "1A1A1A1A1A1A1A1A1A1A1A1A1A1A1A1A1A1A1A1A1A1A1A1A1A1A1A1A1A1A1A1A";
+
+		@Test
+		void hasCorrectSize() {
+			// Act + Assert:
+			assertThat(CryptoTypes.PublicKey.SIZE, equalTo(32));
+		}
+
+		@Test
+		void copyConstructorPreservesBytes() {
```
Jaguar0625, line 106: isn't this relevant for all types?

sdk/java/src/test/java/org/symbol/sdk/CryptoTypesTest.java
```diff
@@ -0,0 +1,199 @@
...
+			final SecureRandom random = new SecureRandom();
+			final byte[] bytes = new byte[32];
+			random.nextBytes(bytes);
+			final CryptoTypes.PublicKey original = new CryptoTypes.PublicKey(bytes);
+
+			// Act:
+			final CryptoTypes.PublicKey copy = new CryptoTypes.PublicKey(original);
+
+			// Assert:
+			assertThat(copy.bytes(), equalTo(original.bytes()));
```
Jaguar0625, line 117: i think you want to check is not same ?

sdk/java/src/test/java/org/symbol/sdk/CryptoTypesTest.java
```diff
@@ -0,0 +1,199 @@
...
+			assertThat(CryptoTypes.Signature.SIZE, equalTo(64));
+		}
+
+		@Test
+		void canCreateZero() {
+			// Act:
+			final CryptoTypes.Signature zero = CryptoTypes.Signature.zero();
+
+			// Assert:
+			assertThat(zero.bytes().length, equalTo(64));
```
Jaguar0625: check if contents are actually zero ?

Jaguar0625 (commented):
Good job 👍. Java is hopelessly verbose 😵‍💫!

sdk/java/src/test/java/org/symbol/sdk/impl/Ed25519Test.java
Wayonb: Yeah, I noticed that JS impl did not have any tests, but I need some to verify the implementation. With that said, these tests can be removed.

sdk/java/src/main/java/org/symbol/sdk/Bip32.java
Wayonb: Yea, I had a TODO to update the impl.  The packages I found were not updated, iirc.  
What do you think of - https://github.com/lightsail-network/mnemonic4j (it's Kotlin)
https://github.com/NovaCrypto/BIP39 - This is Java, but it has not been updated for a while

sdk/java/src/test/java/org/symbol/sdk/impl/Ed25519Test.java
Jaguar0625: can either remove them or leave them for now and refactor once you have the nem/symbol specific ones

sdk/java/src/main/java/org/symbol/sdk/Bip32.java
Jaguar0625: either is fine. java one seems a bit more used, but kotlin one is fine if you prefer interface.

sdk/java/src/test/java/org/symbol/sdk/CryptoTypesTest.java
Wayonb: PublicKey is the only Crypto type that has a copy constructor

sdk/java/src/main/java/org/symbol/sdk/Bip32.java
Wayonb: Went with mnemonic4j cause the other package is not available on Maven(only as Jit)

sdk/java/src/main/java/org/symbol/sdk/CatbufferType.java
Wayonb: no, this is called by the sub class.  added some test to show.

sdk/java/src/test/java/org/symbol/sdk/utils/BufferViewTest.java
Wayonb: Yes, but it seems to be the same behavior as JS. 🤔

sdk/java/src/test/java/org/symbol/sdk/utils/BufferViewTest.java
Wayonb: I fixed the logic so it works like JS. So it will create the new buffer.

sdk/java/src/test/java/org/symbol/sdk/BaseValueTest.java
Wayonb: It's the same for all subclasses, so one less for the generator.

sdk/java/src/test/java/org/symbol/sdk/BaseValueTest.java
Wayonb: Yes, but I like to test all combination to be sure 😅

sdk/java/src/test/java/org/symbol/sdk/utils/BufferViewTest.java
Jaguar0625: you're right. i misread the test. behavior is correct. it's operating on subview [2, 5] not original [0, 11]. so expected is subview [2,5].

sdk/java/src/test/java/org/symbol/sdk/impl/Ed25519Test.java
```diff
@@ -0,0 +1,199 @@
...
+				"3D4017C3E843895A92B70AA74D1B7EBC9C982CCF2EC4968CC0CD55F12AF4660C", "72",
+				"92A009A9F0D4CAB8720E820B5F642540A2B27B5416503F8FB3762223EBDB69DA"
+						+ "085AC1E43E15996E458F3613D0F11D8C387B2EAEB4302AEEB00D291612BB0C00");
+
+		// Assert (test2 additionally round-trips through verify):
+		final Tweetnacl.KeyPair kp = Ed25519.keyPairFromSeed(Tweetnacl.HashMode.SHA2_512,
+				Converter.hexToUint8("4CCD089B28FF96DA9DB6C346EC114E0F5B8A319F35ABA624DA8CF6ED4FB8A6FB"));
+		final byte[] msg = Converter.hexToUint8("72");
+		final byte[] sig = Ed25519.sign(Tweetnacl.HashMode.SHA2_512, msg, kp);
+		assertThat(Ed25519.verify(Tweetnacl.HashMode.SHA2_512, msg, sig, kp.publicKey), is(true));
```
Jaguar0625: would move round trip test to separate test

sdk/java/src/test/java/org/symbol/sdk/impl/Ed25519Test.java
Jaguar0625: yes, it is > L, but i don't think it is still on the curve?

sdk/java/src/main/java/org/symbol/sdk/BaseValue.java
```diff
@@ -0,0 +1,134 @@
...
+	private static void requireRange(final long value, final int size, final boolean isSigned) {
+		if (1 != size && 2 != size && 4 != size && 8 != size)
+			throw new IllegalArgumentException(String.format("\"size\" (%d) must be 1, 2, 4 or 8 bytes", size));
+
+		if (8 == size)
+			return;
+
+		final int bits = 8 * size;
+		final long lowerBound = isSigned ? -(1L << (bits - 1)) : 0L;
+		final long upperBound = (1L << (isSigned ? bits - 1 : bits)) - 1;
```
Jaguar0625, line 116: consider refactoring into helper function since you have this bounds checking in multiple places

sdk/java/src/test/java/org/symbol/sdk/utils/ArrayHelpersTest.java
```diff
@@ -1,675 +1,726 @@
...
+			final byte[] part = {
+					1, 2, 3
+			};
+			final byte[] result = ArrayHelpers.concat(part);
+
+			// Assert: content equal, but a new array (concat always allocates)
+			assertThat(result, equalTo(new byte[]{
+					1, 2, 3
+			}));
+			assertThat(result != part, equalTo(true));
```
Jaguar0625: use sameInstance?

sdk/java/src/test/java/org/symbol/sdk/utils/ArrayHelpersTest.java
```diff
@@ -1,675 +1,726 @@
...
+			};
+
+			// Act:
+			final byte[] reversed = ArrayHelpers.reverse(input);
+
+			// Assert: input untouched, result is a distinct array
+			assertThat(input, equalTo(new byte[]{
+					1, 2, 3
+			}));
+			assertThat(reversed != input, equalTo(true));
```
Jaguar0625: sameInstance?

sdk/java/src/test/java/org/symbol/sdk/utils/BufferViewTest.java
```diff
@@ -4,259 +4,572 @@
...
+		}
+
+		@Test
+		void canCreateShrinkingSubview() {
+			// Arrange:
+			final BufferView view = newView();
+			view.shiftRight(3);
+
+			// Act:
+			final BufferView window = view.window(2);
```
Jaguar0625, line 178: consider either increasing length of original buffer or shifting less, so window is not to end of original buffer, but a subset. similar thing in a shift test.

sdk/java/src/test/java/org/symbol/sdk/utils/BufferViewTest.java
```diff
@@ -4,259 +4,572 @@
...
+
+		@Test
+		void canPeek1ByteSigned() {
+			assertPeekInt(new byte[]{
+					(byte) 0xFF, 0x01, 0x02
+			}, 1, true, -1L);
+		}
+
+		@Test
+		void canPeek2ByteUnsignedLittleEndian() {
```
Jaguar0625: not sure if you need "LittleEndian" in test names since there are no big endian variants

sdk/java/src/test/java/org/symbol/sdk/utils/BufferViewTest.java
```diff
@@ -4,259 +4,572 @@
...
+
+		@Test
+		void canPeek4ByteSignedLittleEndian() {
+			assertPeekInt(new byte[]{
+					(byte) 0xFF, (byte) 0xFF, (byte) 0xFF, (byte) 0xFF, 0x00
+			}, 4, true, -1L);
+		}
+
+		@Test
+		void canPeek8ByteLittleEndianU64ReadsBackNegative() {
```
Jaguar0625: consider breaking into (signed, unsigned) two to match < 8 byte reads

sdk/java/src/test/java/org/symbol/sdk/utils/BufferViewTest.java
```diff
@@ -4,259 +4,572 @@
...
+			final BufferView view = new BufferView(new byte[]{
+					0x01, 0x02, 0x03, 0x04
+			});
+
+			// Act + Assert: the window bounds check fires before the read (clear message, not a raw ByteBuffer error).
+			assertThrows(IndexOutOfBoundsException.class, () -> view.peekInt(8, false));
+		}
+
+		@Test
+		void peekAtDifferentPositions() {
```
Jaguar0625: canPeek

sdk/java/src/test/java/org/symbol/sdk/utils/ConverterTest.java
```diff
@@ -1,220 +1,551 @@
...
+
+		@Test
+		void readsInt32Unsigned() {
+			// 0x80 high bit: 0x80000000 unsigned reads back as -2147483648 when signed
+			assertReads(new byte[]{
+					0x00, 0x00, 0x00, (byte) 0x80
+			}, 4, true, -2147483648L);
+		}
+
+		@Test
```
Jaguar0625, line 266: would move after `readsUint32`. consider also matching readsInt64 test (where isSigned is true)

sdk/java/src/test/java/org/symbol/sdk/utils/ConverterTest.java
```diff
@@ -1,220 +1,551 @@
...
+		void writesInt32Negative() {
+			assertWrites(-1L, 4, true, new byte[]{
+					(byte) 0xFF, (byte) 0xFF, (byte) 0xFF, (byte) 0xFF
+			});
+		}
 
-		// Act:
-		final String actual = Converter.bytesToHex(input);
+		@Test
+		void writesUint64() {
```
Jaguar0625: would move up after writesUint32

sdk/java/src/test/java/org/symbol/sdk/utils/ConverterTest.java
```diff
@@ -1,220 +1,551 @@
...
+
+		@Test
+		void toIntParsesAndRangeChecksStrings() {
+			assertThat(Converter.toInt("255"), equalTo(255));
+			assertThat(Converter.toInt("0xFF"), equalTo(255));
+			assertThrows(IllegalArgumentException.class, () -> Converter.toInt("4294967296")); // beyond the unsigned 32-bit range
+		}
+
+		@Test
+		void toIntRejectsMagnitudesAboveUnsignedRange() {
```
Jaguar0625: indicate in name that String parsing is being tested ?

sdk/java/src/test/java/org/symbol/sdk/utils/ConverterTest.java
```diff
@@ -1,220 +1,551 @@
...
+		@Test
+		void toIntHexAcceptsFullUnsignedRange() {
+			// hex reads the full unsigned u32 range; a set bit 31 returns the negative two's-complement int
+			assertThat(Converter.toInt("0x7FFFFFFF"), equalTo(Integer.MAX_VALUE));
+			assertThat(Converter.toInt("0x80000000"), equalTo(Integer.MIN_VALUE));
+			assertThat(Converter.toInt("0xFFFFFFFF"), equalTo(-1));
+			// beyond the unsigned 32-bit range is rejected
+			assertThrows(IllegalArgumentException.class, () -> Converter.toInt("0x100000000"));
+		}
+	}
```
Jaguar0625, line 577: would generally recommend having the toInt and toLong tests mirror each other since they are testing the same behaviors. everything is covered, but would expect a bit more symmetry in tests, e.g. `to{Int|Long}HexAcceptsFullUnsignedRange`

sdk/java/src/test/java/org/symbol/sdk/BaseValueTest.java
```diff
@@ -0,0 +1,565 @@
...
+	private static void canCreateSigned(final long raw, final int size, final long expected) {
+		// Act:
+		final Bv v = new Bv(raw, size, true);
+
+		// Assert:
+		assertThat(v.size(), equalTo(size));
+		assertThat(v.value(), equalTo(expected));
+	}
+
+	private static void assertOutOfRange(final Runnable r, final String bitWidth) {
```
Jaguar0625: bitWidth => expected | message ?

Jaguar0625: since you are passing "must be 1, 2, 4 or 8 bytes" in cannotCreateWithUnsupportedSize

sdk/java/src/test/java/org/symbol/sdk/BaseValueTest.java
```diff
@@ -0,0 +1,565 @@
...
+
+		// note: there is no signed long (size 8) out-of-range case — it admits any 64-bit pattern
+
+		@Test
+		void cannotCreateSignedByteOutsideRange() {
+			assertOutOfRange(() -> new Bv(-0x81L, 1, true), "8-bit");
+			assertOutOfRange(() -> new Bv(0x80L, 1, true), "8-bit");
+		}
+
+		@Test
```
Jaguar0625, line 201: add short test ?

sdk/java/src/test/java/org/symbol/sdk/BaseValueTest.java
```diff
@@ -0,0 +1,565 @@
...
+			final Bv v1 = new Bv(42L, 4, false);
+			final Bv v2 = new Bv(42L, 4, false);
+			final Bv v3 = new Bv(42L, 8, false);
+
+			// Act + Assert: equal values must hash equally (the one property the hashCode contract guarantees)
+			assertThat(v1.hashCode(), equalTo(v2.hashCode()));
+
+			// a value differing only in size is a distinct object — assert that at the equals level; pinning distinct hash
+			// codes would test an accidental non-collision that the contract does not require (unequal objects may collide)
+			assertEquality(v1, v3, false);
```
Jaguar0625: do you mean to check hash codes here ?

sdk/java/src/test/java/org/symbol/sdk/BaseValueTest.java
Jaguar0625: as well as some with leading zeros

sdk/java/src/test/java/org/symbol/sdk/Bip32Test.java
```diff
@@ -0,0 +1,330 @@
...
+			};
+
+			for (final String[] testCase : cases) {
+				final String language = testCase[0];
+				final Bip32 bip32 = new Bip32("ed25519", language);
+				// fixed mnemonic -> pinned root private key (pins seed derivation per language)
+				final Bip32.Bip32Node node = bip32.fromMnemonic(testCase[1], "password");
+				assertThat(language, Converter.uint8ToHex(node.privateKey.bytes()), equalTo(testCase[2]));
+				// a freshly generated mnemonic in this language still round-trips to a real 32-byte root key
+				assertThat(language, bip32.fromMnemonic(bip32.random(), "password").privateKey.bytes(), not(equalTo(new byte[32])));
```
Jaguar0625, line 257: // Arrange // Act // Assert comments

sdk/java/src/test/java/org/symbol/sdk/CatbufferTypeTest.java
```diff
@@ -0,0 +1,300 @@
...
+		assertCoerces(CatbufferType::asInt, 300, 300);
+		assertCoerces(CatbufferType::asInt, "300", 300);
+	}
+
+	@Test
+	void asLongCoercesNumberAndHexString() {
+		// numbers coerce; asLong also accepts a 0x-hex string (the canonical u64 >= 2^63 form).
+		assertCoerces(CatbufferType::asLong, 300L, 300L);
+		assertCoerces(CatbufferType::asLong, "0x12C", 300L);
+	}
```
Jaguar0625, line 287: would test both dec and hex string in both

sdk/java/src/test/java/org/symbol/sdk/CipherTypesTest.java
```diff
@@ -0,0 +1,191 @@
...
+
+		public void assertCanEncrypt() {
+			assertCanEncryptOrDecrypt(true);
+		}
+
+		public void assertCanDecrypt() {
+			assertCanEncryptOrDecrypt(false);
+		}
+
+		void assertDecryptRejectsWrongIv() {
```
Jaguar0625: since this is not generic, you can move it out

sdk/java/src/test/java/org/symbol/sdk/CipherTypesTest.java
```diff
@@ -0,0 +1,191 @@
...
+								+ "47F8434133989BA77DDA561B7E3701B9A83C3BA7660C666BA59FEF96598EB621"
+								+ "544C63806D509AC47697412F9564EB0A2E1F72F6599F5666AF34CFFCA06573FF"
+								+ "B4F47B02F59F21C64363DAECB977B4415F19FDDA3C9AAE5066A57B669FFAA257",
+						"C877A76BF595560772167C6E3BCC705305DB9C6FCBEB90F4FEA85116038BC53C"
+								+ "3FA5B4B4EA0DE5CC534FBE1CF9AE44824C6C2C0A5C885BD8C3CDC906F1267573"
+								+ "7E434B983E1E231A52A275DB5FB1A0CAC6A07B3B7DCB19482A5D3B06A9317A54"
+								+ "826CEA6B36FCE452FA9B5475E2AAF25499499D8A8932A19EB987C903BD8502FE")
+		};
+
+		final CipherTestHelper<CipherTypes.AesGcmCipher> helper = new CipherTestHelper<>(CipherTypes.AesGcmCipher::new, testCases);
```
Jaguar0625, line 149: use similar helper above too ?

sdk/java/src/test/java/org/symbol/sdk/CryptoTypesTest.java
Jaguar0625: since you're cloning, consider adding check for not(sameInstance)

sdk/java/src/test/java/org/symbol/sdk/CryptoTypesTest.java
```diff
@@ -0,0 +1,179 @@
...
+			assertThat(zero.toString(), equalTo("0".repeat(128)));
```
Jaguar0625: consider standardizing zero check between Hash256 and Signature

sdk/java/src/test/java/org/symbol/sdk/impl/Ed25519Test.java
Jaguar0625: can you explain this test?

sdk/java/src/test/java/org/symbol/sdk/impl/SharedKeyHelpersTest.java
Jaguar0625: unsure if you want to move AAA comments?

sdk/java/README.md
Jaguar0625: update readme?

Jaguar0625 (commented):
👍

sdk/java/src/test/java/org/symbol/sdk/utils/BufferViewTest.java
```diff
@@ -285,7 +285,7 @@ private static void assertPeekBytesThrows(final int size) {
 		}
 
 		@Test
-		void copiesLeadingBytesWithoutAdvancing() {
+		void canCopiesLeadingBytesWithoutAdvancing() {
```
Jaguar0625: canCopies => canCopy

sdk/java/src/test/java/org/symbol/sdk/BaseValueTest.java
```diff
@@ -170,6 +173,12 @@ void cannotCreateSignedByteOutsideRange() {
 			assertOutOfRange(() -> new Bv(0x80L, 1, true), "8-bit");
 		}
 
+		@Test
+		void cannotCreateSignedShortOutsideRange() {
+			assertOutOfRange(() -> new Bv(-0x800001L, 2, true), "16-bit");
+			assertOutOfRange(() -> new Bv(0x800000L, 2, true), "16-bit");
```
Jaguar0625: did you intentionally not use one more/less than max/min  (0x8000, -0x8001)

sdk/java/src/test/java/org/symbol/sdk/CatbufferTypeTest.java
```diff
@@ -275,12 +275,14 @@ void asIntCoercesNumberAndString() {
 		// numbers coerce; asInt also accepts a decimal string.
 		assertCoerces(CatbufferType::asInt, 300, 300);
 		assertCoerces(CatbufferType::asInt, "300", 300);
+		assertCoerces(CatbufferType::asLong, "0x12C", 300L);
```
Jaguar0625: asInt

Jaguar0625 (commented):
👍

sdk/java/src/main/java/org/symbol/sdk/BaseValue.java
Wayonb: This is in reference to WriteInt? 🤔
The checks are a little different.

Jaguar0625: you're right - ignore

Jaguar0625 (approved):
🎊

PR #2049, opened 2026-06-18
[jenkins] fix: allow different version of Java in CI

Wayonb:
problem: currently only Java 11 is used in CI
solution: add multiple versions of Java to the CI build


jenkins/shared-library/resources/buildEnvironment/baseImages.yaml
```diff
@@ -4,13 +4,16 @@ ubuntu-lts:
   image: ubuntu:24.04
   python: 3.12
   nodejs: 24
+  java: 11
```
Jaguar0625, line 7: why are you leaving this as 11? shouldn't it be somewhere between 17 and 21?

Wayonb: Java 11 is the current LTS.  If I upgrade it, it will break NEM and SN builds.  But you are current, after we upgrade NEM then the Java should be 17, 21 and 25.

Jaguar0625 (approved):
👍

Jaguar0625 (approved):
👍

Jaguar0625 (approved):
👍

PR #2050, opened 2026-06-18
[jenkins] fix: default to public build if build configuration is not set

Wayonb:
problem: The default CI build configuration is not set when triggered
solution: default to public build if configuration is not set


Jaguar0625:
why default to public instead of private?

Wayonb:
> why default to public instead of private?

For private builds, Jenkins will delete the package-lock.json file, which causes issues on certain projects(mainly with React projects).  At some point, private build should be the default again, but we need to upgrade all those projects.

Jaguar0625 (approved):
👍

PR #2066, opened 2026-06-29
[client/catapult] fix: reverse CatRealloc copyTo args, wipe freed pool slots

### LINKED ISSUE #2030: [client/catapult] Reversed arguments in CatRealloc corrupt returned buffer and pool state
The custom OpenSSL's [memory hook `CatRealloc` ](https://github.com/symbol/symbol/blob/dev/client/catapult/src/catapult/crypto/OpensslMemory.cpp#L51-L63) passes its arguments to 'copyTo' in the wrong order. 

When the hook is engaged to reallocate a pool managed buffer the old data is silently lost, the returned buffer contains uninitialized memory and the original pool slot is is overwritten with garbage before being freed.

The actual signature of `copyTo` is `(void* pDestination, const void* pSource, size_t destinationSize)` but the actual call swaps destination and source causing the unwanted behavior.

```
		auto result = malloc(newSize);
		if (!result)
			return nullptr;

		GetPool().copyTo(p, result, newSize);
		GetPool().free(p);
		return result;
```

This is also confirmed by the two [copyTo tests](https://github.com/symbol/symbol/blob/dev/client/catapult/tests/catapult/crypto/OpensslMemoryTests.cpp#L203-L238) that correctly pass the arguments in the right order.

There's actually no test available for `CatRealloc`

Being a _hook_ to OpenSSL's memory management is not sure when/if this code triggers: the criticality though stands in the fact it might be engaged by changed memory policies in a third-party component.

AndreaLanfranchi:
## Summary

- Fixes reversed `copyTo(p, result, newSize)` → `copyTo(result, p, newSize)` in `CatRealloc` (`OpensslMemory.cpp:57`). Bug caused OpenSSL to receive an uninitialized heap buffer instead of a copy of the original pool slot, corrupting all subsequent crypto operations on that context.
- Replaces `volatile` zero-wipe loop in `FixedSizePool::free()` with `OPENSSL_cleanse` — guaranteed not elided by the compiler, consistent with OpenSSL's own key material erasure convention.

## What the bug did

`CatRealloc` is the custom `CRYPTO_realloc` hook installed at startup via `CRYPTO_set_mem_functions`. When OpenSSL reallocates a pool-managed buffer, the hook must copy old data to the new heap block before freeing the pool slot. With arguments reversed:

- `result` (freshly `malloc`'d, uninitialized) was passed as `pSource`
- `p` (the live pool slot) was passed as `pDestination`

OpenSSL received a garbage buffer. Any crypto operation on that context — SHA-512 signature verification/signing, SHA3-256 entity hashing, AES-256-GCM decryption, SHA-256 HMAC — would operate on corrupted state. The bug is latent under the original OpenSSL version but activates on any version upgrade that changes internal realloc patterns.

## Changes

| File | Change |
|------|--------|
| `src/catapult/crypto/OpensslMemory.cpp` | Swap `copyTo` arguments in `CatRealloc` |
| `src/catapult/crypto/OpensslMemory.h` | Replace `volatile` loop with `OPENSSL_cleanse` in `FixedSizePool::free()`; add `<openssl/crypto.h>` include |
| `tests/catapult/crypto/OpensslMemoryTests.cpp` | 5 new tests: slot zeroing on free, data preservation through realloc, source slot zeroing after realloc, end-to-end via `CRYPTO_malloc`/`CRYPTO_realloc` |

## Test plan

- [ ] `SpecializedOpensslPoolAllocator_ReallocPreservesData` — sentinel bytes survive realloc
- [ ] `SpecializedOpensslPoolAllocator_ReallocWipesSourceSlot` — source slot zeroed after realloc
- [ ] `SpecializedOpensslPoolAllocator_FreeWipesSlotContent` — `OPENSSL_cleanse` zeros slot on free
- [ ] `FixedSizePool_FreeWipesSlotContent` — same at `FixedSizePool` layer
- [ ] `SpecializedOpensslPoolAllocator_ReallocViaOpenSslApi` — end-to-end through `CRYPTO_malloc`/`CRYPTO_realloc`

Fixes #2030

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/client/catapult/src/catapult/crypto/OpensslMemory.h
+++ b/client/catapult/src/catapult/crypto/OpensslMemory.h
@@ -86,7 +86,6 @@ namespace catapult { namespace crypto {
 			auto diff = static_cast<size_t>(reinterpret_cast<uint8_t*>(ptr) - m_pBuffer);
 			size_t index = diff / TPoolTraits::Element_Size;
 
-			OPENSSL_cleanse(ptr, TPoolTraits::Element_Size);
 			m_occupied[index].clear();
 		}
 
--- a/client/catapult/src/catapult/crypto/OpensslMemory.h
+++ b/client/catapult/src/catapult/crypto/OpensslMemory.h
@@ -24,6 +24,7 @@
 #include <atomic>
 #include <memory>
 #include <vector>
+#include <openssl/crypto.h>
 
 namespace catapult { namespace crypto {
 
@@ -85,6 +86,7 @@ namespace catapult { namespace crypto {
 			auto diff = static_cast<size_t>(reinterpret_cast<uint8_t*>(ptr) - m_pBuffer);
 			size_t index = diff / TPoolTraits::Element_Size;
 
+			OPENSSL_cleanse(ptr, TPoolTraits::Element_Size);
 			m_occupied[index].clear();
 		}
 
--- a/client/catapult/tests/catapult/crypto/OpensslMemoryTests.cpp
+++ b/client/catapult/tests/catapult/crypto/OpensslMemoryTests.cpp
@@ -377,46 +377,6 @@ namespace catapult { namespace crypto {
 
 	// region CatRealloc and free
 
-	TEST(TEST_CLASS, FixedSizePool_FreeWipesSlotContent) {
-		// Arrange: allocate a slot and fill with a known pattern.
-		PoolBuffer buffer;
-		TestPool pool(buffer.data());
-		auto* pElement = pool.tryAllocate();
-		ASSERT_TRUE(!!pElement);
-		std::fill(pElement, pElement + Element_Size, uint8_t(0xAB));
-
-		// Act:
-		pool.free(pElement);
-
-		// Assert: re-allocating returns the same slot address (index-0 is freed first)
-		// and the slot is now zeroed.
-		auto* pReused = pool.tryAllocate();
-		ASSERT_EQ(pElement, pReused);
-		auto numZeroBytes = static_cast<size_t>(std::count(pReused, pReused + Element_Size, uint8_t(0)));
-		EXPECT_EQ(Element_Size, numZeroBytes) << "free() must zero the slot before releasing it";
-
-		pool.free(pReused);
-	}
-
-	TRAIT_BASED_ALLOCATOR_TEST(SpecializedOpensslPoolAllocator_FreeWipesSlotContent) {
-		// Arrange: allocate a pool slot and fill with a known pattern.
-		Allocator allocator;
-		auto* pSlot = allocator.allocate(TTraits::Unaligned_Element_Size);
-		ASSERT_TRUE(!!pSlot);
-		std::fill(pSlot, pSlot + TTraits::Unaligned_Element_Size, uint8_t(0xAB));
-
-		// Act:
-		allocator.free(pSlot);
-
-		// Assert: re-allocating yields the same address with zeroed content.
-		auto* pReused = allocator.allocate(TTraits::Unaligned_Element_Size);
-		ASSERT_EQ(pSlot, pReused);
-		auto numZeroBytes = static_cast<size_t>(std::count(pReused, pReused + TTraits::Unaligned_Element_Size, uint8_t(0)));
-		EXPECT_EQ(TTraits::Unaligned_Element_Size, numZeroBytes) << "free() must zero the slot before releasing it";
-
-		allocator.free(pReused);
-	}
-
 	TRAIT_BASED_ALLOCATOR_TEST(SpecializedOpensslPoolAllocator_ReallocPreservesData) {
 		// Verifies the corrected CatRealloc argument order: copyTo(result, p, size).
 		// Arrange: allocate a pool slot and fill with a known sentinel.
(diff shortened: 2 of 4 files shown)


client/catapult/src/catapult/crypto/OpensslMemory.h
```diff
@@ -85,6 +86,7 @@ namespace catapult { namespace crypto {
 			auto diff = static_cast<size_t>(reinterpret_cast<uint8_t*>(ptr) - m_pBuffer);
 			size_t index = diff / TPoolTraits::Element_Size;
 
+			OPENSSL_cleanse(ptr, TPoolTraits::Element_Size);
```
Jaguar0625: why do you want to explicitly zero all memory?

for sensitive data, we already wiping securely, so this would just be in effect for non-sensitive data, which seems like unnecessary overhead?

client/catapult/tests/catapult/crypto/OpensslMemoryTests.cpp
```diff
@@ -375,4 +375,152 @@ namespace catapult { namespace crypto {
...
+		constexpr size_t New_Size = TTraits::Unaligned_Element_Size + 64;
+		auto pResult = std::make_unique<uint8_t[]>(New_Size);
+		allocator.copyTo(pResult.get(), pSlot, New_Size);
+		allocator.free(pSlot);
+
+		// Assert: source slot is zeroed — verified by re-allocating it.
+		auto* pReused = allocator.allocate(TTraits::Unaligned_Element_Size);
+		ASSERT_EQ(pSlot, pReused) << "Expected original slot to be returned on next allocation";
+		auto numZeroBytes = static_cast<size_t>(std::count(pReused, pReused + TTraits::Unaligned_Element_Size, uint8_t(0)));
+		EXPECT_EQ(TTraits::Unaligned_Element_Size, numZeroBytes) << "source slot must be zeroed after realloc";
```
Jaguar0625: i believe you want to check full size of realloc'd buffer - TTraits::Unaligned_Element_Size + 64 - not the original size?

client/catapult/tests/catapult/crypto/OpensslMemoryTests.cpp
```diff
@@ -375,4 +375,152 @@ namespace catapult { namespace crypto {
...
+
+		allocator.free(pReused);
+	}
+
+	TEST(TEST_CLASS, SpecializedOpensslPoolAllocator_ReallocViaOpenSslApi) {
```
Jaguar0625: i am not quite sure what is the purpose of this test? it seems to be testing behavior of CRYPTO_set_mem_functions? 

it is also likely going to poison the tests run after this test to use these custom functions instead of the default (non-pooled) ones.

Jaguar0625 (commented):
👍

client/catapult/src/catapult/crypto/OpensslMemory.h
AndreaLanfranchi: If this is only for OpenSSL internal structs then I agree I bit redundant cause EVP's resets already wipe their memory. But I found in-depth defense made sense in case this FixedPool was used for something else. Honestly I didn't dig enough but in my experience zeroing a **known** released memory segment is always a good habit.  
If performance is a concern will remove even if the overhead cost is tiny,

client/catapult/tests/catapult/crypto/OpensslMemoryTests.cpp
AndreaLanfranchi: I think you confused pReused (the released pool slot) with pResult (the reallocated one) which has the extra 64 bytes size. The goal is to assert the original released pool slot has been zeroed. However if we decide to remove the cleanse per [this comment](https://github.com/symbol/symbol/pull/2066#discussion_r3491601204) then this test is not needed

client/catapult/tests/catapult/crypto/OpensslMemoryTests.cpp
AndreaLanfranchi: Agreed on both points. I didn't think about the persistance of custom functions after test succeeds. My mistake. Removing.

client/catapult/src/catapult/crypto/OpensslMemory.h
Jaguar0625: yes, i was mainly thinking about performance. would be good for @Wayonb to comment.

Wayonb: Yeah, since sensitive data is already zero, I would remove this.  Am I assuming AddressSanitizer should detect this? 🤔

AndreaLanfranchi: Ran a quick standalone microbenchmark to check, since guessing isn't useful here.

`OPENSSL_cleanse` on the actual pool element sizes (Pool2 = 104B, Pool3 = 224B, Pool1 = 400B), measured against a loop overhead baseline, O2, on my Ubuntu laptop:

104B: about 5.1 ns per call
224B: about 8.5 ns per call
400B: about 13.6 ns per call

That is the added cost per `free()`. Any real crypto op these pools back (hash, sign, EVP context reset) runs from 1 to 10 microseconds, which is 1000 to 10000 ns. Cleanse adds well under 1% of that per call. It will not show up against normal timing noise even under heavy load.

This was not a formal benchmark committed to `tests/bench/`, just an ad hoc check to put a number on it. Happy to add a proper one there if useful for the record. Given the margin, I would treat the removal decision as mainly about the stale data leak threat model rather than performance.

AndreaLanfranchi: @Wayonb re: "Am I assuming AddressSanitizer should detect this?"

ASan is not relevant here.

ASan poisons a block's shadow memory only when the real allocator's `free()` runs, then flags later access to that memory. Buffer content is irrelevant to it. Cleansed bytes and stale bytes look identical to ASan.

Also, this pool's `free()` never deallocates anything. It just clears a used flag on the slot. The underlying buffer stays allocated for the whole pool's lifetime. ASan never sees a real `free()` call for individual slots, so it has no visibility into slot reuse at all.

Cleanse's actual job: stop a reused slot from leaking the previous tenant's bytes to whoever gets it next.

AndreaLanfranchi: Removed in 9f32119a9.

client/catapult/tests/catapult/crypto/OpensslMemoryTests.cpp
AndreaLanfranchi: Cleanse removed, so this test is gone too. Dropped in 9f32119a9.

Jaguar0625 (approved):
👍

PR #2068, opened 2026-06-29
[client/catapult] fix: add reentrancy guards to LocalNode boot and shutdown

### LINKED ISSUE #2044: [client/catapult] DefaultLocalNode::shutdown() not idempotent causes double saveStateToDisk() on destruction
Investigating sporadic silent aborts while running batch iterations [HappyBlockchainIntegrityTests](https://github.com/symbol/symbol/issues/616) on Windows this emerges:

[`LocalNodeTestContext::reset()`](https://github.com/symbol/symbol/blob/dev/client/catapult/tests/int/node/test/LocalNodeTestContext.h#L209-L215) calls [`DefaultLocalNode::shutdown()`](https://github.com/symbol/symbol/blob/dev/client/catapult/src/catapult/local/server/LocalNode.cpp#L338-L343) explicitly then drops the unique_ptr, triggering `~DefaultLocalNode()` which calls `shutdown()` a second time. 

As `DefaultLocalNode::shutdown()` has no re-entry guard `saveStateToDisk()` runs twice in quick succession.
The second invocation operates on filesystem state left by the first: moveTo() inside SaveStateToDirectoryWithCheckpointing attempts to purge and rename a state/ directory that was just written milliseconds earlier. On Windows this races with OS file-handle retention (antivirus, search indexer), causing std::filesystem::remove or std::filesystem::rename to throw. 
The exception escapes ~DefaultLocalNode(), which is implicitly noexcept (since C++11), triggering std::terminate() and abort() with no log output.

On Linux, although the handle release by OS is immediate, the double saveStateToDisk is nevertheless redundant.

Need to protect shutdown() from re-entrancy.

Worth mentioning that [`saveStateToDisk()`](https://github.com/symbol/symbol/blob/dev/client/catapult/src/catapult/local/server/LocalNode.cpp#L346-L353) is apparently guarded by `m_isBooted` but problem is that is set true in boot() but never cleared in shutdown(), so is totally ineffective.

AndreaLanfranchi:
## Problem

`LocalNode::Impl` lacked guards against double-boot and double-shutdown, leaving the node in an undefined state if either method was called more than once.

Specific issues:
- `boot()` could be called twice, re-registering plugins and re-initializing subsystems on an already-running node.
- `shutdown()` could be called on an unbooted node or called multiple times, causing redundant teardown and potential use-after-free in pool shutdown.
- `saveStateToDisk()` guarded on `m_isBooted` internally, creating implicit coupling instead of an explicit contract at the call site.

## Changes

- `boot()`: throws `CATAPULT_THROW_RUNTIME_ERROR` if called when already booted.
- `shutdown()`: early-returns if node was never booted or has already been shut down.
- Adds `m_isShutdown` boolean member (initialized to `false`), set to `true` on first shutdown entry.
- Removes the redundant `m_isBooted` guard from `saveStateToDisk()` — `shutdown()` now owns that invariant.

Fixes #2044

### DIFF
--- a/client/catapult/src/catapult/local/server/LocalNode.cpp
+++ b/client/catapult/src/catapult/local/server/LocalNode.cpp
@@ -221,7 +221,8 @@ namespace catapult { namespace local {
 							m_dataDirectory))
 					, m_pTransactionStatusSubscriber(m_pBootstrapper->subscriptionManager().createTransactionStatusSubscriber())
 					, m_pluginManager(m_pBootstrapper->pluginManager())
-					, m_isBooted(false) {
+					, m_isBooted(false)
+					, m_isShutdown(false) {
 				ValidateNodes(m_pBootstrapper->staticNodes());
 				AddLocalNode(m_nodes, m_pBootstrapper->config());
 			}
@@ -232,6 +233,9 @@ namespace catapult { namespace local {
 
 		public:
 			void boot() {
+				if(m_isBooted)
+					CATAPULT_THROW_RUNTIME_ERROR("local node is already booted");
+
 				CATAPULT_LOG(info) << "registering system plugins";
 				m_pluginModules = LoadAllPlugins(*m_pBootstrapper);
 
@@ -336,6 +340,10 @@ namespace catapult { namespace local {
 
 		public:
 			void shutdown() override {
+				if (!m_isBooted || m_isShutdown)
+					return;
+
+				m_isShutdown = true;
 				utils::StackLogger stackLogger("shutting down local node", utils::LogLevel::info);
 
 				m_pBootstrapper->pool().shutdown();
@@ -344,9 +352,6 @@ namespace catapult { namespace local {
 
 		private:
 			void saveStateToDisk() {
-				// only save to storage if boot succeeded
-				if (!m_isBooted)
-					return;
 
 				constexpr auto SaveStateToDirectoryWithCheckpointing = extensions::SaveStateToDirectoryWithCheckpointing;
 				SaveStateToDirectoryWithCheckpointing(m_dataDirectory, m_config.Node, m_catapultCache, m_score.get());
@@ -404,6 +409,7 @@ namespace catapult { namespace local {
 			std::vector<utils::DiagnosticCounter> m_counters;
 			extensions::BannedNodeIdentitySink m_bannedNodeIdentitySink;
 			bool m_isBooted;
+			bool m_isShutdown;
 		};
 	}
 
--- a/client/catapult/src/catapult/local/server/LocalNode.cpp
+++ b/client/catapult/src/catapult/local/server/LocalNode.cpp
@@ -233,7 +233,7 @@ namespace catapult { namespace local {
 
 		public:
 			void boot() override {
-				if(m_isBooted)
+				if (m_isBooted)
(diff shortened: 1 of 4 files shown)


client/catapult/src/catapult/local/server/LocalNode.cpp
```diff
@@ -232,6 +233,9 @@ namespace catapult { namespace local {
 
 		public:
 			void boot() {
+				if(m_isBooted)
+					CATAPULT_THROW_RUNTIME_ERROR("local node is already booted");
```
Jaguar0625: can you add new test for this behavior?

AndreaLanfranchi: Done in [226a92c](https://github.com/symbol/symbol/pull/2068/commits/226a92c03592e06f41ccc33bceeec950c565c156)

client/catapult/src/catapult/local/server/LocalNode.cpp
```diff
@@ -231,7 +232,10 @@ namespace catapult { namespace local {
 			}
 
 		public:
-			void boot() {
+			void boot() override {
+				if(m_isBooted)
```
Jaguar0625: should be space after if

client/catapult/tests/int/node/basic/LocalNodeBasicTests.cpp
```diff
@@ -157,4 +157,29 @@ namespace catapult { namespace local {
...
+
+	TEST(TEST_CLASS, BootThrowsWhenNodeIsAlreadyBooted) {
+		// Arrange: create and boot a local node
+		TestContext context(NodeFlag::Regular);
+
+		// Act & Assert: calling boot again on an already booted node throws
+		EXPECT_THROW(context.localNode().boot(), catapult_runtime_error);
+	}
+
+	TEST(TEST_CLASS, ShutdownIsIdempotentAndCanBeCalledMultipleTimes) {
```
Jaguar0625, line 171: i assume this fails on windows without changes?

Jaguar0625 (approved):
👍

client/catapult/src/catapult/local/server/LocalNode.cpp
AndreaLanfranchi: Linters didn't catch it and there's no rule for spacing `if` before `(`
Anyway I'm fixing it.

client/catapult/tests/int/node/basic/LocalNodeBasicTests.cpp
AndreaLanfranchi: Not sure what you mean: `ShutdownIsIdempotentAndCanBeCalledMultipleTimes` behaves the same regardless the os. Nothing is this PR is platform specific

Jaguar0625: i mean **prior** to this PR, i'd expect the test to fail (so it is verifying this change).

AndreaLanfranchi: Correct. Prior to this PR, `shutdown()` called `saveStateToDisk()` even on an unbooted node, and could be called multiple times. On Windows, the double `saveStateToDisk()` caused intermittent failures due to delayed file handle release. PR #2071 addressed the underlying file handle issue; this PR removes the double call at the source.

PR #2071, opened 2026-07-01
[client/catapult] Retry RocksDB/state-dir ops on Windows delayed handle release

### LINKED ISSUE #2045: [client/catapult] Windows' delayed file handle release prevents reopen RocksDB
Still related to [HappyBlockchainIntegrityTests](https://github.com/symbol/symbol/issues/616) investigations.

The test performs a [boot / shutdown / reboot in quick succession](https://github.com/symbol/symbol/blob/dev/client/catapult/tests/int/node/stress/HappyBlockchainIntegrityTests.cpp#L361-L373) of LocalNodes.

This causes RocksDB to be opened, closed and almost immediately re-opened when Windows have not released yet its file handles causing an exception for DENIED ACCESS on rename operations. This causes the whole test to fail.

Solution: introduce retry logic inside [`RocksDatabase::RocksDatabase(const RocksDatabaseSettings& settings)`](https://github.com/symbol/symbol/blob/dev/client/catapult/src/catapult/cache_db/RocksDatabase.cpp#L156) constructor.

Workaround : on Windows always provide some delay over storage re-access operations

AndreaLanfranchi:
## Summary
Fixes #2045 — Windows delays releasing file handles after close, causing RocksDB reopen (and the paired state-directory rename in `LocalNodeStateSerializer::moveTo`) to fail with ACCESS_DENIED during rapid boot/shutdown/reboot cycles.

- Retry with exponential backoff on Windows only, gated to the specific transient error (`IOError` for RocksDB open, `permission_denied` for the rename) — not a blanket retry-on-any-failure.
- POSIX path unchanged in behavior (single attempt, no backoff) — the underlying OS quirk doesn't exist there.
- Retry loop extracted into a generic, injectable `utils::RetryWithBackoff` helper so the gating logic is unit-testable without needing real permission-denied filesystem/DB state.

## Test plan
- [x] New `RetryUtilsTests` cover the generic helper: success-first-try, non-retryable error stops immediately, retries-until-success, exhausts-attempts, `numAttempts=1` never retries.
- [x] New `RocksDatabaseTests.IsRetryableAfterFailedOpen_ReturnsTrueOnlyForIoError` covers the RocksDB-specific gate against `IOError`/`OK`/`InvalidArgument`/`NotFound`/`Corruption`/`NotSupported`.
- [x] Existing `RocksDatabaseTests` (35/35) and `LocalNodeStateFileStorageTests` (15/15) pass unchanged.
- [x] Local structure/style lint clean (0 violations).

### DIFF
--- a/client/catapult/src/catapult/cache_db/RocksDatabase.cpp
+++ b/client/catapult/src/catapult/cache_db/RocksDatabase.cpp
@@ -187,7 +187,6 @@ namespace catapult { namespace cache {
 #endif
 		auto status = utils::RetryWithBackoff(
 				[this, &dbOptions, &columnFamilies, &pDb]() {
-					m_handles.clear();
 					return rocksdb::DB::Open(dbOptions, m_settings.DatabaseDirectory, columnFamilies, &m_handles, &pDb);
 				},
 				IsRetryableAfterFailedOpen,
--- a/client/catapult/src/catapult/cache_db/RocksDatabase.cpp
+++ b/client/catapult/src/catapult/cache_db/RocksDatabase.cpp
@@ -195,7 +195,7 @@ namespace catapult { namespace cache {
 #endif
 		m_pDb.reset(pDb);
 		if (!status.ok())
-			CATAPULT_THROW_RUNTIME_ERROR_2("couldn't open database", m_settings.DatabaseDirectory, status.ToString());
+			CATAPULT_THROW_RUNTIME_ERROR_1(("couldn't open database: " + status.ToString()).c_str(), m_settings.DatabaseDirectory);
 	}
 
 	RocksDatabase::~RocksDatabase() {
--- a/client/catapult/src/catapult/cache_db/RocksDatabase.cpp
+++ b/client/catapult/src/catapult/cache_db/RocksDatabase.cpp
@@ -23,12 +23,12 @@
 #include "RocksInclude.h"
 #include "RocksPruningFilter.h"
 #include "catapult/config/CatapultDataDirectory.h"
-#include "catapult/exceptions.h"
 #include "catapult/utils/HexFormatter.h"
 #include "catapult/utils/Logging.h"
 #include "catapult/utils/PathUtils.h"
 #include "catapult/utils/RetryUtils.h"
 #include "catapult/utils/StackLogger.h"
+#include "catapult/exceptions.h"
 #include <chrono>
 #include <thread>
 
@@ -155,6 +155,10 @@ namespace catapult { namespace cache {
 		}
 	}
 
+	bool IsRetryableAfterFailedOpen(const rocksdb::Status& status) {
+		return !status.ok() && status.IsIOError();
+	}
+
 	RocksDatabase::RocksDatabase() = default;
 
 	RocksDatabase::RocksDatabase(const RocksDatabaseSettings& settings)
@@ -186,7 +190,7 @@ namespace catapult { namespace cache {
 					m_handles.clear();
 					return rocksdb::DB::Open(dbOptions, m_settings.DatabaseDirectory, columnFamilies, &m_handles, &pDb);
 				},
-				[](const rocksdb::Status& openStatus) { return !openStatus.ok() && openStatus.IsIOError(); },
+				IsRetryableAfterFailedOpen,
 				Num_Open_Attempts,
 				[Num_Open_Attempts](uint32_t attempt, const rocksdb::Status& openStatus) {
 					auto delayMs = 100u << attempt;
--- a/client/catapult/src/catapult/cache_db/RocksDatabase.cpp
+++ b/client/catapult/src/catapult/cache_db/RocksDatabase.cpp
@@ -200,7 +200,7 @@ namespace catapult { namespace cache {
 				});
 		m_pDb.reset(pDb);
 		if (!status.ok())
-			CATAPULT_THROW_RUNTIME_ERROR_1(("couldn't open database: " + status.ToString()).c_str(), m_settings.DatabaseDirectory);
+			CATAPULT_THROW_RUNTIME_ERROR_2("couldn't open database", m_settings.DatabaseDirectory, status.ToString());
 	}
 
 	RocksDatabase::~RocksDatabase() {
--- a/client/catapult/src/catapult/cache_db/RocksDatabase.cpp
+++ b/client/catapult/src/catapult/cache_db/RocksDatabase.cpp
@@ -27,10 +27,10 @@
 #include "catapult/utils/HexFormatter.h"
 #include "catapult/utils/Logging.h"
 #include "catapult/utils/PathUtils.h"
+#include "catapult/utils/RetryUtils.h"
 #include "catapult/utils/StackLogger.h"
-#ifdef _WIN32
+#include <chrono>
 #include <thread>
-#endif
 
 namespace catapult { namespace cache {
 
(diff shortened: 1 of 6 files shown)


client/catapult/src/catapult/cache_db/RocksDatabase.cpp
```diff
@@ -169,10 +177,31 @@ namespace catapult { namespace cache {
...
+		// On Windows, the OS may hold file handles open briefly after a prior instance closes,
+		// causing RocksDB's internal CURRENT rename to fail with ACCESS_DENIED. Retry with backoff.
+#ifdef _WIN32
+		constexpr uint32_t Num_Open_Attempts = 5;
+#else
+		constexpr uint32_t Num_Open_Attempts = 1;
+#endif
+		auto status = utils::RetryWithBackoff(
+				[this, &dbOptions, &columnFamilies, &pDb]() {
+					m_handles.clear();
```
Jaguar0625: is there scenario where m_handles is not empty? would avoid clear since items should be cleaned with DestroyColumnFamilyHandle

client/catapult/src/catapult/cache_db/RocksDatabase.cpp
```diff
@@ -169,10 +177,31 @@ namespace catapult { namespace cache {
...
+					auto delayMs = 100u << attempt;
+					CATAPULT_LOG(warning)
+							<< "RocksDB open failed (attempt " << (attempt + 1) << "/" << Num_Open_Attempts << "): "
+							<< openStatus.ToString() << ", retrying in " << delayMs << "ms";
+					std::this_thread::sleep_for(std::chrono::milliseconds(delayMs));
+				});
 		m_pDb.reset(pDb);
 		if (!status.ok())
-			CATAPULT_THROW_RUNTIME_ERROR_2("couldn't open database", m_settings.DatabaseDirectory, status.ToString());
+			CATAPULT_THROW_RUNTIME_ERROR_1(("couldn't open database: " + status.ToString()).c_str(), m_settings.DatabaseDirectory);
```
Jaguar0625: why change this? we generally try to avoid allocations in `throw` statements where possible

client/catapult/src/catapult/extensions/LocalNodeStateFileStorage.cpp
```diff
@@ -193,7 +199,38 @@ namespace catapult { namespace extensions {
...
+		// On Windows a transient open handle on the destination - delete-pending state files, antivirus,
+		// or the search indexer - can make the rename fail with ACCESS_DENIED even when it would succeed on
+		// POSIX (which frees the name on unlink). Retry with backoff before giving up; otherwise the
+		// filesystem_error escapes the block dispatcher consumer and terminates the process.
+		// (cf. RocksDatabase open-retry on Windows.)
+#ifdef _WIN32
+		constexpr uint32_t Num_Rename_Attempts = 5;
+#else
+		constexpr uint32_t Num_Rename_Attempts = 1;
+#endif
```
Jaguar0625, line 213: consider moving this constant to RetryUtils.h as something a bit more generic, e.g. `Max_File_IO_Retry_Count` since you're using same values in both places

client/catapult/src/catapult/utils/RetryUtils.h
```diff
@@ -0,0 +1,41 @@
...
+#include <stdint.h>
+
+namespace catapult { namespace utils {
+
+	/// Runs \a operation up to \a numAttempts times, retrying only while \a shouldRetry returns \c true
+	/// for the most recent result and further attempts remain. \a onRetry is invoked with the zero-based
+	/// attempt index and the result before every retry; it is expected to apply backoff.
+	/// \note \a shouldRetry must return \c false for a result that indicates success.
+	template<typename TOperation, typename TShouldRetry, typename TOnRetry>
+	auto RetryWithBackoff(TOperation operation, TShouldRetry shouldRetry, uint32_t numAttempts, TOnRetry onRetry) {
```
Jaguar0625: name is bit misleading since there is no backoff logic in here. that happens in on retry. perhaps have this function do the calculation at least; if not the wait.

client/catapult/tests/catapult/utils/RetryUtilsTests.cpp
```diff
@@ -0,0 +1,151 @@
...
+		auto operation = [&context]() {
+			++context.NumOperationCalls;
+			return Success();
+		};
+
+		// Act:
+		auto result = RetryWithBackoff(operation, IsRetryable, 5u, context.onRetry());
+
+		// Assert:
+		EXPECT_FALSE(result);
```
Jaguar0625: works due to error_code implicit bool operator but would prefer more explicit: `EXPECT_EQ(SUCCESS, std::error_code)`

client/catapult/tests/catapult/utils/RetryUtilsTests.cpp
```diff
@@ -0,0 +1,151 @@
...
+		EXPECT_EQ(2u, context.RetryCalls.size());
+	}
+
+	TEST(TEST_CLASS, NeverRetries_WhenNumAttemptsIsOne) {
+		// Arrange:
+		RetryContext context;
+		auto operation = [&context]() {
+			++context.NumOperationCalls;
+			return RetryableError();
+		};
```
Jaguar0625: consider helper function like `MakeOperation(context, RetryableError())`

Jaguar0625 (commented):
👍

client/catapult/src/catapult/cache_db/RocksDatabase.cpp
AndreaLanfranchi: Actually `DestroyColumnFamilyHandle` is called only on db dtor, and I was unsure whether a failed open was leaving the vector dirty. Checked `DBImpl::Open` (rocksdb 10.6.2) — it clears `handles` itself as its first action, and on any failure it deletes every handle it created and clears the vector again before returning. So this `m_handles.clear()` never does anything: the vector is always empty going in and coming out of a failed `Open`. Dropping it, thanks.

client/catapult/src/catapult/cache_db/RocksDatabase.cpp
AndreaLanfranchi: Fair, no strong reason to break that convention here. `CATAPULT_THROW_RUNTIME_ERROR_2` already attaches both params without the extra concat. Reverting to the original two arg form.

client/catapult/src/catapult/extensions/LocalNodeStateFileStorage.cpp
AndreaLanfranchi: `RetryUtils.h` should stay pure mechanism, no domain policy. Attempt count is tied to each operation's failure window, not to "retry with backoff" itself. Regardless of the specific case at hand, this utility could be used for other retriable operations in their own domains. Both call sites use 5/1 today, but that is coincidence, not a shared invariant. If one op's window ever needs to diverge, a shared constant would silently couple them. Keeping `Num_Open_Attempts` and `Num_Rename_Attempts` local seems more honest. Open to it if you think they really are the same policy.

client/catapult/src/catapult/utils/RetryUtils.h
AndreaLanfranchi: I concur the name is misleading about the function body only, but the logic is built up with a combination of lambdas. However, I will move the delay calculation and sleep into the helper itself, with a `baseDelayMs` parameter (default 100, guarded so 0 skips the wait) so tests stay instant without mocking. Callers' `onRetry` becomes logging only.

client/catapult/tests/catapult/utils/RetryUtilsTests.cpp
AndreaLanfranchi: Done as part of the backoff refactor. Both spots now use `EXPECT_EQ(std::error_code(), result)`.

client/catapult/tests/catapult/utils/RetryUtilsTests.cpp
AndreaLanfranchi: Ok will implement.

Jaguar0625 (approved):
👍

PR #2072, opened 2026-07-01
[client/catapult] fix: null dereference in BlockchainSyncConsumer.cpp

### LINKED ISSUE #2015: [client/catapult] Null dereference in BlockchainSyncConsumer.cpp
The [code here](https://github.com/symbol/symbol/blob/dev/client/catapult/src/catapult/consumers/BlockchainSyncConsumer.cpp#L338-L344) is not guarded against a possibly empty HashRange.

At node startup the network finalized height is 0 so the actual active test is false.
As a result loadHashesFrom(0, 1) returns an empty range.
However the code doesn't check emptyness and tries to deref the first element
```
return heightHashPair.Hash == *storageHashRange.cbegin();
```

On Debug builds this throws an assertion. On Release is UB.

Wrokin on it

AndreaLanfranchi:
Fixes #2015

At node startup network finalized height is Height(0), so the height guard was false and code called loadHashesFrom(Height(0), 1), returning empty HashRange. Dereferencing its value-initialized iterator tripped MSVC debug assertion, UB in release.

## Summary
- Guard empty hash range before dereferencing (990a49d3c)
- Remove now-redundant height check (f862bd5c8)
- Collapse empty-check into single return expression (dd32aa5cb)

### DIFF
--- a/client/catapult/src/catapult/consumers/BlockchainSyncConsumer.cpp
+++ b/client/catapult/src/catapult/consumers/BlockchainSyncConsumer.cpp
@@ -336,10 +336,13 @@ namespace catapult { namespace consumers {
 
 		private:
 			static bool Contains(const io::BlockStorageView& storageView, const model::HeightHashPair& heightHashPair) {
-				if (storageView.chainHeight() < heightHashPair.Height)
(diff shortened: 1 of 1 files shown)


Jaguar0625 (approved):
👍

PR #2074, opened 2026-07-02
[client/catapult] fix: compile missing mongo plugin test/src files (#1992)

### LINKED ISSUE #1992: [client/catapult] Test cpp file never compiled
In 
symbol/client/catapult/extensions/mongo/plugins/lock_secret/tests/CMakeLists.txt

We have
```
set(TARGET_NAME tests.catapult.mongo.plugins.locksecret)

include_directories(${PROJECT_SOURCE_DIR}/plugins/txes/lock_secret)
catapult_mongo_plugin_tests_no_deps(${TARGET_NAME} mappers)

add_subdirectory(int)
add_subdirectory(test)
```

This call
```
catapult_mongo_plugin_tests_no_deps(${TARGET_NAME} mappers)
```

passes the target name and the directory where to grab sources from (`mappers`).
The callee though, defined [here](https://github.com/symbol/symbol/blob/dev/client/catapult/extensions/mongo/plugins/CMakeLists.txt#L26-L31),  has this body
```
function(catapult_mongo_plugin_tests_no_deps TARGET_NAME)
	catapult_test_executable_target(${TARGET_NAME} mongo)
	catapult_add_mongo_dependencies(${TARGET_NAME})

	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}" PARENT_SCOPE)
endfunction()
```

The incoming source dir (mappers) is not forwarded to called functions and the result is the cpp file into `mappers` is never compiled.

AndreaLanfranchi:
Fixes #1992

catapult_mongo_plugin_tests_no_deps dropped extra source dir args, so lock_secret and lock_hash mappers cpp files were never compiled.

Merged the no deps / with deps helper pairs, both for tests and for src, into one helper each:

Tests: catapult_mongo_plugin_tests now decides deps linkage by checking if TARGET catapult.mongo.plugins.<name>.deps exists, and always forwards extra source dir args to the underlying glob.

Src: catapult_mongo_plugin_src now decides deps linkage by checking if a storages subdir exists (that is where CacheStorage code lives, and it is the reason the deps lib is needed).

## Test plan

Full local build passes. Full local test suite passes, 87/87 (mongod active for int tests).

### DIFF
--- a/client/catapult/extensions/mongo/plugins/CMakeLists.txt
+++ b/client/catapult/extensions/mongo/plugins/CMakeLists.txt
@@ -2,55 +2,42 @@ cmake_minimum_required(VERSION 3.23)
 
 set(PLUGIN_CATAPULT_LIBS catapult.mongo)
 
-function(catapult_mongo_plugin_src_no_deps PLUGIN_BASE_NAME)
-	catapult_shared_library_target(${PLUGIN_BASE_NAME} "${ARGN}")
-	catapult_add_mongo_dependencies(${PLUGIN_BASE_NAME})
-	target_link_libraries(${PLUGIN_BASE_NAME} ${PLUGIN_CATAPULT_LIBS})
+# a plugin needs a deps lib when it has a cache storage (living under a "storages" subdir);
+# the cache storage needs access to the plugin's core cache, hence the extra .deps link target
+function(catapult_mongo_plugin_src PLUGIN_BASE_NAME)
+	if(IS_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/storages")
+		# create a deps lib
+		catapult_library_target(${PLUGIN_BASE_NAME}.deps "${ARGN}")
+		catapult_add_mongo_dependencies(${PLUGIN_BASE_NAME}.deps)
+		target_link_libraries(${PLUGIN_BASE_NAME}.deps ${PLUGIN_CATAPULT_LIBS})
+
+		# create a plugin dll
+		catapult_shared_library_target(${PLUGIN_BASE_NAME})
+		target_link_libraries(${PLUGIN_BASE_NAME} ${PLUGIN_BASE_NAME}.deps)
+	else()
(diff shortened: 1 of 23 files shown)


Jaguar0625:
Change looks good, but confused about this comment:

> catapult_mongo_plugin_tests_no_deps dropped extra source dir args, so lock_secret and lock_hash mappers cpp files were never compiled.


I can see them being build in [dev builds](https://jenkins.symbolsyndicate.us/job/Symbol/job/generated/job/symbol/job/client-catapult/job/dev/1702/consoleFull). 
Is this some MSVC quirk?

AndreaLanfranchi:
Not MSVC specific. Checked the dev build log you linked: the files compiled there are HashLockMapperTests.cpp and SecretLockMapperTests.cpp (top level, always built).

The actually affected files are named differently and live under tests/mappers/: HashLockInfoMapperTests.cpp and SecretLockInfoMapperTests.cpp. Searched the full dev build log, zero occurrences of either filename, confirming they were never compiled on dev.

On this PR's build they now show up:
```
[1824/2153] Building CXX object .../lock_hash/tests/CMakeFiles/tests.catapult.mongo.plugins.lockhash.dir/mappers/HashLockInfoMapperTests.cpp.o
[1831/2153] Building CXX object .../lock_secret/tests/CMakeFiles/tests.catapult.mongo.plugins.locksecret.dir/mappers/SecretLockInfoMapperTests.cpp.o
```
and their gtest suites run:
```
[----------] 1 test from HashLockInfoMapperTests
[----------] 1 test from SecretLockInfoMapperTests
```

Root cause: catapult_mongo_plugin_tests_no_deps never forwarded its extra source dir argument (mappers) down to catapult_test_executable_target, so the glob step never scanned that subdir. Similarly named sibling files at the top level masked the gap, easy to miss on a quick look at the build log.

Jaguar0625 (approved):
👍

PR #2075, opened 2026-07-02
[client/catapult] fix: close TLS sockets gracefully to stop Windows RST storm (#2060)

### LINKED ISSUE #2060: [client/catapult] Abortive TLS closure causes RST storm on Windows
## Observable symptoms
On Windows, `HappyBlockchainIntegrityTests.MultiNodeDenseNetworkCanReachConsensus_*` intermittently fails: after the test reboots all nodes, those lagging behind could not reach the best-chain score within the 45 seconds window. The trigger is an abrupt TLS socket close that makes Winsock emit an abortive **RST** instead of a FIN. This causes a self-amplifying connection-reset storm so the network spends its time resetting and re-handshaking connections instead of syncing.

## The defect

[`SocketGuard`](https://github.com/symbol/symbol/blob/dev/client/catapult/src/catapult/ionet/PacketSocket.cpp#L78) is the per-socket close coordinator which implements two close patterns both ending with a hard-close of the lowest layer:

- [`abortImpl`](https://github.com/symbol/symbol/blob/dev/client/catapult/src/catapult/ionet/PacketSocket.cpp#L154) (immediate hard close) ends in `lowest_layer().close()`.
- the _graceful_ branch [`closeImpl`](https://github.com/symbol/symbol/blob/dev/client/catapult/src/catapult/ionet/PacketSocket.cpp#L125) issues an SSL `async_shutdown` **and** a 1-byte `async_write`, then hard-closes the lowest layer **as soon as the write completes** but it never completes the shutdown handshake and never drains the receive buffer.

On Windows, calling `closesocket()` while unread bytes are still sitting in the receive buffer makes Winsock send an **RST** (abortive close), not a FIN.

## The storm building up - Walkthrough

Each node keeps two kinds of connections:
- a **writer pool** (`net/PacketWriters.cpp`) : outbound connections used to send sync requests (chain-statistics, block requests);
- **readers** (`net/PacketReaders.cpp` + `net/ChainedSocketReader.cpp`) : inbound accepted connections that serve requests.

Because every node in the test is a symmetric peer, every node do initiate closes **and** suffers its neighbours' closes.
Here's how :
1. **Intentional closure** A connection is intentionally closed for a legit reason which can be either [aging](https://github.com/symbol/symbol/blob/dev/client/catapult/src/catapult/extensions/PeersConnectionTasks.cpp#L196-L215) 
or previous [sync failure](https://github.com/symbol/symbol/blob/dev/client/catapult/src/catapult/net/PacketWriters.cpp#L398-L401). 
2. **Abortive closure** As the close is abortive `pSocket->close()` runs through `SocketGuard`, which hard-closes the lowest layer with unread RX data buffered -> **Winsock RST** (see "The defect" above).
3. **The RST hits peers two ways**
	
*a) Make in flight sync writes to fail*

When a peer is in the middle of a request on its writer connection. The request is sent via [`api::RemoteRequestDispatcher::send`](https://github.com/symbol/symbol/blob/dev/client/catapult/src/catapult/api/RemoteRequestDispatcher.h#L61-L63)

```cpp
m_io.write(packetPayload, [...](auto code) {
    if (ionet::SocketOperationCode::Success != code)
        return callback(RemoteChainResult::Write_Error, ResultType());
    [...]
```
The RST makes that write fail with `WSAECONNRESET`, which is treated as `Write_Error` and logged as **"write to remote node failed"**. On such error the error-handler created by [`PacketWriters::pickOne`](https://github.com/symbol/symbol/blob/dev/client/catapult/src/catapult/net/PacketWriters.cpp#L330-L332) fires and calls `pThis->removeWriter()` which implies **another RST**. [Broadcast writes](https://github.com/symbol/symbol/blob/dev/client/catapult/src/catapult/net/PacketWriters.cpp#L309-L319) have the same pattern.

*b) It kills the peer's reader.*

On the inbound side the RST makes the reader's `async_read_some` fail with `WSAECONNRESET`.
[`mapReadErrorCodeToSocketOperationCode`](https://github.com/symbol/symbol/blob/dev/client/catapult/src/catapult/ionet/PacketSocket.cpp#L36-L47) maps this to `Read_Error` (only a clean EOF maps to `Closed`). [`ChainedSocketReader`](https://github.com/symbol/symbol/blob/dev/client/catapult/src/catapult/net/ChainedSocketReader.cpp#L60-L74)
 logs "read completed with error" and invokes the completion handler. In [`PacketReaders::createReader`](https://github.com/symbol/symbol/blob/dev/client/catapult/src/catapult/net/PacketReaders.cpp#L294-L301)

 ```cpp
if (ionet::SocketOperationCode::Closed == code)
    pThis->m_readers.close(identity, id);   // clean close: drop just this socket
else
    pThis->m_readers.close(identity);       // error: drop ALL sockets for this peer
```

This is a hidden amplifier: an **abortive** read error closes **every** connection to
that identity, whereas a clean EOF would close only the one socket. So one RST can tear
down several reader connections at once each of which hard-closes and emits **more RSTs**.

4. **Immediate rebuild** : The dropped connections are rebuilt right away: the
[connect-peers task](https://github.com/symbol/symbol/blob/dev/client/catapult/src/catapult/extensions/PeersConnectionTasks.cpp#L196-L215) reconnects missing peers on its interval, and `removeWriter` frees the pool slot so the next sync reconnects. The test's `MakeTaskConfigurationFast` shortens these intervals, so rebuilds are fast.
**Each rebuild is a full TLS handshake**, and the freshly rebuilt connection is immediately used for the next sync write which races into another peer that is itself mid-close hence another RST.

5. **Amplifier gain > 1.** Each abortive close causes, on average, more than one further
abortive close. In a test log observation the ~1,400 benign close amplified into ~5,000+ RSTs in the 45 s
window; ~86% of all connection deaths were abortive rather than graceful. Lagging nodes were dealing
most of the time with connections closure and reopening so they could never catch up within the time window.

## Why Linux never storms
The identical dirty-close code runs on Linux, and Linux *also* sends an RST on close-with-unread-data. The difference is *when the local write observes the peer's RST*. On Linux the sync write's bytes are typically accepted into the kernel send buffer and the write completes locally **before** the peer's RST is processed so step (3a) never marks the write failed, `removeWriter` never fires, and the feedback loop never ignites. On Windows (IOCP) a pending `WSASend` is failed immediately when the RST lands, so the write *does* fail and the cascade starts. 

## Conclusions

An abortive socket closure is not wrong in principle: it can be implemented as a defensive strategy to avoid piling up TIME_WAIT sockets on systems under heavy load or to immediately drop misbehaving peers. What is wrong here is that abortive pattern is applied **to normal, healthy closes** that should have been orderly. Using abrtive closure for regular teardown is the bug.

Why this is wrong :

- **TLS layer:** skipping `close_notify` ends the secure stream with a truncation the peer cannot distinguish from a truncation attack. Sloppy everywhere, not just on Windows.
- **TCP layer:** closing with unread RX data turns a "goodbye" (FIN) into an "error/abort" (RST) and discards already-delivered bytes. Also everywhere.

As stated before Linux is not immune : it simply doesn't have the trigger that ignites it. Windows have.

## The fix

Ref [this commit](https://github.com/symbol/symbol/commit/faecb410e56109fffa1544298a1bf37d7657601d).

Make `SocketGuard::closeImpl` perform a graceful **drained** close:

1. cancel pending ops;
2. `async_shutdown` send our SSL `close_notify` and consume the peer's; this also  drains buffered receive records so the peer sees an orderly shutdown rather than a truncation;
3. drain any remaining received bytes to EOF on the raw next layer;
4. close the lowest layer;

all bounded by a short (100 ms) `steady_timer` so teardown stays prompt if the peer stalls. `abort()` is unchanged (immediate hard close) for the already-closed / timeout paths.

**Windows (native, MSVC RelWithDebInfo):**
- Unit tests: `ionet` 565/565, `net` 138/138 (including `ReaderCanBeStopped`). 
- `HappyBlockchainIntegrityTests` dense soak: **265+ iterations over 13.5 h, zero  failures** (randomized per-node heights).
- Sample-window metrics: write failures ~1528 -> ~29; abortive reader closes collapsed correspondingly; `stream truncated` reads -> 0.
- Passing-run wall time fell from ~5-6 min to ~3 min (Linux-like) as the TLS-handshake reconnect churn disappeared. (Even pre-fix runs that *passed* were passing marginally, paying that churn tax the whole time.)

**Linux (WSL2):**
- `HappyBlockchainIntegrityTests` soak: **257 iterations over 6 h 31 m, zero failures**   Confirms the graceful close (epoll path) does not regress readers-shutdown stability. 
- Passing-run wall time fell from ~2 min to ~1 min 30 second so a benefit also on Linux is recorded.

### PR DESCRIPTION
**Pull request by AndreaLanfranchi**
## Summary
- Ports the graceful drained TLS close from `experimental` (faecb410e) to fix #2060: abortive closes on Windows send RST instead of FIN when unread RX data is buffered, triggering a reset/reconnect storm that starves `HappyBlockchainIntegrityTests` of sync time.
- `SocketGuard::closeImpl` now completes the SSL shutdown handshake, drains remaining RX bytes to EOF, then hard-closes — bounded by a 100ms timer.
- Includes the two companion test-helper fixes from `experimental` (tolerate `Closed`/`Read_Error` race in `PacketWritersTests`/`ChainedSocketReaderTests`), a lint fix for two comment typos, and a small conciseness cleanup in `SocketGuard`.

## Test plan
- [x] Local build (Debug) green
- [x] `tests.catapult.ionet` 566/566 passing
- [x] `tests.catapult.net` 138/138 passing
- [x] Local lint (`checkProjectStructure.py`) 0 violations
- [x] CI green

### DIFF
--- a/client/catapult/src/catapult/ionet/PacketSocket.cpp
+++ b/client/catapult/src/catapult/ionet/PacketSocket.cpp
@@ -84,7 +84,8 @@ namespace catapult { namespace ionet {
 					: m_strand(boost::asio::make_strand(ioContext))
 					, m_strandWrapper(m_strand)
 					, m_socket(ioContext, sslContext)
-					, m_drainTimer(ioContext) {
+					, m_drainTimer(ioContext)
+					, m_isFinalizing(false) {
 				m_isClosed.clear();
 				m_isClosed.test_and_set();
 			}
@@ -93,7 +94,8 @@ namespace catapult { namespace ionet {
 					: m_strand(boost::asio::make_strand(ioContext))
 					, m_strandWrapper(m_strand)
 					, m_socket(std::move(socket), sslContext)
-					, m_drainTimer(ioContext) {
+					, m_drainTimer(ioContext)
+					, m_isFinalizing(false) {
 				m_isClosed.clear();
 				m_isClosed.test_and_set();
 			}
@@ -217,7 +219,7 @@ namespace catapult { namespace ionet {
 			Socket m_socket;
 			boost::asio::steady_timer m_drainTimer;
 			std::array<uint8_t, 1024> m_drainBuffer;
-			std::atomic_bool m_isFinalizing = false;
+			std::atomic_bool m_isFinalizing;
 			std::atomic_flag m_isClosed;
 		};
 
@@ -571,6 +573,7 @@ namespace catapult { namespace ionet {
 					: m_strandWrapper(pSocketGuard->strand())
 					, m_socket(pSocketGuard, options, *this)
 					, m_id(s_idCounter.fetch_add(1))
+					, m_isClosePending(false)
 			{}
 
 			~StrandedPacketSocket() override {
@@ -693,7 +696,7 @@ namespace catapult { namespace ionet {
 			thread::StrandOwnerLifetimeExtender<StrandedPacketSocket> m_strandWrapper;
 			SocketType m_socket;
 			SocketIdentifier m_id;
-			std::atomic<bool> m_isClosePending{ false };
+			std::atomic<bool> m_isClosePending;
 		};
 
 		std::atomic<uint64_t> StrandedPacketSocket::s_idCounter(1);
--- a/client/catapult/src/catapult/ionet/PacketSocket.cpp
+++ b/client/catapult/src/catapult/ionet/PacketSocket.cpp
@@ -139,7 +139,7 @@ namespace catapult { namespace ionet {
 				// peer's. The subsequent raw drain empties whatever is still buffered. On Windows, closing a
 				// socket with unread data buffered makes Winsock emit an abortive RST instead of a FIN; that
 				// RST fails peers' in-flight writes (WSAECONNRESET) and drives a reset/reconnect storm.
-				// The whole sequence is bounded by a short timer so teardown stays prompt if the peer stalls.
+				// This whole sequence is bounded by a short timer so teardown stays prompt if the peer stalls.
 				boost::system::error_code ignoredEc;
 				m_socket.lowest_layer().cancel(ignoredEc);
 
@@ -375,8 +375,8 @@ namespace catapult { namespace ionet {
 
 			void readSome(const PacketSocket::ReadCallback& callback, bool allowMultiple) {
 				// once a close has been requested, do not issue another socket read: closeImpl runs the SSL
-				// shutdown handshake and a receive drain on the same stream, and a concurrent async_read_some
-				// would both race that teardown and deliver a stale packet after the owner asked to stop.
+				// shutdown handshake and a receive drain on the same stream; a concurrent async_read_some
+				// would race that teardown and could deliver a stale packet after the owner asked to stop.
 				if (m_wrapper.isClosePending())
 					return callback(SocketOperationCode::Read_Error, nullptr);
 
--- a/client/catapult/src/catapult/ionet/PacketSocket.cpp
+++ b/client/catapult/src/catapult/ionet/PacketSocket.cpp
@@ -84,8 +84,7 @@ namespace catapult { namespace ionet {
 					: m_strand(boost::asio::make_strand(ioContext))
 					, m_strandWrapper(m_strand)
 					, m_socket(ioContext, sslContext)
-					, m_drainTimer(ioContext)
-					, m_isFinalizing(false) {
+					, m_drainTimer(ioContext) {
 				m_isClosed.clear();
 				m_isClosed.test_and_set();
 			}
@@ -94,8 +93,7 @@ namespace catapult { namespace ionet {
 					: m_strand(boost::asio::make_strand(ioContext))
 					, m_strandWrapper(m_strand)
 					, m_socket(std::move(socket), sslContext)
-					, m_drainTimer(ioContext)
-					, m_isFinalizing(false) {
+					, m_drainTimer(ioContext) {
 				m_isClosed.clear();
 				m_isClosed.test_and_set();
 			}
@@ -214,13 +212,12 @@ namespace catapult { namespace ionet {
 		private:
 			static constexpr auto Drain_Timeout = std::chrono::milliseconds(100);
 
-		private:
 			Strand m_strand;
 			thread::StrandOwnerLifetimeExtender<SocketGuard> m_strandWrapper;
 			Socket m_socket;
 			boost::asio::steady_timer m_drainTimer;
 			std::array<uint8_t, 1024> m_drainBuffer;
-			bool m_isFinalizing;
+			bool m_isFinalizing = false;
 			std::atomic_flag m_isClosed;
 		};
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: client/catapult/src/catapult/ionet/PacketSocket.cpp [hunk 1]
```diff
@@ -172,10 +210,14 @@ namespace catapult { namespace ionet {
...
 		private:
+			static constexpr auto Drain_Timeout = std::chrono::milliseconds(100);
+
 			Strand m_strand;
 			thread::StrandOwnerLifetimeExtender<SocketGuard> m_strandWrapper;
 			Socket m_socket;
-			uint8_t m_sentinelByte;
+			boost::asio::steady_timer m_drainTimer;
+			std::array<uint8_t, 1024> m_drainBuffer;
+			std::atomic_bool m_isFinalizing = false;
```
Jaguar0625: please set initial member values in initializer lists  for consistency with rest of codebase here and below. while correct, it's a bit hard to read having some variables initialized in initialized list and others where declared.

client/catapult/src/catapult/ionet/PacketSocket.cpp
```diff
@@ -561,17 +609,25 @@ namespace catapult { namespace ionet {
 			}
 
 			void close() override {
+				// set synchronously (before the posted handler runs) so an in-flight reader stops
+				// re-issuing socket reads as soon as a close is requested
+				m_isClosePending = true;
```
Jaguar0625, line 617: just question: any reason you wanted to do this outside of the strand context?

Jaguar0625 (commented):
👍

client/catapult/src/catapult/ionet/PacketSocket.cpp
AndreaLanfranchi: I will do but this causes unneeded repetitions that conflict with the conciseness requested elsewhere.

client/catapult/src/catapult/ionet/PacketSocket.cpp
AndreaLanfranchi: Deliberate. close()/abort() can be called from any thread. postCloseOnce dispatches the actual close onto the strand, but that dispatch is async, so a caller could call close() then readSome() before the posted handler runs. Setting the flag synchronously before the post closes that window.

client/catapult/src/catapult/ionet/PacketSocket.cpp
AndreaLanfranchi: I personally find way more difficult to follow the interleaving of public / private scopes within a class declaration. Imo there should be 1 public region and 1 private.

Jaguar0625: > I will do but this causes unneeded repetitions that conflict with the conciseness requested elsewhere.

i agree, but there is also value in having consistency across the codebase that is almost 10 years old now. 

> I personally find way more difficult to follow the interleaving of public / private scopes within a class declaration. Imo there should be 1 public region and 1 private.

yes, PacketSocket went a bit overboard with that. eventually we landed on better compromise, where there are still multiple sections but for different types of symbols (one for variables, functions, static functions, etc).

Jaguar0625 (approved):
👍

PR #2078, opened 2026-07-03
[client/catapult] fix: remove unneeded lambda capture in RocksDatabase

AndreaLanfranchi:
## Summary
* `Num_Open_Attempts` is `constexpr`; capturing it in the retry lambda triggers clang's `-Wunused-lambda-capture` (`-Werror`), breaking clang builds.
* Removed the capture; the constant is still usable inside the lambda body without odr-use.

## Test plan
- [x] `cmake --build --preset conan-relwithdebinfo` no longer errors on this file


Jaguar0625 (approved):
👍

Issue #2081, opened 2026-07-04
[client/catapult] Missing docs for building and running with sanitizers (TSAN/ASAN/UBSan) locally


AndreaLanfranchi:
### Problem

`docs/README.md` mentions sanitizer suppressions (`docs/README.md:66` to `82`), but no build guide (`BUILD-docker.md`, `BUILD-conan.md`, `BUILD-manual.md`) explains how to actually produce a sanitizer build locally. Finding the path requires digging through `jenkins/catapult/*.py` and `jenkins/catapult/configurations/`.

Also easy to trip on a hidden constraint: sanitizer builds and conan builds are mutually exclusive, enforced in code, `jenkins/catapult/BasicBuildManager.py:19`:

```python
if self.sanitizers and self.use_conan:
    raise RuntimeError('sanitizer cannot be used with conan-based builds')
```

This applies to all sanitizer configs, not just thread sanitizer:
- `jenkins/catapult/configurations/amd64/clang-tsan.yaml` sets `sanitizers: thread`
- `jenkins/catapult/configurations/amd64/clang-ausan.yaml` sets `sanitizers: address,undefined`

Both only pair with `tests-metal.yaml` (`use_conan: false`), never `tests-conan.yaml`. Nothing in `docs/` states this, so anyone following `BUILD-conan.md` and then trying to add a sanitizer on top hits a confusing runtime error with no doc pointing at the real cause.

### Suggested fix

Add a short `docs/BUILD-sanitizers.md` (or a section in `docs/README.md`) covering:
1. Sanitizer builds use the metal/vcpkg path, not conan, and why (mutual exclusion in `BasicBuildManager.py`).
2. Local reproduction command, mirroring `scripts/ci/build.sh` but with sanitizer configs:

```bash
python3 jenkins/catapult/runDockerBuild.py \
  --compiler-configuration jenkins/catapult/configurations/amd64/clang-tsan.yaml \
  --build-configuration jenkins/catapult/configurations/tests-metal.yaml \
  --operating-system ubuntu \
  --user "$(id -u):$(id -g)" \
  --destination-image-label clang-tsan-local \
  --source-path "$(git rev-parse --show-toplevel)"
```

Swap `clang-tsan.yaml` for `clang-ausan.yaml` for ASAN/UBSan.

3. Link to existing suppressions list and `sanitizer_ignorelist.txt`.

AndreaLanfranchi:
[...]

PR #2086, opened 2026-07-06
[client/catapult] fix: warm up OpenSSL RAND_bytes cache before spawning test worker threads

### LINKED ISSUE #2080: [client/catapult] TSAN data race in OpenSSL RAND_bytes fetch cache during parallel signature verification
### Summary

CI run: https://github.com/symbol/symbol/runs/85098913171

TSAN flags a data race inside `libcrypto` (`hashtable.c: match_key` / `alloc_new_value`) during `DispatcherServiceTests.CanDisableSpamThrottling`. Race is between two `IoThreadPool` worker threads both calling `RAND_bytes` for the first time.

### Root cause

Race is not in catapult logic. It is triggered by OpenSSL's lazy algorithm fetch cache.

Call chain:
`DispatcherServiceTests_CanDisableSpamThrottling` boots dispatcher service, which runs `VerifyMulti` batch signature verification. `VerifyMulti` uses `ParallelFor` to spread work across the dispatcher's isolated `IoThreadPool`. Each worker thread invokes the random filler lambda from `DispatcherService.cpp` (`CreateRandomFiller`), which calls `SecureRandomGenerator::fill`, which calls `RAND_bytes` (`src/catapult/crypto/SecureRandomGenerator.cpp:44`).

The first `RAND_bytes` call per process triggers OpenSSL 3.x to lazily fetch the RAND algorithm from the loaded provider and populate an internal `OSSL_METHOD_STORE` hashtable. When two worker threads hit this uninitialized cache at the same time, the read/write on that hashtable is not properly synchronized under TSAN, producing the reported race in `memcmp`/`memcpy`.

`SetupOpensslCryptoFunctions` (`src/catapult/crypto/OpensslInit.cpp`) loads the default and legacy providers into the process global `OSSL_LIB_CTX` once in `TestMain.cpp` before tests run, but that only initializes the providers themselves, not the per algorithm fetch cache, which stays lazy until first use.

### Proposed fix

Warm up `RAND_bytes` once, single threaded, in `TestMain.cpp` right after `SetupOpensslCryptoFunctions()`, before `RUN_ALL_TESTS()`. This forces the fetch cache populate to happen serially, so later concurrent calls from worker threads only read from an already populated cache.

```cpp
std::cout << "Initializing OpenSSL crypto functions" << std::endl;
auto pOpensslContext = catapult::crypto::SetupOpensslCryptoFunctions();

{
	uint8_t warmupBuffer[16];
	catapult::crypto::SecureRandomGenerator().fill(warmupBuffer, sizeof(warmupBuffer));
}
```

Alternative or additional fix: check whether `Signer.cpp` batch verify path independently triggers other lazy EVP fetches (digest, pkey ops) that need the same warm up. Also worth checking current libcrypto build in CI image (BuildId `2914801d1b5322127d6f65ae9153093648e598de`), since later OpenSSL 3.x point releases hardened `OSSL_METHOD_STORE` locking and may fix this at the source.

### Stack trace

<details>
<summary>TSAN report</summary>

```
SUMMARY: ThreadSanitizer: data race (/usr/catapult/tests/tests.catapult.sync+0xe68ae) in memcmp
  Read of size 8 by thread T1040:
    #0 memcmp
    #1 match_key hashtable.c (libcrypto.so.3)
    #2 DispatcherService.cpp:73 (CreateRandomFiller lambda)
    ...
  Previous write of size 8 by thread T1042:
    #0 memcpy
    #1 alloc_new_value hashtable.c (libcrypto.so.3)
    #2 DispatcherService.cpp:73 (CreateRandomFiller lambda)
    ...
```

</details>

AndreaLanfranchi:
## Summary

Fixes #2080.

TSAN flagged a data race inside libcrypto's `OSSL_METHOD_STORE` hashtable during `DispatcherServiceTests.CanDisableSpamThrottling`. Two `IoThreadPool` worker threads called `RAND_bytes` for the first time concurrently during batch signature verification, racing on OpenSSL's lazy algorithm fetch cache.

This warms up `RAND_bytes` once, single threaded, in `TestMain.cpp` right after `SetupOpensslCryptoFunctions()`, before any worker pool spawns. Later concurrent calls from worker threads only read from an already populated cache.

## Test plan

- [x] Standard CI (lint, build, tests) green on `fix/2080`.
- [ ] TSAN run against this branch not yet confirmed. The sanitizer build (`clang-tsan` + `tests-metal`) is a separate manually/schedule triggered Jenkins job, not part of the default PR pipeline, and I don't have trigger permission on it. Opening this PR for review while that confirmation is pending.


AndreaLanfranchi:
Converted to draft as premature without a final word from #2081

PR #2092, opened 2026-07-07
[linters]: add docs to the gitlint category

Wayonb:
problem: gitlint is failing due to no docs category
solution: add a docs to the gitlint category

### DIFF
--- a/linters/scripts/lint_last_commit.sh
+++ b/linters/scripts/lint_last_commit.sh
@@ -4,7 +4,7 @@ set -ex
 
 GITLINT_COMMON_MODULES="dependency|jenkins|monorepo"
 GITLINT_MODULES="${GITLINT_COMMON_MODULES}|$(cat "$(git rev-parse --show-toplevel)"/.gitlintmodules)"
-GITLINT_CATEGORIES="feat|bug|fix|build|perf|task"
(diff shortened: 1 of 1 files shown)


segfaultxavi:
What is exactly failing? How can I reproduce it?
I am totally lost as to what are you trying to fix here.

Do we have an explanation somewhere of what these categories mean?
Should it be updated to describe the new category?

If a doc is wrong I consider that a bug, and if I add a new docs page, that is a feature.
Should ALL my commits to docs be simply tagged `docs` now?

Jaguar0625 (approved):
👍

Jaguar0625:
> What is exactly failing? How can I reproduce it? I am totally lost as to what are you trying to fix here.
> 
> Do we have an explanation somewhere of what these categories mean? Should it be updated to describe the new category?
> 
> If a doc is wrong I consider that a bug, and if I add a new docs page, that is a feature. Should ALL my commits to docs be simply tagged `docs` now?

there is a linter that runs to check commit message on dev commits. one of its steps is to check the subcomponent name in format:

```
[SUBCOMPONENT]: message
```

the list of allowable SUBCOMPONENT comes from var @Wayonb updated. it was missing "docs" as valid subcomponent, so commit linter failed. this will fix that by adding docs as valid subcomponent.

PR #2097, opened 2026-07-08
Revert "[mkdocs] fix: stop patching ezglossary"

segfaultxavi:
This reverts commit 4fd0bf72a86c3b3acb59d3200877d7ac5f531401.

There was a problem with the ezglossary plugin when the definition of a term included other terms.
We had a patch to fix it while waiting for an upstream fix.
The fix arrived so I removed the patch.
But it turns out the fix is incomplete, we still have issues when terms contain spaces.
I opened [a new issue](https://github.com/realtimeprojects/mkdocs-ezglossary/issues/35) upstream,
but while we wait, I am bringing back our patch.

### DIFF
--- a/mkdocs/scripts/ci/ezglossary.patch
+++ b/mkdocs/scripts/ci/ezglossary.patch
@@ -1,0 +1,12 @@
+--- a/src/mkdocs_ezglossary_plugin/plugin.py
++++ b/src/mkdocs_ezglossary_plugin/plugin.py
+@@ -239,6 +239,9 @@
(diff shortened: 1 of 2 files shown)


segfaultxavi (commented):
By the way, this is the reason why I didn't remove the patch after updating ezglossary.
It was kind of documented in 3bcf11e04dbe9e59e50495d12457374149d47d90 but I failed to detect it when I removed the patch.

segfaultxavi:
This is now part of https://github.com/symbol/symbol/pull/2099

PR #2099, opened 2026-07-08
[mkdocs] feat: update build dependencies

segfaultxavi:
Dependabot suggested that we updated a lot of our build deps.
This resulted in a couple of build breaks and incorrect rendering of the site.

This PR updates all suggested dependencies and fixes the breakage due to:
- Changed API of mkdocstrings: had to update our templates
- glightbox now interferes with ezglossary: had to update the ezglossary patch
- glightbox is stricter: had to fix errors in our templates
- Site-global links changed behavior: had to strip leading dashes from links

Additionally, a Japanese page was fixed that had diverged from the English one.
The patch would have conflicted if commited separately, so I pushed it in this PR.

Testing all this is the hard part.
- Manually inspected the new site for a while.
- Built the new version and dev and compared the HTML output using Codex.
- Examined pages which used the affected plugins.

No problems were detected.


mkdocs/scripts/ci/ezglossary.patch
```diff
@@ -0,0 +1,30 @@
+--- a/src/mkdocs_ezglossary_plugin/plugin.py
```
Jaguar0625, line 1: did you add this back intentionally? i thought in your last PR you removed a (this?) patch?

Jaguar0625 (approved):
👍

Wayonb (approved):
👍🏾

mkdocs/scripts/ci/ezglossary.patch
segfaultxavi: I had to add it back, yes. The reason is in the commit message of https://github.com/symbol/symbol/commit/2f8ee4a12cf583984690229cf3adde0bfb39f67a

And in another PR that resurrects this patch... which I now realize can be dropped.

PR #2106, opened 2026-07-13
[sdk/java]: Java model generator

Wayonb:
Add Java model generator.  Each class is written in its own file. 
The generator will create sweep tests to test the generated model classes


sdk/java/generator/AbstractEnumFormatter.py
```diff
@@ -0,0 +1,83 @@
...
+
+	def _value_literal(self, value):
+		"""Render an integer literal in the right Java type for this type's backing value."""
+		if 'long' == self.value_type:
+			return f'{value}L'
+
+		return str(value)
+
+	def _deserialize_body(self):
+		raise NotImplementedError('subclasses emit their own deserialize body')
```
Jaguar0625, line 33: should this be abstract ?

sdk/java/generator/AbstractEnumFormatter.py
```diff
@@ -0,0 +1,83 @@
...
+		"""Wrap a ``toString`` body into an overriding String-typed method descriptor."""
+		descriptor = MethodDescriptor(body=body)
+		descriptor.annotations = ['@Override']
+		descriptor.result = 'String'
+		return descriptor
+
+	def get_json_descriptor(self):
+		# JSON projection is the underlying numeric value.
+		descriptor = MethodDescriptor(body='return this.value;')
+		descriptor.result = self.value_type
```
Jaguar0625, line 81: shouldn't long value be string ?

sdk/java/generator/AbstractTypeFormatter.py
```diff
@@ -0,0 +1,95 @@
...
+
+	@abstractmethod
+	def get_ctor_descriptor(self) -> MethodDescriptor:
+		pass
+
+	def get_compare_to_descriptor(self) -> MethodDescriptor:  # pylint: disable=no-self-use
+		return None
+
+	def get_sort_descriptor(self) -> MethodDescriptor:  # pylint: disable=no-self-use
+		return None
```
Jaguar0625: would format with disable line as first body line to match rest of file style

sdk/java/generator/AbstractTypeFormatter.py
```diff
@@ -0,0 +1,95 @@
...
+	def get_implements_clause(self):
+		"""Java ``implements X, Y`` clause; return empty string for none."""
+		# pylint: disable=no-self-use
+		return ''
+
+	def get_permits(self):
+		"""Java ``permits X, Y`` list for a sealed hierarchy; a non-empty list flips the class
+		declaration to ``public sealed``. The default ``[]`` emits a regular class."""
+		# pylint: disable=no-self-use
+		return []
```
Jaguar0625, line 45: question: is there specific reason you wanted/needed to use permits ?

sdk/java/generator/FlagsTypeFormatter.py
```diff
@@ -0,0 +1,134 @@
...
+	def get_str_descriptor(self):
+		named_constants = ', '.join(f'{self.typename}.{entry.name}' for entry in self.enum_type.values)
+		named_names = ', '.join(f'"{self.typename}.{entry.name}"' for entry in self.enum_type.values)
+		body = (
+			f'final {self.typename}[] candidates = {{ {named_constants} }};\n'
+			f'final String[] names = {{ {named_names} }};\n'
+			f'for (int i = 0; i < candidates.length; ++i) {{\n'
+			f'\tif (candidates[i].value == this.value)\n'
+			f'\t\treturn names[i];\n'
+			f'}}\n'
```
Jaguar0625: not sure if this extra loop is needed ?

sdk/java/src/main/java/org/symbol/sdk/nem/models/LinkAction.java
```diff
@@ -0,0 +1,143 @@
...
+import java.util.HashMap;
+import java.util.Locale;
+import java.util.Map;
+
+import org.symbol.sdk.Serializer;
+import org.symbol.sdk.utils.BufferView;
+import org.symbol.sdk.utils.Converter;
+
+public enum LinkAction implements Serializer {
+	LINK(1L), UNLINK(2L);
```
Jaguar0625: can you format so all values are on separate lines (similar to Flags)?

sdk/java/src/main/java/org/symbol/sdk/nem/models/LinkAction.java
```diff
@@ -0,0 +1,143 @@
...
+	 * Deserializes an instance of this object from a byte payload.
+	 *
+	 * @param payload Byte payload to deserialize.
+	 * @return Deserialized object.
+	 * @throws IndexOutOfBoundsException if the buffer is truncated or a declared length overruns it
+	 * @throws IllegalArgumentException if a field carries an invalid value (e.g. a reserved field or unknown discriminator)
+	 */
+	public static LinkAction deserialize(final byte[] payload) {
+		return deserialize(new BufferView(payload));
+	}
```
Jaguar0625: why do you need two deserialize variants?

sdk/java/src/main/java/org/symbol/sdk/nem/models/LinkAction.java
```diff
@@ -0,0 +1,143 @@
...
+	/**
+	 * Converts this object to a JSON-serializable representation.
+	 *
+	 * @return JSON-serializable representation of this object.
+	 */
+	public long toJson() {
+		return this.value;
+	}
+
+}
```
Jaguar0625, line 132: remove blank line?

sdk/java/src/main/java/org/symbol/sdk/symbol/models/MosaicFlags.java
```diff
@@ -0,0 +1,201 @@
...
+				MosaicFlags.NONE, MosaicFlags.SUPPLY_MUTABLE, MosaicFlags.TRANSFERABLE, MosaicFlags.RESTRICTABLE, MosaicFlags.REVOKABLE
+		};
+		final String[] names = {
+				"MosaicFlags.NONE", "MosaicFlags.SUPPLY_MUTABLE", "MosaicFlags.TRANSFERABLE", "MosaicFlags.RESTRICTABLE",
+				"MosaicFlags.REVOKABLE"
+		};
+		for (int i = 0; i < candidates.length; ++i) {
+			if (candidates[i].value == this.value)
+				return names[i];
+		}
```
Jaguar0625: not sure if optimization is needed ?

sdk/java/src/main/java/org/symbol/sdk/nem/models/TransactionFactory.java
```diff
@@ -0,0 +1,109 @@
...
+		return supplier.get();
+	}
+
+	private static long toKey(final long value) {
+		return value;
+	}
+
+	private static long toKey(final long a, final long b) {
+		return (a << 32) | (b & 0xFFFFFFFFL);
+	}
```
Jaguar0625: presumably only one of these is needed per factory? can you conditionally generate only needed one ?

sdk/java/generator/format.py
```diff
@@ -0,0 +1,108 @@
...
+def _block_end(lines, index):
+	depth = 0
+	while index < len(lines):
+		depth += lines[index].count('{') - lines[index].count('}')
+		if 0 >= depth:
+			return index
+
+		index += 1
+
+	return index - 1
```
Jaguar0625: refactor since only braces are different

sdk/java/generator/format.py
```diff
@@ -0,0 +1,108 @@
...
+		return end
+
+
+def separate_if_blocks(source):
+	"""Insert a blank line after every if/else construct that is directly followed by another statement."""
+	lines = source.split('\n')
+	insert_after = set()
+	for index, line in enumerate(lines):
+		stripped = line.strip()
+		if not (stripped.startswith('if (') or stripped.startswith('if(')):
```
Jaguar0625: are we generating both variants? 🤔

sdk/java/generator/Generator.py
```diff
@@ -0,0 +1,319 @@
...
+		# pod/enum/flags types carry a generated `parse(Object)` factory; structs go through setField.
+		if isinstance(formatter, (PodTypeFormatter, EnumTypeFormatter, FlagsTypeFormatter)):
+			factory_entries.append(formatter.typename)
+
+		# Abstract structs need an additional factory.
+		if DisplayType.STRUCT == ast_model.display_type and ast_model.is_abstract:
+			factory_formatter = FactoryFormatter(factory_map, ast_model)
+			emit(factory_formatter.typename, str(TypeFormatter(factory_formatter, package_name)))
+
+	_write_models_manifest(output_directory, package_name, sorted(module_entries), sorted(factory_entries))
```
Jaguar0625, line 219: do we need this / how is this manifest used?

sdk/java/generator/Generator.py
```diff
@@ -0,0 +1,319 @@
...
+
+	# the generated per-model sweep test lives in the test tree
+	if 'models' == output_directory.name and output_directory.parent.name in ('nem', 'symbol'):
+		_write_models_sweep_test(output_directory, package_name, sweep_entries, ast_models)
+	else:
+		ModelsSweepTestFormatter(sweep_entries, ast_models).render()
```
Jaguar0625: why are you autogenerating tests as part of this vs separate script? shouldn't the vectors be sufficient? i believe that's all that's used in JS/PY sdks?

sdk/java/src/main/java/org/symbol/sdk/nem/models/Amount.java
```diff
@@ -0,0 +1,74 @@
...
+import org.symbol.sdk.BaseValue;
+import org.symbol.sdk.utils.BufferView;
+import org.symbol.sdk.utils.Converter;
+
+public final class Amount extends BaseValue<Amount> {
+	public static final int SIZE = 8;
+	public static final boolean IS_SIGNED = false;
+
+	/**
+	 * Creates a Amount.
```
Jaguar0625: a => an ?

sdk/java/generator/PodTypeFormatter.py
```diff
@@ -0,0 +1,115 @@
...
+				f'\t\treturn new {typename}(org.symbol.sdk.ByteArray.toBytes(rawValue));\n')
+		else:
+			parse_doc = f'Parses a raw value ({typename}, number, or decimal/0x-hex string) into a {typename}.'
+			parse_body = (
+				f'{typed_branch}'
+				'\t\tif (rawValue instanceof String string)\n'
+				f'\t\t\treturn new {typename}(org.symbol.sdk.utils.Converter.toLong(string));\n'
+				'\t\tif (rawValue instanceof Number number)\n'
+				f'\t\t\treturn new {typename}(org.symbol.sdk.utils.Converter.toLong(number));\n'
+				'\t\tthrow new IllegalArgumentException(\n'
```
Jaguar0625: toLong => toInt for pods where size != 8 ?

sdk/java/generator/printers.py
```diff
@@ -0,0 +1,373 @@
...
+
+	@staticmethod
+	def get_default_value():
+		return 'new java.util.ArrayList<>()'
+
+	@property
+	def is_variable_size(self):
+		descriptor = self.descriptor
+		is_constrained = descriptor.field_type.is_byte_constrained or descriptor.extensions.is_contents_abstract
+		return is_constrained and descriptor.field_type.alignment
```
Jaguar0625, line 196: why do you need to check alignment, isn't is_constrained sufficient?

sdk/java/generator/printers.py
```diff
@@ -0,0 +1,373 @@
...
+	"""Reference to another generated type (POD int, POD byte-array, enum, struct)."""
+
+	def __init__(self, descriptor, name=None):
+		super().__init__(descriptor, name)
+		self.type_hint = {
+			DisplayType.INTEGER: 'pod:',
+			DisplayType.BYTE_ARRAY: 'pod:',
+			DisplayType.TYPED_ARRAY: 'pod:',
+			DisplayType.ENUM: 'enum:',
+			DisplayType.STRUCT: 'struct:',
```
Jaguar0625: remove trailing comma ?

sdk/java/generator/StructTypeFormatter.py
```diff
@@ -0,0 +1,783 @@
...
+				steps.append(
+					f'java.util.Arrays.compareUnsigned(\n'
+					f'\t\torg.symbol.sdk.utils.Transforms.{method}(this.{java_field}.bytes()),\n'
+					f'\t\torg.symbol.sdk.utils.Transforms.{method}(other.{java_field}.bytes()))')
+
+		declaration = 'final int c = ' if 2 == len(steps) else 'int c = '
+		body_lines = []
+		for index, step in enumerate(steps[:-1]):
+			body_lines.append(f'{declaration if 0 == index else "c = "}{step};')
+			body_lines.append('if (0 != c) return c;')
```
Jaguar0625, line 266: do you need this optimization ?

sdk/java/generator/StructTypeFormatter.py
```diff
@@ -0,0 +1,783 @@
...
+		if not self.is_type_abstract:
+			body += 'return instance;'
+
+		return body
+
+	def get_deserialize_descriptor(self):
+		descriptor = MethodDescriptor(
+			body=self._deserialize_body(),
+			arguments=['org.symbol.sdk.utils.BufferView view'] if not self.is_type_abstract
+			else ['org.symbol.sdk.utils.BufferView cursor', f'{self.typename} instance'])
```
Jaguar0625: why are you conditioning the name of cursor vs view instead of using the same name everywhere?

sdk/java/src/main/java/org/symbol/sdk/nem/models/MosaicDefinition.java
```diff
@@ -0,0 +1,342 @@
...
+		this.id = value;
+	}
+
+	/**
+	 * Gets the description field.
+	 *
+	 * @return The description field.
+	 */
+	public byte[] getDescription() {
+		return this.description == null ? null : this.description.clone();
```
Jaguar0625: yoda?

sdk/java/generator/StructTypeFormatter.py
```diff
@@ -0,0 +1,783 @@
...
+			body = (
+				f'return {self.field_name(field)} == null'
+				f' ? null'
+				f' : {self.field_name(field)}.clone();')
+			result = java_type
+		elif field.is_conditional:
+			# Conditional/union fields are genuinely nullable — return Optional<T> so the API
+			# surfaces the presence/absence semantics instead of relying on a null sentinel.
+			body = f'return java.util.Optional.ofNullable({self.field_name(field)});'
+			result = f'java.util.Optional<{java_type}>'
```
Jaguar0625, line 553: question: why did you choose to use Optional ?

sdk/java/generator/StructTypeFormatter.py
```diff
@@ -0,0 +1,783 @@
...
+	def get_extra_methods(self):
+		"""Emit the reflection-free descriptor surface (typeHints/setField/getField)."""
+		methods = []
+		type_hints = self._render_type_hints_method()
+		if type_hints is not None:
+			methods.append(type_hints)
+
+		methods.append(self._render_set_field_method())
+		methods.append(self._render_get_field_method())
+		to_non_verifiable = self._render_to_non_verifiable_method()
```
Jaguar0625, line 595: why are you generating this vs implementing on the TransactionFactory?

sdk/java/generator/StructTypeFormatter.py
```diff
@@ -0,0 +1,783 @@
...
+		if not self._ast_models:
+			return None
+
+		if self.struct.name.startswith('NonVerifiable'):
+			return None
+
+		target_name = f'NonVerifiable{self.struct.name}'
+		for candidate in self._ast_models:
+			if candidate.name == target_name:
+				return candidate
```
Jaguar0625: use next ?

sdk/java/generator/StructTypeFormatter.py
```diff
@@ -0,0 +1,783 @@
...
+				f'\tpublic {sibling.name} toNonVerifiable()',
+				'\t\tthrow new IllegalStateException("toNonVerifiable() not implemented on " + getClass().getName());\n')
+
+		source_field_names = {field.extensions.printer.name for field in self.non_reserved_fields(include_inherited=True)}
+		sibling_formatter = self._sibling_formatter(sibling)
+		copy_lines = []
+		for field in sibling_formatter.non_reserved_fields(include_inherited=True):
+			name = field.extensions.printer.name
+			if name in source_field_names:
+				copy_lines.append(f'\t\tresult.setField("{name}", this.{name});')
```
Jaguar0625: why use setFields vs field-specific setters?

sdk/java/generator/StructTypeFormatter.py
```diff
@@ -0,0 +1,783 @@
...
+		return None
+
+	def _render_to_non_verifiable_method(self):
+		"""Emit the typed {@code toNonVerifiable()} on NEM Transaction + its concrete subclasses."""
+		sibling = self._find_non_verifiable_sibling()
+		if sibling is None:
+			return None
+
+		if self.is_type_abstract:
+			# Abstract Transaction → emit a concrete default that throws.
```
Jaguar0625, line 617: is this part of the interface ?

sdk/java/generator/StructTypeFormatter.py
```diff
@@ -0,0 +1,783 @@
...
+			'\t\treturn TYPE_HINTS;\n')
+
+	def _render_field_switch(self, doc_lines, signature, case_rhs, default_line, *, is_expression):
+		cases = [
+			f'\t\t\tcase "{field.extensions.printer.name}" -> {case_rhs(field)};'
+			for field in self.non_reserved_fields(include_inherited=False)]
+		body_inner = '\n'.join(cases) + '\n' if cases else ''
+		switch_open = '\t\treturn switch (name) {\n' if is_expression else '\t\tswitch (name) {\n'
+		switch_close = '\t\t};\n' if is_expression else '\t\t}\n'
+		return (
```
Jaguar0625: why not use MethodDescriptor here ?

sdk/java/generator/TypeFormatter.py
```diff
@@ -0,0 +1,357 @@
...
+		_, name = _split_argument(argument)
+		description = f'New {setter_property} value.' if setter_property and 'value' == name else _argument_description(name)
+		param_lines.append(f'@param {name} {description}')
+
+	return_line = None
+	if result not in (None, 'void'):
+		if accessor and 'get' == accessor[0]:
+			return_line = f'@return The {accessor[1]} field.'
+		else:
+			return_line = f'@return {_METHOD_RETURNS.get(method_name, "Result of the operation.")}'
```
Jaguar0625: should it be error if description does not exist vs using default text  ?

sdk/java/generator/TypeFormatter.py
```diff
@@ -0,0 +1,357 @@
...
+	def __str__(self):
+		return self._render()
+
+	# -- public API ---------------------------------------------------------------------------
+
+	def _render(self):  # pylint: disable=too-many-locals, too-many-statements
+		typename = self.provider.typename
+		modifiers = 'public final '
+		if self.provider.is_type_abstract:
+			modifiers = 'public '
```
Jaguar0625: can make single line

sdk/java/generator/TypeFormatter.py
```diff
@@ -0,0 +1,357 @@
...
+
+		base = self.provider.get_base_class()
+		impls = self.provider.get_implements_clause()
+		extends = f' extends {base}' if base else ''
+		implements = f' implements {impls}' if impls else ''
+
+		# Sealed hierarchies: abstract bases declare their concrete subclasses via `permits` so downstream callers can
+		# exhaustive-switch over the closed set.
```
Jaguar0625, line 255: is this needed by factory ?

sdk/java/generator/TypeFormatter.py
```diff
@@ -0,0 +1,357 @@
...
+	def _render(self):  # pylint: disable=too-many-locals, too-many-statements
+		typename = self.provider.typename
+		modifiers = 'public final '
+		if self.provider.is_type_abstract:
+			modifiers = 'public '
+
+		base = self.provider.get_base_class()
+		impls = self.provider.get_implements_clause()
+		extends = f' extends {base}' if base else ''
+		implements = f' implements {impls}' if impls else ''
```
Jaguar0625: impls / implements is confusing naming imo

sdk/java/generator/TypeFormatter.py
```diff
@@ -0,0 +1,357 @@
...
+			body += indent('\n'.join(fields)) + '\n'
+
+		ctor_descriptor = self.provider.get_ctor_descriptor()
+		if ctor_descriptor is not None:
+			ctor_method = _generate_method(ctor_descriptor, is_constructor=True, owner_typename=typename)
+			body += indent(ctor_method) + '\n'
+
+		extra = self.provider.get_extra_methods()
+		for raw in extra:
+			body += raw + '\n'
```
Jaguar0625: any reason not to include these in methods vs formatting separately ?

sdk/java/generator/TypeFormatter.py
```diff
@@ -0,0 +1,357 @@
...
+		value_type = self.provider.value_type
+
+		constants = self.provider.get_enum_constants()
+		const_lines = ',\n\t'.join(constants) + ';'
+
+		fields = f'\tprivate final {value_type} value;\n'
+
+		static_block_lines = []
+		if hasattr(self.provider, 'get_static_block_lines'):
+			static_block_lines = self.provider.get_static_block_lines()
```
Jaguar0625: when can this be unset for enum ?

sdk/java/generator/TypeFormatter.py
```diff
@@ -0,0 +1,357 @@
...
+		ctor_descriptor = self.provider.get_ctor_descriptor()
+		ctor = _generate_method(ctor_descriptor, is_constructor=True, owner_typename=typename)
+
+		methods = []
+		_append_if_not_none(methods, _build_method('size', self.provider.get_size_descriptor()))
+		_append_if_not_none(methods, _build_method('serialize', self.provider.get_serialize_descriptor()))
+		_append_if_not_none(methods, _build_method('deserialize', self.provider.get_deserialize_descriptor()))
+
+		for descriptor in self.provider.get_getter_setter_descriptors():
+			methods.append(_generate_method(descriptor))
```
Jaguar0625: _build_getters_setters ?

sdk/java/generator/TypeFormatter.py
```diff
@@ -0,0 +1,357 @@
...
+
+		static_block = ('\n' + indent('\n'.join(static_block_lines)) + '\n') if static_block_lines else ''
+
+		ctor_descriptor = self.provider.get_ctor_descriptor()
+		ctor = _generate_method(ctor_descriptor, is_constructor=True, owner_typename=typename)
+
+		methods = []
+		_append_if_not_none(methods, _build_method('size', self.provider.get_size_descriptor()))
+		_append_if_not_none(methods, _build_method('serialize', self.provider.get_serialize_descriptor()))
+		_append_if_not_none(methods, _build_method('deserialize', self.provider.get_deserialize_descriptor()))
```
Jaguar0625: keep order consistent across classes and enums

sdk/java/scripts/ci/setup_build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+set -ex
+
+python3 -m pip install -r "$(git rev-parse --show-toplevel)/catbuffer/parser/requirements.txt"
```
Jaguar0625: i think this is unneeded, since it is being installed in setup_lint

sdk/java/scripts/run_catbuffer_generator.sh
```diff
@@ -0,0 +1,53 @@
...
+		generate_code "${name}" "${name}/models"
+	done
+	# The Python templates emit canonical-but-unwrapped Java; Spotless (eclipse formatter) owns
+	# line-wrapping/whitespace, so format the freshly generated tree to match the committed style.
+	# SKIP_SPOTLESS is set by the `gradle generateModels` task, which finalizes with spotlessApply
+	# itself — running gradlew here too would nest one Gradle build inside another and deadlock.
+	if [[ -z "${SKIP_SPOTLESS:-}" ]]; then
+		# -PspotlessGenerated scopes the formatter to the generated subtrees only (see build.gradle.kts).
+		(cd "$(git rev-parse --show-toplevel)/sdk/java" && ./gradlew --no-daemon spotlessApply -PspotlessGenerated)
+	fi
```
Jaguar0625, line 45: is good, but then do you need to do all the indentation manually in the generator (esp thinking about format.py)

sdk/java/scripts/run_catbuffer_generator.sh
```diff
@@ -0,0 +1,53 @@
...
+elif [[ "$1" = "dryrun" ]]; then
+	echo "running dryrun"
+	for name in "nem" "symbol"; do
+		generate_code "${name}" "${name}_dryrun/models"
+		rm -rf "$(git rev-parse --show-toplevel)/sdk/java/src/main/java/org/symbol/sdk/${name}_dryrun"
```
Jaguar0625: did you intentionally remove diff ?

sdk/java/scripts/run_catbuffer_generator.sh
```diff
@@ -0,0 +1,53 @@
...
+		rm -rf "${models_dir}"
+		generate_code "${name}" "${name}/models"
+	done
+	# The Python templates emit canonical-but-unwrapped Java; Spotless (eclipse formatter) owns
+	# line-wrapping/whitespace, so format the freshly generated tree to match the committed style.
+	# SKIP_SPOTLESS is set by the `gradle generateModels` task, which finalizes with spotlessApply
+	# itself — running gradlew here too would nest one Gradle build inside another and deadlock.
+	if [[ -z "${SKIP_SPOTLESS:-}" ]]; then
+		# -PspotlessGenerated scopes the formatter to the generated subtrees only (see build.gradle.kts).
+		(cd "$(git rev-parse --show-toplevel)/sdk/java" && ./gradlew --no-daemon spotlessApply -PspotlessGenerated)
```
Jaguar0625: might want to rename to `spotlessOnlyGenerated` or `spotlessGeneratedOnly`

Jaguar0625 (commented):
👍

sdk/java/generator/AbstractTypeFormatter.py
Wayonb: createTransactionFromTypedDescriptor() takes a SymbolTransactionDescriptor interface, so limiting which classes can implement it, since I know the descriptor class will convert all transaction values correctly.
This is why I included them initially, so you could see the entire design.

sdk/java/generator/FlagsTypeFormatter.py
Wayonb: The first loop will handle NONE/ALL correctly, but let me see if it can be improved.
One issue I noticed but forgot to bring up is that not all flags enums have a NONE value.  Not sure if this is a schema issue.
In the JavaScript AccountRestrictionFlags, if the value is 0, the toString() will fail, while other values return an empty string.

Wayonb: The first loop will handle NONE/ALL correctly, but let me see if it can be improved.
One issue I noticed but forgot to bring up is that not all flags enums have a NONE value.  Not sure if this is a schema issue.
In the JavaScript AccountRestrictionFlags, if the value is 0, the toString() will fail, while other values return an empty string.

sdk/java/src/main/java/org/symbol/sdk/nem/models/LinkAction.java
Wayonb: Let me check, I think it's the spotless formatting that is doing this.

Wayonb: Let me check, I think it's the spotless formatting that is doing this.

sdk/java/src/main/java/org/symbol/sdk/nem/models/LinkAction.java
Wayonb: I thought most would always have bytes as the payload to start.  If you want to do this as in JS, we can always use a ByteBuffer and pass it to the BufferView.  I consider the BufferView class as internal.  Let me know your thoughts.

sdk/java/src/main/java/org/symbol/sdk/nem/models/TransactionFactory.java
Wayonb: It is generated code, so I did not think much of it 😅  yeah, should be able to remove

sdk/java/generator/format.py
Wayonb: no, can be remove

sdk/java/generator/Generator.py
Wayonb: The model factories are created by the generator, and this makes it easier to include them in the transaction, block factory.  If something changes in the schema, these factories are automatically updated.

sdk/java/generator/Generator.py
Wayonb: I added these tests because
1. wanted unit test by default.
2. The catbuffer schema can be updated to add new types without the vectors getting added.
3. for code coverage

I think the vectors are e2e for all the factories, which are good.
I can move it to another generator if you want, but I thought it best to generate at the same time as the model code, so it's not missed.

sdk/java/generator/PodTypeFormatter.py
Wayonb: The BaseValue class stores everything as long.  The ctor will verify that the value is correct to fit the size.
If you want to store long vs int then I would need to make the BaseValue class to also take the type and deal with un/boxing.

sdk/java/generator/printers.py
Wayonb: Yeah, I thought so too, but this is the same for JS/Python to filter out receipts.  I added the comment also.

sdk/java/generator/StructTypeFormatter.py
Wayonb: Why not? Is it readability?

sdk/java/generator/StructTypeFormatter.py
Wayonb: In deserializeInto(), the arg is a cursor since it is not a snapshot and will update the caller's view.
This is similar to JS's deserialize(payload) vs. _deserialize(view) methods.

sdk/java/generator/StructTypeFormatter.py
Wayonb: I thought that is the Java way to deal with nulls.  Do you prefer to return null?

sdk/java/generator/StructTypeFormatter.py
Wayonb: I tried to avoid using Java reflection, so it was easier to let each class own this, since it knows at generation time.

sdk/java/generator/StructTypeFormatter.py
Wayonb: yea, the transaction class

sdk/java/generator/AbstractTypeFormatter.py
Jaguar0625: i guess i am asking what is benefit of using permits instead of just using interface? is there some common expanded interface among those permitted classes or is it just that we don't expect anyone else to implement it (so it more a safety measure / defensive programming)?

sdk/java/src/main/java/org/symbol/sdk/nem/models/LinkAction.java
Jaguar0625: > we can always use a ByteBuffer and pass it to the BufferView

that is what i was wondering

> I consider the BufferView class as internal

idk if i see a need for that, since a byte[] can be wrapped in it (without any copying). if anything, would add byte[] overload to TransactionFactory. but, for generated code, not expected to be called directly, i think single version is fine.

sdk/java/generator/Generator.py
Jaguar0625: i would prefer a separate generator 🙏

sdk/java/generator/StructTypeFormatter.py
Jaguar0625: yea, my first instinct was readability, but i guess it's fine

sdk/java/generator/StructTypeFormatter.py
Jaguar0625: in JS, they are different _types_. here i think they are the same type?

sdk/java/generator/StructTypeFormatter.py
Jaguar0625: idk, my instinct would be null (since these are reference types). but, if Optional is preferred way in java you can leave it.

sdk/java/generator/StructTypeFormatter.py
Jaguar0625: just add a comment to that effect please

sdk/java/generator/StructTypeFormatter.py
Jaguar0625: is there reason Transaction is not abstract class ?

sdk/java/scripts/run_catbuffer_generator.sh
Wayonb: yea, I first tried to get the generator to create correctly formatted code but spotless always failed 😅 
So spotless was the easy option, but I left all the formatting in place.  Plan to revisit this to see if I can improve

sdk/java/generator/AbstractTypeFormatter.py
Wayonb: The interface returns a Map<String, Object>, so the value can be anything. So yeah, also an extra level of safety/defensive.
There is an override of createTransactionFromTypedDescriptor() that takes Map<String, Object>, but it's not public since I don't see why anyone would want to use this, but I could be wrong 😅

sdk/java/src/main/java/org/symbol/sdk/nem/models/LinkAction.java
Wayonb: The TransactionFactory already takes both, so I will take a look at updating the deserialize() to take ByteBuffer, and this will match JS using Uint8Array

sdk/java/generator/StructTypeFormatter.py
Wayonb: Its used in the transaction factory to determine the transaction type and version.  same as JS/Python.

sdk/java/generator/ModelsSweepTestGenerator.py
```diff
@@ -0,0 +1,52 @@
+"""Standalone generator for the per-model sweep test. Invoked by catparser
+(``-g generator.ModelsSweepTestGenerator``) separately from the model generator: it reconstructs the
+model formatters and emits only ``ModelsSweepTest.java`` into the mirrored test tree.
+"""
+
+from pathlib import Path
+
+from catparser.generators.util import build_factory_map, extend_models
+
+from .Generator import _derive_package_name, create_printer, to_type_formatter_instance
```
Jaguar0625: remove _ prefix since being imported

sdk/java/generator/StructTypeFormatter.py
```diff
@@ -573,7 +569,7 @@ def _create_setter(self, field):
 		method_name = f'set{capitalize(printer_name)}'
 		# byte[] fields: defensive clone so subsequent caller mutations don't bleed into the model.
 		if isinstance(field.extensions.printer, ArrayPrinter):
-			body = f'{self.field_name(field)} = value == null ? null : value.clone();'
+			body = f'{self.field_name(field)} = null == value ? null : value.clone();'
```
Jaguar0625: the `= null == value`  looks weird, but i don't see anything like that in generated code. is this branch reached? 🤔

sdk/java/generator/StructTypeFormatter.py
```diff
@@ -664,51 +662,55 @@ def _sibling_formatter(self, sibling):
...
+		# pylint: disable=no-self-use
+		descriptor = MethodDescriptor(method_name='typeHints', body='return TYPE_HINTS;')
+		descriptor.result = 'java.util.Map<String, String>'
+		descriptor.annotations = ['@Override']
+		descriptor.documentation = [
+			'Returns the type-hint map driving the rule-based descriptor pipeline.', '', '@return Field-name → rule-key map.']
+		return descriptor
 
-	def _render_field_switch(self, doc_lines, signature, case_rhs, default_line, *, is_expression):
+	def _render_field_switch(self, doc_lines, case_rhs, default_line, *, method_name, arguments, result, is_expression):
```
Jaguar0625, line 664: why is there a * argument? 🤔

sdk/java/generator/SweepTestSupport.py
```diff
@@ -0,0 +1,31 @@
+"""Shared plumbing for the standalone sweep-test generators (generator.ModelsSweepTestGenerator and
+generator.DescriptorSweepTestGenerator)."""
+
+from pathlib import Path
+
+from .format import separate_if_blocks
+from .Generator import _finalize_imports
```
Jaguar0625: remove _ prefix since being imported

sdk/java/scripts/ci/lint_python.sh
```diff
@@ -15,4 +15,5 @@ find . -type f -name "*.py" -print0 | PYTHONPATH=. xargs -0 python3 -m pycodesty
 SEPARATOR="$([ "$(uname -o)" = "Msys" ] && echo ";" || echo ":")"
 find . -type f -name "*.py" -print0 | \
 	PYTHONPATH=".${SEPARATOR}$(git rev-parse --show-toplevel)/catbuffer/parser" xargs -0 python3 -m pylint \
-	--rcfile "$(git rev-parse --show-toplevel)/linters/python/.pylintrc"
+	--rcfile "$(git rev-parse --show-toplevel)/linters/python/.pylintrc" \
+	--enable=useless-suppression
```
Jaguar0625, line 19: what does this flag do ?

sdk/java/scripts/run_catbuffer_generator.sh
```diff
@@ -32,20 +33,27 @@ if [[ $# -eq 0 ]]; then
...
 	fi
 elif [[ "$1" = "dryrun" ]]; then
-	echo "running dryrun"
+	echo "running dryrun diff"
+	git_root="$(git rev-parse --show-toplevel)"
+	base="${git_root}/sdk/java/src/main/java/org/symbol/sdk"
 	for name in "nem" "symbol"; do
 		generate_code "${name}" "${name}_dryrun/models"
-		rm -rf "$(git rev-parse --show-toplevel)/sdk/java/src/main/java/org/symbol/sdk/${name}_dryrun"
+		generate_code "${name}" "${name}_dryrun/models" generator.ModelsSweepTestGenerator
```
Jaguar0625, line 52: i thought ModelsSweepTestGenerator would not generate to alternative directory ? 🤔

sdk/java/src/main/java/org/symbol/sdk/utils/BufferView.java
```diff
@@ -19,14 +19,28 @@ public BufferView(final byte[] buffer) {
...
 		return source.slice().order(ByteOrder.LITTLE_ENDIAN);
 	}
 
+	/**
+	 * Returns a {@link ByteBuffer} over the visible window (this view is not advanced).
+	 *
+	 * @return ByteBuffer over the current window.
+	 */
+	public ByteBuffer buffer() {
+		return rebasedSlice(buffer);
```
Jaguar0625, line 41: isn't this same as this.buffer ?

sdk/java/src/main/java/org/symbol/sdk/utils/BufferView.java
```diff
@@ -19,14 +19,28 @@ public BufferView(final byte[] buffer) {
...
 	private static ByteBuffer rebasedSlice(final ByteBuffer source) {
 		return source.slice().order(ByteOrder.LITTLE_ENDIAN);
 	}
 
+	/**
+	 * Returns a {@link ByteBuffer} over the visible window (this view is not advanced).
+	 *
+	 * @return ByteBuffer over the current window.
+	 */
+	public ByteBuffer buffer() {
```
Jaguar0625, line 40: might want to add test coverage for this and newly public ctor

Jaguar0625 (commented):
👍

sdk/java/src/main/java/org/symbol/sdk/utils/BufferView.java
Wayonb: No, the slice always creates a ByteBuffer with an index of 0, unlike this.buffer, the index could have progressed.

sdk/java/scripts/run_catbuffer_generator.sh
Wayonb: Kept the logic the same as the model generator.
The models go to ``src/main`` while the sweeptests go to ``src/test`` folder
The models need a separate folder since there are multiple files, but I kept the same logic for tests.

sdk/java/scripts/ci/lint_python.sh
Wayonb: I noticed during my refactor that I was leaving pylint disable statements when they were no longer needed.  This will cause the linter to fail in those cases, so I can clean them up.

sdk/java/generator/StructTypeFormatter.py
Wayonb: I think I initially added this because some of the methods had too many positional args.  I can remove and just disable pylint where needed.

sdk/java/generator/StructTypeFormatter.py
Wayonb: It's in a few of the models to clone bytes.
```
	public void setName(final byte[] value) {
		this.name = null == value ? null : value.clone();
	}
```

sdk/java/generator/Generator.py
```diff
@@ -223,7 +224,7 @@ def emit(typename, source):
 def _write_models_manifest(output_directory, package_name, module_entries, factory_entries):
 	"""Write ``Models.java``: the name→Class MODULE map and the name→parse FACTORIES map."""
 	lines = [
-		'// Auto-generated by sdk/java/generator. Do not edit directly.',
+		'// Auto-generated by generator. Do not edit directly.',
```
Jaguar0625: can call auto_generated_header ?

sdk/java/scripts/run_catbuffer_generator.sh
Jaguar0625: i thought this would skip since parent is not in (nem, symbol) ?
```py
	# dryrun output dirs (…_dryrun/models) do not match the real layout, so render without writing — a
	# schema change that breaks sweep emission still fails loudly in CI.
	if 'models' == output_directory.name and output_directory.parent.name in ('nem', 'symbol'):
		write_sweep_test_file(
			output_directory, derive_package_name(output_directory), header=_HEADER, imports=_SWEEP_TEST_IMPORTS,
			filename='ModelsSweepTest.java', body=formatter.render())
		return 1

```

sdk/java/src/test/java/org/symbol/sdk/utils/BufferViewTest.java
```diff
@@ -580,4 +583,154 @@ void canPeekAtDifferentPositions() {
...
+
+			// Assert: the view exposes the remaining bytes, rebased to start at its own first byte.
+			assertThat(view.length(), equalTo(5));
+			assertThat(view.peekBytes(view.length()), equalTo(new byte[]{
+					2, 3, 4, 5, 6
+			}));
+		}
+
+		@Test
+		void canhonorsSourcePositionAndLimit() {
```
Jaguar0625: canHonor ?

Jaguar0625 (approved):
👍

sdk/java/scripts/run_catbuffer_generator.sh
Wayonb: The path gets updated in the write_sweep_test_file() to the test folder.

Jaguar0625 (approved):
👍

sdk/java/scripts/run_catbuffer_generator.sh
Wayonb: I just cleaned up my branch and realized you are correct.  The sweep test wasn't getting written in dry run. I had the test file on disk so I thought it was working. 😅   I will add the fix to the next PR

PR #2109, opened 2026-07-15
[mkdocs] fix: format main manual sections as cards

zero4862:
Convert the section lists on the Developer and User manual landing pages into Material for [MkDocs grids](https://squidfunk.github.io/mkdocs-material/reference/grids/).

This is an opinionated change, but I think it makes the pages easier to scan, since the cards are displayed side by side, and feels more actionable.

image

image

I formatted the layout as two columns in the user manual because we'll be adding more cards in #2094.

If this looks good, I'll port it to NEM.

### DIFF
--- a/mkdocs/overrides/assets/stylesheets/extra.css
+++ b/mkdocs/overrides/assets/stylesheets/extra.css
@@ -153,6 +153,27 @@
 	color: var(--md-default-fg-color--light);
 }
 
+/* Force three equal-width cards per row, stacking on small screens */
+.md-typeset .grid.cards.cols-3 {
+	grid-template-columns: repeat(3, minmax(0, 1fr));
+}
+
+@media screen and (max-width: 44.9375em) {
+	.md-typeset .grid.cards.cols-3 {
+		grid-template-columns: 1fr;
+	}
+}
+
+/* Pin the "learn more" link to the bottom of each card so links align */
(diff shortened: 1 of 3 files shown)


segfaultxavi:
I appreciate trying to improve this page. But I am going to reject the cards for these reasons:
- We do not have landing pages for the sections, so the links necessarily point to individual pages inside the sections. This makes those pages special when they are not, and it's a huge pet peeve of mine 😄 
- Mixing cards with regular text confuses my boomer brain. Is this a menu or a text page? When selecting pages I prefer to use the navigation bar, where the whole scope is visible.
- We already have cards in the home screen, so in my mind, cards = home screen. Adding more card-based menus makes it harder to recall where a particular menu was (this is not backed by actual science!).

But I leave the PR open so we can discuss it further if you want. Otherwise, you can close it.

zero4862:
@segfaultxavi Fair enough, we can close this one. As I said, it's opinionted.

Two smaller things, unrelated to the cards. Icons render as solid black, which IMO looks a bit off (specially mobile and shoestring).

image

Also, the link titles "The Symbol Desktop Wallet" don't match the nav entries "Symbol Desktop Wallet". Same for Shoestring.

Happy to open a separate PR if you're up for it.

segfaultxavi:
Good points both!

I've already pushed a commit to the Mobile Guides branch :)

PR #2110, opened 2026-07-16
Docs staging

segfaultxavi:
Docs pending review

### DIFF
--- a/mkdocs/snippets/devbook/mosaics/create_mosaic.mjs
+++ b/mkdocs/snippets/devbook/mosaics/create_mosaic.mjs
@@ -80,7 +80,7 @@ try {
 	// --- CREATING MOSAIC DEFINITION ---
 	console.log('\n--- Creating mosaic definition ---');
 	// [>step-3]
-	const nonce = Math.floor(Date.now() / 1000) & 0x7FFFFFFF;
+	const nonce = Math.floor(Date.now() / 1000) % 0x100000000;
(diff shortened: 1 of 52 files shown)


mkdocs/snippets/devbook/mosaics/create_mosaic.mjs
```diff
@@ -80,7 +80,7 @@ try {
 	// --- CREATING MOSAIC DEFINITION ---
 	console.log('\n--- Creating mosaic definition ---');
 	// [>step-3]
-	const nonce = Math.floor(Date.now() / 1000) & 0x7FFFFFFF;
+	const nonce = Math.floor(Date.now() / 1000) % 0x100000000;
```
Jaguar0625, line 83: did you intend to update py code too? py code is currently: 
```py
nonce = int(time.time()) & 0xFFFFFFFF
```

Jaguar0625 (commented):
👍

mkdocs/snippets/devbook/mosaics/create_mosaic.mjs
segfaultxavi: Hm. Good point.
I only wanted to make sure both PY and JS were masking the same number of bits, but it looks a bit strange now if they don't use the same operator.
I'll fix it.

Jaguar0625 (approved):
👍

PR #2111, opened 2026-07-17
[sdk/java]: Java descriptor generator

Wayonb:
1. Add Java descriptor generator
2. Add Java descriptor sweep test generator
3. Add bash scripts to run both generators refactored.
4. Add NEM and Symbol Transaction Factory, which pulled in Network, Address, Rule Base factory, etc
5. Add tests for all


sdk/java/scripts/run_catbuffer_common.sh
```diff
@@ -0,0 +1,102 @@
...
+
+	local git_root
+	git_root="$(git rev-parse --show-toplevel)"
+
+	PYTHONPATH="${git_root}/catbuffer/parser:${git_root}/sdk/java" python3 -m catparser \
+		--schema "${git_root}/catbuffer/schemas/$1/${CATBUFFER_SCHEMA}" \
+		--include "${git_root}/catbuffer/schemas/$1" \
+		--output "${git_root}/sdk/java/src/main/java/org/symbol/sdk/$2" \
+		--quiet \
+		--generator "${3:-${CATBUFFER_GENERATOR}}"
```
Jaguar0625: would remove default and always pass 3rd parameter now

sdk/java/src/main/java/org/symbol/sdk/nem/NemTransactionFactory.java
```diff
@@ -0,0 +1,192 @@
...
+
+	/**
+	 * Attaches a signature to a transaction.
+	 *
+	 * @param transaction Transaction object.
+	 * @param signature Signature to attach.
+	 * @return JSON transaction payload.
+	 */
+	public static String attachSignature(final Transaction transaction, final Signature signature) {
+		transaction.setField("signature", new org.symbol.sdk.nem.models.Signature(signature.bytes()));
```
Jaguar0625: setSignature ?

sdk/java/src/main/java/org/symbol/sdk/nem/NemTransactionFactory.java
```diff
@@ -0,0 +1,192 @@
...
+
+	/**
+	 * Generates a JSON representation of transaction that can be sent to a node.
+	 *
+	 * @param transaction Transaction object.
+	 * @return JSON transaction payload.
+	 */
+	public static String toJson(final Transaction transaction) {
+		final String transactionHex = Converter.uint8ToHex(toNonVerifiableTransaction(transaction).serialize());
+		final org.symbol.sdk.nem.models.Signature signature = (org.symbol.sdk.nem.models.Signature) transaction.getField("signature");
```
Jaguar0625: getSignature ?

sdk/java/src/main/java/org/symbol/sdk/nem/NemTransactionFactory.java
```diff
@@ -0,0 +1,192 @@
...
+		final Function<Object, Object> nemTypeConverter = value -> {
+			if (value instanceof Address handWritten) {
+				final byte[] encoded = handWritten.toString().getBytes(java.nio.charset.StandardCharsets.UTF_8);
+				return new org.symbol.sdk.nem.models.Address(encoded);
+			}
+
+			return null;
+		};
+		final RuleBasedTransactionFactory factory = new RuleBasedTransactionFactory(nemTypeConverter, typeRuleOverrides);
+		factory.registerParsers(Models.FACTORIES);
```
Jaguar0625: probably would rename to something like ENUM_FACTORIES or ENUM_MAPPINGS

sdk/java/src/main/java/org/symbol/sdk/symbol/AddressRules.java
```diff
@@ -0,0 +1,26 @@
...
+	}
+
+	/**
+	 * Parses a descriptor value into a resolved model address, rejecting a namespace alias — used as the {@code Address} rule for fields
+	 * that cannot hold one (e.g. a block beneficiary or an alias transaction's target).
+	 *
+	 * @param value Descriptor value (Address, base32 string, byte array, or SDK ByteArray).
+	 * @return Resolved model address.
+	 */
+	static org.symbol.sdk.symbol.models.Address resolved(final Object value) {
```
Jaguar0625, line 19: where are you calling this ?

sdk/java/src/main/java/org/symbol/sdk/symbol/SymbolTransactionFactory.java
```diff
@@ -0,0 +1,258 @@
...
+		final BiConsumer<Network, CatbufferType> idAssigner = ID_ASSIGNERS.get((TransactionType) target.getField("type"));
+		if (null != idAssigner)
+			idAssigner.accept(network, target);
+
+		return target;
+	}
+
+	// post-processors that derive a transaction's artifact id after it is built, keyed by transaction type so a new
+	// artifact-bearing type is a data addition here rather than another branch in createAndExtend
+	private static final Map<TransactionType, BiConsumer<Network, CatbufferType>> ID_ASSIGNERS = Map.of(
```
Jaguar0625: CatbufferType => Transaction ?

sdk/java/src/main/java/org/symbol/sdk/symbol/SymbolTransactionFactory.java
```diff
@@ -0,0 +1,258 @@
...
+
+	/**
+	 * Attaches a signature to a transaction.
+	 *
+	 * @param transaction Transaction object.
+	 * @param signature Signature to attach.
+	 * @return JSON transaction payload.
+	 */
+	public static String attachSignature(final Transaction transaction, final Signature signature) {
+		transaction.setField("signature", new org.symbol.sdk.symbol.models.Signature(signature.bytes()));
```
Jaguar0625: setSignature ?

sdk/java/src/main/java/org/symbol/sdk/Network.java
```diff
@@ -0,0 +1,229 @@
...
+		 * @param names Names for which to search.
+		 * @return First network with a name in the supplied list.
+		 */
+		public static <TNetwork extends Network<?, ?>> TNetwork findByName(final List<TNetwork> networks, final List<String> names) {
+			for (TNetwork network : networks) {
+				if (names.contains(network.name))
+					return network;
+			}
+
+			throw new IllegalArgumentException(String.format("no network found with name '%s'", String.join(", ", names)));
```
Jaguar0625: any reason not to use streaming api here like in findByIdentifier?

sdk/java/src/main/java/org/symbol/sdk/RuleBasedTransactionFactory.java
```diff
@@ -0,0 +1,158 @@
...
+	public RuleBasedTransactionFactory() {
+		this(null, null);
+	}
+
+	/**
+	 * Registers a batch of named value-parse rules — typically a generated {@code Models.FACTORIES} registry.
+	 *
+	 * @param parsers Rule name → parse function.
+	 */
+	public void registerParsers(final Map<String, Function<Object, Object>> parsers) {
```
Jaguar0625: register/add PodParsers

sdk/java/src/test/java/org/symbol/sdk/AbstractNetworkTest.java
```diff
@@ -0,0 +1,230 @@
...
+		}
+
+		@Test
+		void derivesTestnetAddress() {
+			assertDerivesExpectedAddress(testnet(), AddressVector::testnet);
+		}
+	}
+
+	@Test
+	void addressRoundTripsThroughString() {
```
Jaguar0625: this looks like just a decode test, not a roundtrip ?

sdk/java/src/test/java/org/symbol/sdk/AbstractNetworkTest.java
```diff
@@ -0,0 +1,230 @@
...
+		}
+
+		@Test
+		void testnetAcceptsValidAndRejectsTamperedBytes() {
+			assertAcceptsValidAndRejectsTamperedBytes(testnet(), addressVectors()[0].testnet());
+		}
+	}
+
+	@Nested
+	class NetworkLocatorTest {
```
Jaguar0625: aren't these tests subset of tests in NetworkTest.java ?

sdk/java/src/test/java/org/symbol/sdk/NetworkTimestampTest.java
```diff
@@ -0,0 +1,173 @@
...
+			assertThat(new FakeTimestamp(1).isEpochal(), is(false));
+		}
+
+		@Test
+		void addSecondsAdvancesByCorrectAmount() {
+			// Arrange:
+			final FakeTimestamp t = new FakeTimestamp(100);
+
+			// Act + Assert:
+			assertThat(((FakeTimestamp) t.addSeconds(50)).timestamp, equalTo((long) (150)));
```
Jaguar0625: would prefer splitting act + assert

sdk/java/src/test/java/org/symbol/sdk/NetworkTimestampTest.java
```diff
@@ -0,0 +1,173 @@
...
+	final class TimestampBase {
+		@Test
+		void epochalTimestampIsRecognized() {
+			// Act + Assert:
+			assertThat(new FakeTimestamp(0).isEpochal(), is(true));
+			assertThat(new FakeTimestamp(1).isEpochal(), is(false));
+		}
+
+		@Test
+		void addSecondsAdvancesByCorrectAmount() {
```
Jaguar0625: techinically addSeconds is testing test function

sdk/java/src/test/java/org/symbol/sdk/NetworkTimestampTest.java
```diff
@@ -0,0 +1,173 @@
...
+		assertThat(actual, equalTo(expected));
+	}
+
+	@Nested
+	final class DatetimeConverter {
+		@Test
+		void canConvertNetworkTimestampToDatetimeInSeconds() {
+			// Arrange:
+			final Instant epoch = Instant.parse("2020-01-01T00:00:00Z");
+			final NetworkTimestamp.NetworkTimestampDatetimeConverter conv = new NetworkTimestamp.NetworkTimestampDatetimeConverter(epoch,
```
Jaguar0625: conv => converter ?

sdk/java/src/test/java/org/symbol/sdk/TransactionDescriptorProcessorTest.java
```diff
@@ -0,0 +1,570 @@
...
+	@Nested
+	class LookupValue {
+		@Test
+		void cannotLookupValueWhenDescriptorDoesNotContainKey() {
+			assertCannotLookupValueWhenDescriptorDoesNotContainKey(() -> createProcessor());
+		}
+
+		@Test
+		void cannotLookupValueWhenDescriptorDoesNotContainKeyWithConverter() {
+			assertCannotLookupValueWhenDescriptorDoesNotContainKey(() -> createProcessorWithConverter(null));
```
Jaguar0625, line 227: i believe this should bass ! null ?

sdk/java/src/test/java/org/symbol/sdk/symbol/NetworkTimestampTest.java
```diff
@@ -0,0 +1,116 @@
+package org.symbol.sdk.symbol;
```
Jaguar0625: i assume it was too unreadable to try and refactor shared NEM/Symbol tests in java?

sdk/java/src/test/java/org/symbol/sdk/symbol/NetworkTest.java
```diff
@@ -0,0 +1,169 @@
...
+		return new Address(addressString);
+	}
+
+	@Override
+	protected Address addressFromBytes(final byte[] addressBytes) {
+		return new Address(addressBytes);
+	}
+
+	@Nested
+	class NetworkTimestampTest {
```
Jaguar0625, line 65: shouldn't these be in NetworkTimestampTest ?

sdk/java/src/test/java/org/symbol/sdk/symbol/NetworkTimestampTest.java
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	@Test
+	void canCreateEpochalTimestamp() {
+		// Act:
+		final boolean actual = new NetworkTimestamp(0L).isEpochal();
+
+		// Assert:
+		assertThat(actual, is(true));
+		assertThat(actual, is(true));
```
Jaguar0625: double assert ?

sdk/java/src/test/java/org/symbol/sdk/nem/NetworkTimestampTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+		// Assert:
+		assertThat(ts.timestamp, equalTo(123L));
+		assertThat(ts.isEpochal(), is(false));
+	}
+
+	@Test
+	void canCreateEpochalTimestamp() {
+		assertThat(new NetworkTimestamp(0L).isEpochal(), is(true));
+		assertThat(new NetworkTimestamp(0L).isEpochal(), is(true));
```
Jaguar0625: double assert ?

sdk/java/src/test/java/org/symbol/sdk/symbol/NetworkTimestampTest.java
```diff
@@ -0,0 +1,116 @@
...
+		final NetworkTimestamp c = new NetworkTimestamp(43L);
+
+		// Assert:
+		assertThat(a, equalTo(b));
+		assertThat(a.hashCode(), equalTo(b.hashCode()));
+		assertThat(a.equals(c), is(false));
+	}
+
+	@Test
+	void doesNotEqualNemTimestampWithSameRawValue() {
```
Jaguar0625: not sure if needed ?

sdk/java/src/test/java/org/symbol/sdk/nem/NemTransactionFactoryTest.java
```diff
@@ -0,0 +1,374 @@
...
+				"Amount", "Height", "Timestamp", "Signature",
+
+				"LinkAction", "MessageType", "MosaicSupplyChangeAction", "MosaicTransferFeeType", "MultisigAccountModificationType",
+				"NetworkType", "TransactionType",
+
+				// ... plus the registered struct parsers ...
+				"struct:Message", "struct:NamespaceId", "struct:MosaicId", "struct:Mosaic", "struct:SizePrefixedMosaic",
+				"struct:MosaicLevy", "struct:MosaicProperty", "struct:SizePrefixedMosaicProperty", "struct:MosaicDefinition",
+				"struct:MultisigAccountModification", "struct:SizePrefixedMultisigAccountModification", "struct:CosignatureV1",
+				"struct:SizePrefixedCosignatureV1",
```
Jaguar0625, line 69: why do you need SizePrefixedCosignatureV1 and array[]?

sdk/java/src/test/java/org/symbol/sdk/nem/NemTransactionFactoryTest.java
```diff
@@ -0,0 +1,374 @@
...
+		descriptor.put("signerPublicKey", TEST_SIGNER_PUBLIC_KEY);
+		descriptor.put("message", Map.of("messageType", "plain", "message", messageText));
+
+		// Act:
+		final TransferTransactionV2 transaction = (TransferTransactionV2) factory.create(descriptor);
+
+		// Assert:
+		assertThat(transaction.getMessage().get().getMessage(), is(equalTo(messageText.getBytes())));
+	}
+
```
Jaguar0625, line 272: will the java sdk support the auto string handling added to the JS/PY sdks ?

sdk/java/src/test/java/org/symbol/sdk/nem/NemTransactionFactoryTest.java
```diff
@@ -0,0 +1,374 @@
...
+	void canConvertNonVerifiableTransactionToNonVerifiable() {
+		// Arrange:
+		final NonVerifiableTransferTransactionV1 nonVerifiable = new NonVerifiableTransferTransactionV1();
+
+		// Act:
+		final NonVerifiableTransaction result = NemTransactionFactory.toNonVerifiableTransaction(nonVerifiable);
+
+		// Assert: passing a NonVerifiable instance back through the converter yields a fresh
+		// NonVerifiable of the same type.
+		assertThat(result, is(org.hamcrest.Matchers.instanceOf(NonVerifiableTransferTransactionV1.class)));
```
Jaguar0625, line 344: isn't this same instance ?

sdk/java/src/test/java/org/symbol/sdk/nem/NemTransactionFactoryTest.java
```diff
@@ -0,0 +1,374 @@
...
+		// Arrange:
+		final NemTransactionFactory factory = new NemTransactionFactory(Network.TESTNET);
+		final Transaction transaction = factory.create(transferDescriptor());
+		final CryptoTypes.Signature signature = new CryptoTypes.Signature(new byte[CryptoTypes.Signature.SIZE]);
+
+		// Act:
+		final String json = NemTransactionFactory.attachSignature(transaction, signature);
+
+		// Assert:
+		assertThat(json, startsWith("{\"data\":\""));
```
Jaguar0625: remove ?

sdk/java/src/test/java/org/symbol/sdk/symbol/SymbolTransactionFactoryTest.java
```diff
@@ -0,0 +1,522 @@
...
+		// every generated pod/enum/flags (Models.FACTORIES) ...
+		final String[] expected = {
+				"Amount", "BlockDuration", "BlockFeeMultiplier", "Difficulty", "FinalizationEpoch", "FinalizationPoint", "Height",
+				"Importance", "ImportanceHeight", "MosaicId", "MosaicNonce", "MosaicRestrictionKey", "NamespaceId", "Timestamp",
+				"UnresolvedMosaicId", "Hash512", "Signature", "ProofGamma", "ProofScalar", "ProofVerificationHash",
+
+				"MosaicFlags", "AccountRestrictionFlags",
+
+				"AliasAction", "BlockType", "LinkAction", "LockHashAlgorithm", "MosaicRestrictionType", "MosaicSupplyChangeAction",
+				"NamespaceRegistrationType", "NetworkType", "ReceiptType", "TransactionType",
```
Jaguar0625, line 80: this is a little broad since it includes enums that don't appear in txes (e.g. BlockType), but probably ok

sdk/java/src/test/java/org/symbol/sdk/symbol/SymbolTransactionFactoryTest.java
```diff
@@ -0,0 +1,522 @@
...
+			final SymbolTransactionFactory factory = new SymbolTransactionFactory(Network.TESTNET);
+
+			// Act:
+			final CatbufferType transaction = create(factory, transferDescriptor());
+
+			// Assert:
+			assertThat(transaction.getField("type"), is(TransactionType.TRANSFER));
+			assertThat(transaction.getField("version"), is(equalTo(1)));
+			assertThat(transaction.getField("network"), is(NetworkType.TESTNET));
+			assertThat(transaction.getField("signerPublicKey"), is(equalTo(new PublicKey(TEST_SIGNER_PUBLIC_KEY.bytes()))));
```
Jaguar0625, line 172: is this only for top level tx? 🤔

sdk/java/src/test/java/org/symbol/sdk/symbol/SymbolTransactionFactoryTest.java
```diff
@@ -0,0 +1,522 @@
...
+			// Arrange:
+			final SymbolTransactionFactory factory = new SymbolTransactionFactory(Network.TESTNET);
+			final Transaction transaction = factory.create(transferDescriptor());
+			final CryptoTypes.Signature signature = new CryptoTypes.Signature(new byte[CryptoTypes.Signature.SIZE]);
+
+			// Act:
+			final String json = SymbolTransactionFactory.attachSignature(transaction, signature);
+
+			// Assert:
+			assertThat(json, startsWith("{\"payload\": \""));
```
Jaguar0625: needed ?

sdk/java/src/test/java/org/symbol/sdk/symbol/SymbolTransactionFactoryTest.java
```diff
@@ -0,0 +1,522 @@
...
+			// Arrange:
+			final SymbolTransactionFactory factory = new SymbolTransactionFactory(Network.TESTNET);
+			final Transaction transaction = factory.create(transferDescriptor());
+			final CryptoTypes.Signature signature = new CryptoTypes.Signature(patternBytes(CryptoTypes.Signature.SIZE, 0x20));
+			final String attached = SymbolTransactionFactory.attachSignature(transaction, signature);
+
+			// Act:
+			final String json = SymbolTransactionFactory.toJson(transaction);
+
+			// Assert: the payload embeds the serialized transaction, including the attached signature
```
Jaguar0625: are you intending to check the signature is in the payload ?

sdk/java/src/test/java/org/symbol/sdk/RuleBasedTransactionFactoryTest.java
```diff
@@ -0,0 +1,756 @@
...
+		final Map<String, Object> descriptor = new LinkedHashMap<>();
+		descriptor.put("mosaicId", 0x0123456789ABCDEFL);
+		descriptor.put("amount", 123456789123456789L);
+		return descriptor;
+	}
+
+	// region mock module
+
+	// Test stand-ins mirroring the generated shapes: pods carry static parse(Object), structs
+	// carry the setField/getField switch the descriptor processor dispatches through.
```
Jaguar0625, line 38: why do you need these? you want to use them in both pod and struct contexts?

sdk/java/src/test/java/org/symbol/sdk/RuleBasedTransactionFactoryTest.java
```diff
@@ -0,0 +1,756 @@
...
+			// Act:
+			final Object parsed = requireRule(factory, "SigningPublicKey").apply(publicKey);
+
+			// Assert:
+			assertThat(parsed, equalTo("pubkey " + publicKey));
+		}
+	}
+
+	// endregion
+
```
Jaguar0625, line 532: are you missing enum/flags tests ?

sdk/java/src/test/java/org/symbol/sdk/RuleBasedTransactionFactoryTest.java
```diff
@@ -0,0 +1,756 @@
...
+			factory.addStructParser("StructArrayMember", StructArrayMember::new);
+
+			final Map<String, Object> descriptor = new LinkedHashMap<>();
+			descriptor.put("mosaicIds", Arrays.asList(0x0123456789ABCDEFL, 0x3456789123456789L));
+
+			// Act:
+			final StructArrayMember parsed = (StructArrayMember) requireRule(factory, "struct:StructArrayMember").apply(descriptor);
+
+			// Assert:
+			assertThat(parsed.mosaicIds, contains(new MosaicId(0x0123456789ABCDEFL), new MosaicId(0x3456789123456789L)));
```
Jaguar0625: these should be ordered, so equals instead of contains ?

sdk/java/src/test/java/org/symbol/sdk/RuleBasedTransactionFactoryTest.java
```diff
@@ -0,0 +1,756 @@
...
+		}
+
+		@Test
+		void coercesSdkByteArraysViaSetField() {
+			// Arrange:
+			// a hand-written SDK Hash256 on a rule-less field is bridged to its same-named model twin by setField's asByteArray coercion
+			final RuleBasedTransactionFactory factory = new RuleBasedTransactionFactory();
+			factory.addStructParser("StructHashMember", StructHashMember::new);
+
+			final CryptoTypes.Hash256 publicHash = new CryptoTypes.Hash256(
```
Jaguar0625: you want a similar test where this is Models.Hash256

sdk/java/src/test/java/org/symbol/sdk/RuleBasedTransactionFactoryTest.java
```diff
@@ -0,0 +1,756 @@
...
+			descriptor.put("type", 123);
+
+			// Act:
+			final UnresolvedMosaic parsed = (UnresolvedMosaic) factory.createFromFactory(entityFactory, descriptor);
+
+			// Assert:
+			assertThat(parsed.mosaicId, equalTo(new MosaicId(0x0123456789ABCDEFL)));
+			assertThat(parsed.amount, equalTo(new Amount(123456789123456789L)));
+		}
+
```
Jaguar0625, line 959: are you missing `can create struct with type converter` test ?

also the new string conversion tests

Jaguar0625 (commented):
👍

sdk/java/src/main/java/org/symbol/sdk/nem/NemTransactionFactory.java
Wayonb: This factory also contains all the PODs....So maybe POD_FACTORIES?

sdk/java/src/main/java/org/symbol/sdk/symbol/AddressRules.java
Wayonb: In the transaction factory.  It's the pod parser for Address, which verifies it's not an alias, whereas UnresolvedAddress doesn't care.

sdk/java/src/main/java/org/symbol/sdk/symbol/SymbolTransactionFactory.java
Wayonb: It has to support embedded transactions also.  I was thinking if I should create a transaction-based

sdk/java/src/test/java/org/symbol/sdk/AbstractNetworkTest.java
Wayonb: Yes, but it shows that an address string can be converted to an Address object, then back to a string.  It just so happens to use decode/encode.

sdk/java/src/test/java/org/symbol/sdk/symbol/NetworkTest.java
Wayonb: JS has these in Network as well, I think? 🤔

sdk/java/src/test/java/org/symbol/sdk/nem/NemTransactionFactoryTest.java
Wayonb: This is because all the factories are registered, but will fail since they are not included in any transaction field.

sdk/java/src/test/java/org/symbol/sdk/nem/NemTransactionFactoryTest.java
Wayonb: >  auto string handling

maybe I missed this, can you explain?

sdk/java/src/test/java/org/symbol/sdk/nem/NemTransactionFactoryTest.java
Wayonb: No, it's instanceOf which checks the type

sdk/java/src/test/java/org/symbol/sdk/RuleBasedTransactionFactoryTest.java
Wayonb: You mean the override pods? I think JS has the same.  or something else?

sdk/java/src/main/java/org/symbol/sdk/symbol/SymbolTransactionFactory.java
Jaguar0625: that might make sense

sdk/java/src/main/java/org/symbol/sdk/symbol/AddressRules.java
Jaguar0625: i think this is different behavior than other sdks, which don't enforce this at factory level?

if so, i would note that somewhere, and probably add test(s) for this function

sdk/java/src/test/java/org/symbol/sdk/AbstractNetworkTest.java
Jaguar0625: isn't it just testing addressFromString, which is calling ctor? test is fine, i'd just rename it.

sdk/java/src/test/java/org/symbol/sdk/nem/NemTransactionFactoryTest.java
Jaguar0625: primarily these and related tests in symbol https://github.com/symbol/symbol/blob/dev/sdk/javascript/test/nem/TransactionFactory_spec.js#L191-L257

sdk/java/src/test/java/org/symbol/sdk/nem/NemTransactionFactoryTest.java
Jaguar0625: i mean, based on implementation, it seems that it's just returning the object in this case, so the comment seems misleading unless i misunderstand?

```java
		if (transaction instanceof NonVerifiableTransaction nv)
			return nv;	
```

sdk/java/src/test/java/org/symbol/sdk/AbstractNetworkTest.java
```diff
@@ -0,0 +1,253 @@
...
+	@Test
+	void mainnetRejectsTestnetAddress() {
+		assertThat(mainnet().isValidAddressString(addressVectors()[0].testnet()), is(false));
+	}
+
+	@Test
+	void testnetRejectsMainnetAddress() {
+		assertThat(testnet().isValidAddressString(addressVectors()[0].mainnet()), is(false));
+	}
+
```
Jaguar0625, line 201: did you remove tampered checksum test?

sdk/java/src/test/java/org/symbol/sdk/AbstractNetworkTest.java
```diff
@@ -0,0 +1,253 @@
...
+	// endregion
+
+	// region NetworkLocatorC
```
Jaguar0625: ~~C~~ ?

sdk/java/src/test/java/org/symbol/sdk/RuleBasedTransactionFactoryTest.java
Jaguar0625: ok, so this is to avoid the test having a dependency on generated models ?

my confusion was that the js tests do that for Hash256 but not SigningPublicKey (and not the BaseValue types, but you'll need to do that because of java typing)

sdk/java/src/test/java/org/symbol/sdk/symbol/NetworkTest.java
Jaguar0625: sorry my comment wasn't clear. i meant what is the difference between Network.NetworkTimestampTest and NetworkTimestampTest ?

Jaguar0625 (commented):
👍

sdk/java/src/test/java/org/symbol/sdk/nem/NemTransactionFactoryTest.java
Wayonb: ok... so mainly the discovery of nested structs in JSON format. I think I went with the TS implementation initially, since it is also a Typed language. I did not think this would use much, but I can revisit it.

sdk/java/src/main/java/org/symbol/sdk/symbol/AddressRules.java
Wayonb: I will add a note in the factory.
There are a few factory tests that check this. 
[createsAddressAliasWithResolvedAddress](https://github.com/symbol/symbol/blob/a34156d1c7d7e47364d364a842f12940dba6fd62/sdk/java/src/test/java/org/symbol/sdk/symbol/SymbolTransactionFactoryTest.java#L444) and [createRejectsAliasForResolvedAliasTransactionAddress](https://github.com/symbol/symbol/blob/a34156d1c7d7e47364d364a842f12940dba6fd62/sdk/java/src/test/java/org/symbol/sdk/symbol/SymbolTransactionFactoryTest.java#L464)

sdk/java/src/test/java/org/symbol/sdk/AbstractNetworkTest.java
Wayonb: no, they renamed to match the JS - assertCannotValidateInvalidAddress()

sdk/java/src/test/java/org/symbol/sdk/AbstractNetworkTest.java
Wayonb: removed

sdk/java/src/test/java/org/symbol/sdk/symbol/NetworkTimestampTest.java
Wayonb: They should mostly match 😅

sdk/java/src/test/java/org/symbol/sdk/symbol/NetworkTest.java
Wayonb: NetworkTimestampTest file = tests the abstract  NetworkTimestamp.Base class through a FakeTimestamp.  I am not sure if these are needed
Network.NetworkTimestampTest = tests the real per-chain classes through the network surface.
Did you want to remove fake tests or keep them the same as JS?

Wayonb: ah...I see what you are saying.  The extra NetworkTimestampTest file can be removed.  I think it got merged incorrectly.

sdk/java/src/test/java/org/symbol/sdk/TransactionDescriptorProcessorTest.java
Wayonb: Not sure, can you explain this?

sdk/java/src/test/java/org/symbol/sdk/symbol/SymbolTransactionFactoryTest.java
Jaguar0625: ?

sdk/java/src/test/java/org/symbol/sdk/nem/NemTransactionFactoryTest.java
Jaguar0625: i am assuming you don't want stricter test?

sdk/java/src/test/java/org/symbol/sdk/TransactionDescriptorProcessorTest.java
Jaguar0625: ignore, i misinterpreted what createProcessorWithConverter did with null

Jaguar0625 (commented):
👍

sdk/java/src/test/java/org/symbol/sdk/symbol/SymbolTransactionFactoryTest.java
Wayonb: Sorry, I missed this.  No, it's for top-level and embedded

Jaguar0625 (approved):
👍

PR #2126, opened 2026-07-30
[sdk/java]: add examples and vectors tests

Wayonb:
1. add examples as a separate project
2. add vector harness that verifies e2e parsing and signing of transactions
3. add a SymbolAdapter and NemAdapter to have a common interface for chains.  The interface can also apply directly to Nem or Symbol facades but would need to change the static methods.
4. Factory for Block and Receipt but then I noticed that these are not present in the other SDKs....can move. 😅 
5. scripts to run examples and vector tests.  Also to generate the Java docs.


sdk/java/examples/src/main/java/org/symbol/examples/readme/Symbol.java
```diff
@@ -0,0 +1,119 @@
...
+		signAndPrint(facade, typedTransaction);
+
+		System.out.println("*** EXAMPLE CONSTRUCTION FROM TYPED DESCRIPTOR - EMBEDDED ***");
+		final EmbeddedTransaction typedEmbedded = facade.createEmbeddedTransactionFromTypedDescriptor(typedDescriptor, signerPublicKey);
+		printEmbedded(typedEmbedded);
+	}
+
+	// JSON path — same descriptor shape as a JSON document. Integral numbers parse as Integer/Long/BigInteger by
+	// magnitude, so the full u64 range works as bare numbers; decimal or 0x-hex strings are also accepted.
+	private static void jsonExample(final SymbolFacade facade, final CryptoTypes.PublicKey signerPublicKey) {
```
Jaguar0625, line 70: include example for createEmbeddedTransactionFromJson ?

sdk/java/examples/src/main/java/org/symbol/examples/TransactionAggregate.java
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	public static void main(final String[] args) {
+		Path privatePath = null;
+		for (int i = 0; i < args.length; ++i) {
+			if ("--private".equals(args[i]) && i + 1 < args.length)
+				privatePath = Paths.get(args[++i]);
+			else if (args[i].startsWith("--private="))
+				privatePath = Paths.get(args[i].substring("--private=".length()));
+		}
```
Jaguar0625: refactor and move to ExamplesUtils since you're doing similar arg parsing in multiple files

sdk/java/examples/src/main/java/org/symbol/examples/TransactionAggregate.java
```diff
@@ -0,0 +1,100 @@
...
+
+		if (null == privatePath)
+			throw new IllegalArgumentException("missing required --private <path-to-private-key>");
+
+		final SymbolFacade facade = new SymbolFacade("testnet");
+		final KeyPair keyPair = ExamplesUtils.readPrivateKey(privatePath);
+
+		final Path resourcesDir = privatePath.toAbsolutePath().getParent();
```
Jaguar0625: fyi - this is different behavior than js version, which is relative to script (js) file, not private key file

sdk/java/examples/src/main/java/org/symbol/examples/TransactionSign.java
```diff
@@ -0,0 +1,158 @@
...
+				// deterministic header values (fee 625, deadline 12345) keep the output comparable with the JS / Python examples
+				final Map<String, Object> descriptor = JsonDescriptor.parse(json);
+				descriptor.put("signerPublicKey", keyPair.getPublicKey());
+				descriptor.put("fee", 625L);
+				descriptor.put("deadline", 12345L);
+
+				// descriptor strings are UTF-8 text; the secret_proof sample's proof is hex for raw bytes, so decode it at
+				// the call site — the same unhexlify step the JS / Python examples perform
+				if (descriptor.containsKey("proof"))
+					descriptor.put("proof", Converter.hexToUint8((String) descriptor.get("proof")));
```
Jaguar0625, line 99: can you elaborate? i don't see any special handling in JS version

sdk/java/src/main/java/org/symbol/sdk/symbol/SymbolBlockFactory.java
```diff
@@ -0,0 +1,75 @@
+package org.symbol.sdk.symbol;
```
Jaguar0625: is there reason you want this in src vs vectors?

sdk/java/src/test/java/org/symbol/sdk/facade/FacadeFactoryTest.java
```diff
@@ -0,0 +1,300 @@
...
+	private static <T extends CatbufferType, D extends TypedDescriptor, K extends KeyPair> BuiltTransaction buildThroughAdapter(
+			final BlockchainFacade<T, D, K> facade, final String json, final CryptoTypes.PublicKey signerPublicKey) {
+		final T transaction = facade.createTransactionFromJson(json, signerPublicKey, 100L, 3600L);
+		return new BuiltTransaction(transaction, facade.hashTransaction(transaction), facade.extractSigningPayload(transaction));
+	}
+
+	@Test
+	void symbolAdapterHashAndSigningPayloadMatchFacade() {
+		// Arrange:
+		final SymbolFacade real = new SymbolFacade("testnet");
```
Jaguar0625: real => facade ?

sdk/java/src/test/java/org/symbol/sdk/facade/FacadeFactoryTest.java
```diff
@@ -0,0 +1,300 @@
...
+				signerPublicKey);
+		final org.symbol.sdk.symbol.models.Transaction transaction = (org.symbol.sdk.symbol.models.Transaction) built.transaction();
+
+		// Assert: the adapter delegates hashTransaction / extractSigningPayload to the real facade (identical for the same tx)
+		assertThat(built.hash(), is(equalTo(real.hashTransaction(transaction))));
+		assertThat(built.signingPayload(), is(equalTo(real.extractSigningPayload(transaction))));
+	}
+
+	@Test
+	void nemAdapterHashAndSigningPayloadMatchFacade() {
```
Jaguar0625: can you refactor these two tests? and maybe some of the other NEM/Symbol ones as well ?

sdk/java/src/test/java/org/symbol/sdk/facade/FacadeFactoryTest.java
```diff
@@ -0,0 +1,300 @@
...
+		final BlockchainFacade<?, ?, ?> facade = FacadeFactory.create("nem", "testnet");
+		final CryptoTypes.PublicKey signerPublicKey = facade.createKeyPair(TEST_PRIVATE_KEY).getPublicKey();
+
+		// Act:
+		final org.symbol.sdk.nem.models.Transaction transaction = (org.symbol.sdk.nem.models.Transaction) facade
+				.createTransactionFromJson(NEM_TRANSFER_JSON, signerPublicKey, 100000L, 3600L);
+
+		// Assert: NEM fee is absolute (proving 100000 landed as the fee), and deadline is deadlineSeconds past the timestamp
+		assertThat(transaction.getFee().value(), is(equalTo(100000L)));
+		assertThat(transaction.getDeadline().value() - transaction.getTimestamp().value(), is(equalTo(3600L)));
```
Jaguar0625: did you want to test timestamp for symmetry with Symbol version ?

sdk/java/src/test/java/org/symbol/sdk/vectors/CatbufferVectorsHelper.java
```diff
@@ -0,0 +1,246 @@
...
+
+		return maps;
+	}
+
+	static Map<String, Object> toObjectMap(final Object value) {
+		if (!(value instanceof Map<?, ?> map))
+			throw new IllegalArgumentException("expected JSON object map but got " + (null == value ? "null" : value.getClass().getName()));
+
+		final Map<String, Object> result = new LinkedHashMap<>();
+		map.forEach((key, nested) -> result.put((String) key, nested));
```
Jaguar0625: nestd => value ?

sdk/java/src/test/java/org/symbol/sdk/vectors/CatbufferVectorsTest.java
```diff
@@ -0,0 +1,155 @@
...
+
+	private static Iterable<DynamicTest> generateConstructorTests(final String blockchain, final List<String> includes,
+			final Function<String, Serializer> constructor) {
+		final LinkedHashSet<String> schemaNames = new LinkedHashSet<>();
+		for (final Map<String, Object> item : CatbufferVectorsHelper.prepareTestCases(blockchain, includes, null))
+			schemaNames.add((String) item.get("schema_name"));
+
+		final List<DynamicTest> tests = new ArrayList<>();
+		for (final String schemaName : schemaNames)
+			tests.add(DynamicTest.dynamicTest(schemaName, () -> assertCreateFromConstructor(schemaName, constructor)));
```
Jaguar0625: is there reason you don't want to add to tests in first loop?

sdk/java/src/test/java/org/symbol/sdk/vectors/CatbufferVectorsTest.java
```diff
@@ -0,0 +1,155 @@
+package org.symbol.sdk.vectors;
```
Jaguar0625, line 1: can you explain how you decided to split catbuffer.js into this and CatbufferDescriptorVectorsTest?

for example, i'd expect the roundtrip tests here (since they aren't dependent on descriptors). ofc, descriptor tests are fine in other file.

sdk/java/src/test/java/org/symbol/sdk/vectors/NemTypedDescriptorVectorsTest.java
```diff
@@ -0,0 +1,259 @@
...
+
+				return descriptor;
+			}
+
+			case "cosignature_v1":
+				return new CosignatureV1Descriptor((String) plain.get("otherTransactionHash"),
+						(String) plain.get("multisigAccountAddress"));
+
+			case "multisig_transaction_v1": {
+				// nested transaction descriptors were deliberately rejected in the Java API; build the inner model first
```
Jaguar0625: can you not use root descriptor and convert to json?

sdk/java/src/test/java/org/symbol/sdk/vectors/SymbolTypedDescriptorVectorsTest.java
```diff
@@ -0,0 +1,361 @@
...
+	// endregion
+
+	// region value helpers
+
+	private static String[] toStringArray(final Object values) {
+		final List<?> list = (List<?>) values;
+		return list.stream().map(String.class::cast).toArray(String[]::new);
+	}
+
+	private static CryptoTypes.Hash256 transactionsHashOf(final Map<String, Object> plain) {
```
Jaguar0625: maybe rename to get/extract*  ?

Jaguar0625 (commented):
👍

sdk/java/examples/src/main/java/org/symbol/examples/TransactionAggregate.java
Wayonb: Yeah, but with the JS example, the resources folder and the JS file are in the same folder.  
In Java, it's different, and I found this to be the easier way to get the resources path since the private key file is in the same folder.   Did you want to remove the absolute path?

sdk/java/examples/src/main/java/org/symbol/examples/TransactionSign.java
Wayonb: The proof is converted from a hex string to bytes in JS/Python.  Since Jackson JSON only supports UTF-8, I added this to convert explicitly.
I could convert the bytes to a UTF-8 string and put it in the JSON, but then the examples look different.
```
		{
			type: 'secret_proof_transaction_v1',
			recipientAddress: sampleAddress,
			secret,
			hashAlgorithm: 'hash_160',
			proof: (new ByteArray(4, 'C1ECFDFC')).bytes
		}
```

sdk/java/src/main/java/org/symbol/sdk/symbol/SymbolBlockFactory.java
Wayonb: No, I can move to the vectors since I am not sure of a real use case for these.

sdk/java/examples/src/main/java/org/symbol/examples/TransactionAggregate.java
Jaguar0625: i was just pointing it out because i was not sure if it was an intentional difference. i'd just suggest adding a comment.

sdk/java/src/main/java/org/symbol/sdk/symbol/SymbolBlockFactory.java
Jaguar0625: would rather move them to keep the surface of the various sdks consistent

sdk/java/src/test/java/org/symbol/sdk/facade/FacadeFactoryTest.java
Wayonb: Can you explain? For Symbol, the tx doesn't have two timestamps, so I just have to estimate. but maybe misunderstand 😅

sdk/java/src/test/java/org/symbol/sdk/vectors/CatbufferVectorsHelper.java
Wayonb: There is already a value so I used nestedValue

sdk/java/src/test/java/org/symbol/sdk/vectors/CatbufferVectorsTest.java
Wayonb: tbh, the initial split was because I was using reflection in some tests.  But now I've moved away from using reflection.  let me restructure it.

sdk/java/src/test/java/org/symbol/sdk/facade/FacadeFactoryTest.java
```diff
@@ -0,0 +1,603 @@
...
+		return modelClass.cast(transaction);
+	}
+
+	private static CryptoTypes.Signature tamperSignature(final CryptoTypes.Signature signature) {
+		final byte[] bytes = signature.bytes().clone();
+		bytes[0] ^= (byte) 0xFF;
+		return new CryptoTypes.Signature(bytes);
+	}
+
+	/** A transaction built through an adapter, plus one signature for it. */
```
Jaguar0625, line 59: is there reason you're using "adapter" instead of "facade" ?

sdk/java/examples/src/main/java/org/symbol/examples/TransactionAggregate.java
Jaguar0625: please add comment

sdk/java/src/test/java/org/symbol/sdk/facade/FacadeFactoryTest.java
Jaguar0625: symbol is testing the timestamp is relative to now(). nem is not.

sdk/java/src/test/java/org/symbol/sdk/vectors/CatbufferVectorsHelper.java
Jaguar0625: you are using "nested" elsewhere, so you should use "nested" here too, instead of just renaming this one occurrence to "nestedValue"

sdk/java/src/test/java/org/symbol/sdk/vectors/CatbufferVectorsTest.java
```diff
@@ -0,0 +1,183 @@
...
+		final byte[] payload = Converter.hexToUint8((String) item.get("payload"));
+
+		// Act:
+		final Serializer instance = deserializer.apply(ByteBuffer.wrap(payload));
+
+		// Assert: the deserialized type matches the vector schema
+		assertThat("deserialized type must match vector schema_name for " + item.get("test_name"), instance.getClass().getSimpleName(),
+				is(equalTo(item.get("schema_name"))));
+		CatbufferVectorsHelper.assertPayload(item, instance, "roundtrip");
+		assertThat("size mismatch for " + item.get("test_name"), instance.size(), is(equalTo(payload.length)));
```
Jaguar0625, line 104: should you check size before assertPayload? otherwise, it seems assertPayload will always fail when size is not equal?

Jaguar0625 (commented):
👍

sdk/java/src/test/java/org/symbol/sdk/facade/FacadeFactoryTest.java
Wayonb: I was trying to keep the naming consistent since I use facade for the Nem/SymbolFacade, and in some places I have both.

Jaguar0625 (approved):
👍

PR #2128, opened 2026-07-31
[sdk]: fix bug in symbol fee calculation

Jaguar0625:
problem: cosignature size is not being multipled by fee multiplier
solution: multiply it


zero4862 (approved):

Wayonb (approved):
👍🏾

PR #2130, opened 2026-07-31
[sdk/java] fix: symbol fee calculation with cosigners

Wayonb:
problem: cosignature size was not multiplied by fee multiplier
solution: add transaction and cosignature size, then multiply by the fee multiplier


Jaguar0625 (approved):
👍

PR #2140, opened 2026-08-04
[mkdocs] fix: handle 64-bit values safely in tutorial snippets

zero4862:
## Problem

Some tutorial snippets handle uint64 values (amounts, supplies, heights) through floats or JS `Number`, losing precision the REST API deliberately preserves by encoding them as strings.

Also, hex identifiers (mosaic IDs, namespace IDs, metadata keys) are also formatted inconsistently across tutorials, and some drop leading zeros, leading to potentially produce IDs shorter than the 16 characters the REST API requires.

## Solution

* Parse uint64 values with `BigInt` (JS) / `int` (Python) end to end and format amounts with integer div/mod instead of float division.
* Format all 64-bit IDs in one canonical form in both languages: 16-character, zero-padded, uppercase hex, matching the REST API.
* Update the expected-output `.log` files to match.

## Testing

All modified snippets were run against a testnet node, none break, and output matches the updated `.log` files.


segfaultxavi (commented):
You're an absolute legend for taking the time to fix this! :D

segfaultxavi (approved):

PR #2157, opened 2026-08-25
[sdk/java] fix: create a single ctor for descriptors

Wayonb:
problem: descriptors have required values in the constructor and others in setters, which can be unclear
solution: create one constructor with all parameters

### DIFF
--- a/sdk/java/examples/src/main/java/org/symbol/examples/TransactionAggregate.java
+++ b/sdk/java/examples/src/main/java/org/symbol/examples/TransactionAggregate.java
@@ -13,6 +13,8 @@ import org.symbol.sdk.symbol.KeyPair;
 import org.symbol.sdk.symbol.SymbolTransactionFactory;
 import org.symbol.sdk.symbol.descriptors.AggregateCompleteTransactionV3Descriptor;
 import org.symbol.sdk.symbol.descriptors.TransferTransactionV1Descriptor;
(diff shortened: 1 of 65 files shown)


segfaultxavi (approved):
👍

sdk/java/examples/src/main/java/org/symbol/examples/TransactionAggregate.java
```diff
@@ -39,7 +39,7 @@ private static List<EmbeddedTransaction> addEmbeddedTransfers(final SymbolFacade
 			// and other tools that treat messages starting with 00 byte as "plain text"
 			final byte[] prefixed = ArrayHelpers.concat(new byte[1], messageBytes);
 
-			final TransferTransactionV1Descriptor descriptor = new TransferTransactionV1Descriptor(recipientAddress).message(prefixed);
+			final TransferTransactionV1Descriptor descriptor = new TransferTransactionV1Descriptor(recipientAddress, /* mosaics */ null, prefixed);
```
Jaguar0625: since this is example, would probably go with 
```
(List<UnresolvedMosaicDescriptor>) null
```
vs 
```
/* mosaics */ null
```

assuming that doesn't lead to any errors ofc 😅

Jaguar0625 (approved):
👍

Jaguar0625 (commented):
👍

PR #2159, opened 2026-08-27
[mkdocs] feat: use the Descriptor API for transaction creation in all tutorials

segfaultxavi:
Transactions can be created from JS (and Java) from Typed Descriptors, which is more convenient (it takes care or deadlines and fees) and type-safe.
But we were not using this route, to keep the code on par with Python, which does not have typed descriptors.

The addition of the new create_from_descriptor method in Python allows keeping code snippets in all languages parallel, and simplifying the code a bit.
This PR changes ALL tutorials to create transactions from descriptors, in all languages.

The old page about Typed Descriptors on JS has been changed to talk about Manual Transaction Creation instead.


mkdocs/pages/en/devbook/transactions/transfer.md
```diff
@@ -111,11 +125,9 @@ The snippet includes the following fields:
     Amounts are expressed in atomic units, which depend on the mosaic's <divisibility:>.
     For XYM, the divisibility is 6, so 1 XYM must be expressed as `1_000_000`.
 
-Note that the `fee` field is not set in the descriptor.
-Instead, the fee is calculated after the transaction is built, once its size in bytes is known.
-
-The SDK's <dy:FeeCalculator.calculateTransactionFee> helper computes the fee by multiplying the transaction size by
-the fee multiplier.
+The descriptor does not include common transaction fields such as the signer public key, deadline, or fee.
```
Wayonb, line 133: descriptor => typed descriptor.

Wayonb (approved):
👍🏾

mkdocs/pages/en/devbook/transactions/transfer.md
segfaultxavi: The word "descriptors" is used in multiple places, if I turn all of them into "typed descriptors" it will be extremely verbose.
However, I've added text to clarify the difference between typed and untyped descriptors, so I hope there will be no confusion for the reader.

segfaultxavi:
@Wayonb I guess this PR cannot be merged until the changes to the Python SDK are published?

PR #2160, opened 2026-08-27
[jenkins]: add Gradle publisher

Wayonb:
problem: Jenkins does not publish the Gradle project
solution: add logic to publish with Gradle


Jaguar0625 (approved):
👍

PR #2176, opened 2026-09-02
[jenkins] fix: catapult base test image fails to build

Wayonb:
problem: Linux perf tool fails to build due to missing dependencies
solution: add pkg-config and build-essential to the Docker image


Jaguar0625 (approved):
👍

PR #2177, opened 2026-09-02
[jenkins] fix: increase docker openfile limit in Jenkins pipeline

Wayonb:
problem: MongoDB is failing during the Rest and Catapult tests due to too many open files
solution: increase the open file limit in the Jenkins pipeline


Jaguar0625 (approved):
👍

PR #2180, opened 2026-09-04
[jenkins] fix: catapult tests failing on clang build

Wayonb:
problem: catapult tests are failing on clang builds to find libc++.
                This is due to symlinks for the compiler files not getting resolved.
solution: copy the actual compiler file instead of symlinks


Jaguar0625 (approved):
👍

PR #2186, opened 2026-09-07
[jenkins]: upgrade to Clang version 22 and 23

Wayonb:
[jenkins]: upgrade the clang compiler
problem: clang is using an outdated compiler
solution: upgrade to the latest clang versions

[client/catapult]: update code to build with Clang 22/23
1. disable lifetime safety feature
2. add missing headers
3. fix the tools import directories to prevent include conflicts that were triggering the ``shadow-header`` rule.

Also, updated libzmq and rocksdb to build with clang 23 by including missing headers


jenkins/catapult/baseImageDockerfileGenerator.py
```diff
@@ -166,6 +166,9 @@ def libzmq(self):
 			# try to pass additional flags to disable faulty optimizations
 			descriptor.cxxflags += ['-mno-avx', '-mno-avx2']
 
+		if self.is_clang and self.compiler.version <= 23:
```
Jaguar0625: is this intentionally <= instead of >= ?

Jaguar0625 (changes requested):
👍

Jaguar0625 (approved):
👍

PR #2187, opened 2026-09-11
[client/catapult]: fix several failure on the Windows build

Wayonb:
problem: several variables were not captured for lambda use
                 Duplicate symbols since the link was with the DLL
solution: capture all variables needed for the lambda
                 link only with the static .deps library


Jaguar0625 (approved):
👍

PR #2191, opened 2026-09-14
[client/catapult] fix: clang fails if constexpr are captured

Wayonb:
problem: clang fail due to it set ``-Wunused-lambda-capture`` flags on explicit capture as unnecessary
solution: only capture constexpr for MSVC


Jaguar0625 (approved):
👍

PR #2192, opened 2026-09-15
[client/catapult]: update the catapult build to C++20

Wayonb:
problem: catapult cannot upgrade to the latest version of RocksDB since it requires C++20
solution: update catapult build to C++20


client/catapult/tests/test/net/BriefServerRequestorTestUtils.h
```diff
@@ -208,16 +209,20 @@ namespace catapult { namespace test {
...
 		void setServerSocket(const std::shared_ptr<ionet::PacketSocket>& pServerSocket) {
-			std::atomic_store(&m_pServerSocket, pServerSocket);
+			std::lock_guard<std::mutex> guard(m_serverSocketMutex);
+			m_pServerSocket = pServerSocket;
 		}
 
 	private:
 		test::TcpAcceptor m_acceptor;
 
+		// std::atomic<std::shared_ptr> is not implemented by libc++, so guard it with a mutex instead
```
Jaguar0625: this seems to just have been moved from `<atomic>` to `<memory>`

jenkins/catapult/BasicBuildManager.py
```diff
@@ -1,6 +1,7 @@
 from configuration import load_build_configuration, load_compiler_configuration
 
 
+# pylint: disable=too-many-instance-attributes
```
Jaguar0625: this should be after `class BasicBuildManager:``

Jaguar0625 (changes requested):
👍

client/catapult/tests/test/net/BriefServerRequestorTestUtils.h
Wayonb: You mean they are available in the memory header? 🤔

Jaguar0625: yea, see: https://en.cppreference.com/cpp/memory/shared_ptr/atomic2

Wayonb: ok, thanks.  I will take a look, but I'm not sure that Clang implements it with C++20.
https://discourse.llvm.org/t/rfc-implementing-std-atomic-shared-ptr-t-and-std-atomic-weak-ptr-t-p0718r2-for-libc/91586

Wayonb: It still seems to be not supported by Clang - https://github.com/llvm/llvm-project/issues/99980

Wayonb: This is the error from MacOS.
```
  224 |                 std::atomic<std::shared_ptr<ionet::PacketSocket>> m_pServerSocket;
      |                                                                   ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/v1/__atomic/atomic.h:238:17: note: 'std::shared_ptr<catapult::ionet::PacketSocket>' is not trivially copyable
  238 |   static_assert(is_trivially_copyable<_Tp>::value, "std::atomic<T> requires that 'T' be a trivially copyable type");
  ```

Jaguar0625 (approved):
👍

